PR #228, opened 2023-02-22
Locking is optional

kazet:


artemis/module_base.py
```diff
@@ -125,31 +127,34 @@ def run(self, current_task: Task) -> None:
     def internal_process(self, current_task: Task) -> None:
         scan_destination = self._get_scan_destination(current_task)
 
-        lock = ResourceLock(Config.REDIS, f"lock-{scan_destination}")
+        if self.lock_target:
+            lock = ResourceLock(Config.REDIS, f"lock-{scan_destination}")
+
+            try:
+                lock.acquire(max_tries=Config.SCAN_DESTINATION_LOCK_MAX_TRIES)
```
BonusPlay, line 134: We should use RAII for locks, otherwise we're going to forget to release/take one of them at some point.

kazet: how would you change this code?

BonusPlay: yes

BonusPlay (approved):

PR #229, opened 2023-02-23
Defaulting to black 88 default chars per line as it's CERT standard as well

kazet:


BonusPlay:
1) Reduced Readability: Code that has lines of only 88 characters can be difficult to read, especially when dealing with long variable names or complicated expressions. It can cause eye strain and make it harder for developers to quickly parse and understand the code.

2) Inefficient Use of Space: Modern monitors and displays have a lot of horizontal space, so limiting lines to only 88 characters wastes valuable screen real estate. It can also make it difficult to have multiple windows open side by side, which is a common practice for developers.

3) Breaks Standard Conventions: Most programming languages have established conventions for line length, which are typically longer than 88 characters. Using a shorter line length can make code appear non-standard and harder to maintain.

Thank you for reading my chat gpt output.

nazywam:
Changing this in a project, where a limit is already established doesn't seem very productive

PR #233, opened 2023-02-28
Changed Joomla version check so that we ensure newer version for a branch is available

kazet:


artemis/modules/joomla_scanner.py
```diff
@@ -34,15 +35,33 @@ def _is_version_old(self, version: str, age_threshold_days: int = 180) -> bool:
...
+            release_version = release["tag_name"]
+            release_version_parsed = semver.VersionInfo.parse(release_version)
+            if (
+                release_version_parsed.major == version_parsed.major
+                and release_version_parsed.compare(version_parsed) > 0
+            ):
+                is_newer_version_available = True
+
+        # To consider a version old, it must:
+        # - be obsolete enough (we don't want to send notifications if a version has been released yesterday)
```
nazywam, line 61:
```suggestion
        # - be obsolete enough (we don't want to send notifications if a version was released yesterday)
```

test/modules/test_joomla_scanner.py
```diff
@@ -10,6 +11,30 @@ class JoomlaScannerTest(ArtemisModuleTestCase):
     # The reason for ignoring mypy error is https://github.com/CERT-Polska/karton/issues/201
     karton_class = JoomlaScanner  # type: ignore
 
+    def test_is_version_old(self) -> None:
```
nazywam, line 14: Don't we also have to mock time because `datetime.datetime.utcnow().replace(tzinfo=pytz.utc) - datetime.datetime.fromisoformat(release["published_at"])` may change over time?

nazywam (approved):

test/modules/test_joomla_scanner.py
kazet: we didn't, but I added a test that actually check the date difference feature

artemis/modules/joomla_scanner.py
```diff
@@ -34,15 +35,33 @@ def _is_version_old(self, version: str, age_threshold_days: int = 180) -> bool:
...
+            return True
+
+        is_current_version_old = True  # If we don't find the version on releases list, that means, it's old
         for release in data:
             if release["tag_name"] == version:
-                return (
+                is_current_version_old = (
                     datetime.datetime.utcnow().replace(tzinfo=pytz.utc)
                     - datetime.datetime.fromisoformat(release["published_at"])
                 ).days > age_threshold_days
```
BonusPlay, line 48: mmm, worst code formatting I've seen this week.
```suggestion
                current_time = datetime.datetime.utcnow().replace(tzinfo=pytz.utc)
                published_time = datetime.datetime.fromisoformat(release["published_at"]
                time_difference = (current_time - published_time).days
                is_current_version_old = time_difference > age_threshold_days
```
also, it's common to use
```py
from datetime import datetime
```

artemis/modules/joomla_scanner.py
```diff
@@ -34,15 +35,33 @@ def _is_version_old(self, version: str, age_threshold_days: int = 180) -> bool:
...
-        return True
+        is_newer_version_available = False
+        for release in data:
+            release_version = release["tag_name"]
+            release_version_parsed = semver.VersionInfo.parse(release_version)
+            if (
+                release_version_parsed.major == version_parsed.major
+                and release_version_parsed.compare(version_parsed) > 0
+            ):
+                is_newer_version_available = True
```
BonusPlay, line 58: rewrite as well.
```py
if (condition):
     variable = True
```
is not the best practice. Also `release_version` is 15 characters, while `release["tag_name"]` is 19, not really much of a helper. it would be much better to split if into 2 helper variables (since they are long) and remove `release_version` entirely.

Also, `release_version_parsed.compare(version_parsed) > 0` needs a comment. How do you compare 2 semver and result is an int???

BonusPlay (changes requested):

artemis/module_base.py
```diff
@@ -29,6 +29,10 @@ class ArtemisBase(Karton):
 
     task_poll_interval_seconds = 2
 
+    # After this number of tasks, the service will get restarted. This is to prevent
+    # situations such as slow memory leaks.
+    max_num_tasks_to_process = 1000
```
BonusPlay, line 34: Should be configurable

BonusPlay (changes requested):
Last request

BonusPlay (approved):

PR #237, opened 2023-03-01
Retries in crt.sh, not failing silently

kazet:


nazywam:
This should probably be handled with `HTTPAdapter` and a `urllib3.util.Retry` adapter

kazet:
> This should probably be handled with `HTTPAdapter` and a `urllib3.util.Retry` adapter instead of implementing our own logic

with the logic we have we will retry both techniques: SQL access and web access

nazywam:
```
f" WHERE plainto_tsquery('certwatch', '{domain}') @@ identities(certificate)"
```

why are we using string formatting instead of properly calling the query with arguments?

artemis/modules/crtsh.py
```diff
@@ -53,22 +59,33 @@ def query_sql(self, domain: str) -> set[str]:
...
         if response.ok:
+            ct_domains: set[str] = set()
             for cert in response.json():
                 for entry in cert["name_value"].split("\n"):
                     if re.fullmatch(DOMAIN_REGEX, entry):
                         ct_domains.add(entry)
-        return ct_domains
+            return ct_domains
+        else:
+            raise UnableToObtainSubdomainsException()
```
nazywam, line 71: Seems like could just use `response.raise_for_status()` instead of inventing our own exceptions :thinking:

artemis/modules/crtsh.py
```diff
@@ -53,22 +59,33 @@ def query_sql(self, domain: str) -> set[str]:
...
         with self.lock:
-            try:
-                ct_domains = self.query_sql(domain)
-            except OperationalError:
-                ct_domains = self.query_json(domain)
+            for retry_id in range(Config.CRTSH_NUM_RETRIES):
+                try:
+                    ct_domains = self.query_sql(domain)
+                except OperationalError:
+                    try:
```
nazywam, line 80: There has to be a better way of doing this than a code block nested 7 times

kazet:
> ```
> f" WHERE plainto_tsquery('certwatch', '{domain}') @@ identities(certificate)"
> ```
> 
> why are we using string formatting instead of properly calling the query with arguments?

done in a more proper way

and reason? it was already there earlier and didn't yet get fixed

artemis/modules/crtsh.py
```diff
@@ -27,48 +34,67 @@ class CrtshScanner(ArtemisBase):
...
+        try:
+            ct_domains: set[str] = set()
+            conn = connect("postgresql://guest@crt.sh:5432/certwatch")
+            conn.set_session(readonly=True, autocommit=True)
+            with conn.cursor() as cursor:
+                # Validate characters as we are putting the domain inside a LIKE query
+                # Escaping LIKE queries is complex:
+                # https://stackoverflow.com/questions/2106207/escape-sql-like-value-for-postgres-with-psycopg2
+                # so for now let's just whitelist domain characters.
+                assert all([c in string.ascii_letters + "-" + string.digits + "." or ord(c) >= 0x80 for c in domain])
```
nazywam, line 47: This doesn't seem that needed since we construct the query correctly

artemis/modules/crtsh.py
```diff
@@ -27,48 +34,67 @@ class CrtshScanner(ArtemisBase):
...
+                results = cursor.fetchall()
 
-            for row in results:
-                for entry in row:
-                    if re.fullmatch(DOMAIN_REGEX, entry):
-                        ct_domains.add(entry)
-        conn.close()
-        return ct_domains
+                for row in results:
+                    for entry in row:
```
nazywam, line 61: Why do we iterate the row if it returns a single column?

nazywam (dismissed):

artemis/modules/crtsh.py
kazet: fixed

artemis/modules/crtsh.py
kazet: there's % for LIKE queries - and I am too scared to delete thsi

PR #243, opened 2023-03-01
Some FTP servers require TLS

kazet:


artemis/modules/ftp_bruter.py
```diff
@@ -50,20 +52,43 @@ def run(self, current_task: Task) -> None:
...
+                            pass
+
+                    result.welcome = ftp.welcome
+                    result.tls = tls
+
+                    try:
+                        throttle_request(lambda: ftp.login(username, password))
+                        result.credentials.append((username, password))
+                        result.files.extend(ftp.nlst())
+                    except ftplib.error_temp:
```
nazywam, line 84: Assuming we tried to login using cleartext and that didn't shouldn't we somehow detect invalid credentials and not try again to login using them for tls?

nazywam (approved):

artemis/modules/ftp_bruter.py
kazet: maybe but I'm not sure I would be able to detect invalid credentials in a robust way

PR #244, opened 2023-03-02
GAU integration

kazet:


test/modules/test_gau.py
```diff
@@ -0,0 +1,35 @@
...
+        ]
+
+        for entry in data:
+            task = Task(
+                {"type": TaskType.DOMAIN},
+                payload={TaskType.DOMAIN: entry.domain},
+            )
+            results = self.run_task(task)
+
+            found = False
```
nazywam, line 31: You can simplify this to
```python
found = any(x.payload["domain"] == entry.expected_subdomain for x in results)
```

artemis/modules/gau.py
```diff
@@ -0,0 +1,62 @@
...
+                url_parsed = urllib.parse.urlparse(line)
+                if not url_parsed.hostname:
+                    continue
+
+                if is_subdomain(url_parsed.hostname, domain):
+                    domains.add(url_parsed.hostname)
+
+            for domain in domains:
+                task = Task(
+                    {"type": TaskType.DOMAIN},
```
nazywam, line 50: Not missing a `.value` here?

nazywam: On a related note, how do you mitigate feedback loops? This karton service is producing the same headers it accepts, which if I'm not mistaken, can lead to an endless loop

artemis/modules/gau.py
```diff
@@ -0,0 +1,62 @@
...
+                [
+                    "gau",
+                ]
+                + Config.GAU_ADDITIONAL_OPTIONS,
+                self.log,
+                input=domain.encode("ascii"),
+            )
+
+            domains = set()
+            for line in output.decode("ascii", errors="ignore").split("\n"):
```
nazywam, line 40: `.splitlines()` would probably be safer to use

nazywam (approved):

artemis/modules/gau.py
kazet: I don't think .value is needed here, the karton I copied this from works without

regarding feedback loops, here are two kinds of loops possible:

- adding a record that already exists in a cyclical way (e.g. domain -> ip -> revdns -> domain) which is prevented by add_task() being a wrapper that makes sure such task doesn't already exist,
- adding an endless stream of different e.g. subdomains (domain -> git.subdomain -> git.git.subdomain ->  ...) which is not prevented but has also not been observed yet

nazywam (approved):

PR #250, opened 2023-03-04
Resource leaks prevention

kazet:


BonusPlay:
But why?

kazet:
for sslyze, I have observed that it sometimes doesn't remove a thread, which leads to problems when a karton runs for a long time. I don't want to debug and fix all libraries with similar problems, so I added this hack.

PR #306, opened 2023-04-19
The possibility to sort and filter on analysis list

kazet:


templates/index.jinja2
```diff
@@ -5,43 +5,17 @@
...
+        <table id="analysis_list" class="table table-hover">
+            <thead>
+                <tr>
+                    <th scope="col">Target</th>
+                    <th scope="col">Tag</th>
+                    <th scope="col">Status</th>
+                    <th scope="col">Actions</th>
+                </tr>
+            </thead>
+        </table>
```
bulkowy, line 18: Should table render nothing only table headers?

docker-compose.yaml
```diff
@@ -44,7 +44,7 @@ services:
     restart: always
     volumes: ["./docker/karton.ini:/etc/karton/karton.ini"]
 
-  backend:
+  frontend:
```
bulkowy, line 47: or just `web`/`webapp`

templates/index.jinja2
kazet: yup, this is filled by javascript

bulkowy (approved):

PR #389, opened 2023-06-22
Supporting hooks

kazet:


artemis/reporting/export/hooks.py
```diff
@@ -0,0 +1,22 @@
...
+from artemis.reporting.export.hook import ExportHook
+
+
+@functools.lru_cache(maxsize=1)
+def get_all_hooks() -> List[Type[ExportHook]]:
+    """Finds all ExportHooks, i.e. all classes inheriting from ExportHook in artemis.reporting.export.hook_modules."""
+    modules_dir = os.path.join(os.path.dirname(__file__), "hook_modules")
+    for item in os.listdir(modules_dir):
+        if item.endswith(".py") and item != "__init__.py":
+            module_name = item.split(".")[0]
```
bulkowy, line 15: not better `replace(".py", "")`? cuz I think it is what it is about

bulkowy (dismissed):

bulkowy (approved):

PR #400, opened 2023-06-27
Filtering obsolete report types

kazet:


artemis/reporting/export/previous_reports.py
```diff
@@ -7,7 +7,9 @@
 from artemis.reporting.base.report import Report
 
 
-def load_previous_reports(previous_reports_directory: Path) -> List[Report]:
+def load_previous_reports(
+    previous_reports_directory: Path, obsolete_report_types: List[str] = ["almost_expired_ssl_certificate"]
```
bulkowy, line 11: What makes report type obsolete? Should not it be stored in config then?

PR #404, opened 2023-06-29
Removing Artemis functions that  are duplicated by Nuclei

kazet:


artemis/modules/data/bruter_additional_paths.txt
```diff
@@ -12,15 +12,8 @@ graph
 database.sql
 adminbackups
 core
-wp-admin/install.php
```
bulkowy, line 15: which one added nuclei template duplicates wp-admin?

kazet: http/misconfiguration/installer/wp-install.yaml which is used as it's critical-severity

bulkowy (approved):

PR #454, opened 2023-08-01
Prometheus metrics export

kazet:


artemis/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+    PLATFORM_COLLECTOR,
+    PROCESS_COLLECTOR,
+    REGISTRY,
+    start_http_server,
+)
+from prometheus_client.core import GaugeMetricFamily
+from prometheus_client.registry import Collector
+
+
+class ArtemisMetricsCollector(Collector):
```
bulkowy, line 17: I think values provided in this Collector should have been (or should include) sums over period of time (eg. day or scan cycle)

kazet: IMO nope - this endpoint includes current status and calculating other statistics out of that (e.g. daily change) is the resposibility of e.g. the tool that plots the statistics (e.g. Grafana)

bulkowy (approved):

PR #456, opened 2023-08-01
Alerts when there are too much Nuclei reports

kazet:


artemis/reporting/base/reporter.py
```diff
@@ -20,6 +20,12 @@ def get_report_types(cls) -> List[ReportType]:
         """Types of reports provided by this Reporter."""
         return [getattr(cls, name) for name in dir(cls) if isinstance(getattr(cls, name), ReportType)]
 
+    @staticmethod
+    def get_alerts(all_reports: List[Report]) -> List[str]:
+        """This looks at the final reports list and returns messages to be shown to the person
+        that exports the e-mails (e.g. potential false positivves)."""
```
bulkowy, line 26:
```suggestion
        that exports the e-mails (e.g. potential false positives)."""
```

bulkowy (approved):

PR #465, opened 2023-08-08
More ways of detecting wp

kazet:


artemis/modules/webapp_identifier.py
```diff
@@ -38,6 +38,9 @@ def _identify(url: str) -> WebApplication:
             if re.search(webapp_sig, response.text):
                 return webapp_id
 
+        if "/wp-includes/css/" in response:
```
bulkowy, line 41: is it not better to check for tags (`<script>` I assume), that have `.*/wp-includes/.*` in their attributes?

kazet: that would require HTML parsing

bulkowy (approved):

PR #469, opened 2023-08-12
blocklists have tests and more features

kazet:


artemis/reporting/blocklist.py
```diff
@@ -1,19 +1,22 @@
...
 
 
 @dataclasses.dataclass
 class BlocklistItem:
-    domain: str
-    until: Optional[datetime.datetime]
-    report_type: Optional[ReportType]
+    domain: Optional[str] = None
+    ip_range: Optional[Union[ipaddress.IPv4Network, ipaddress.IPv6Network]] = None
+    target_should_contain: Optional[str] = None
```
bulkowy, line 17: I think, it would be better as a `List[str]`, as it could open up more robust filters on targets

bulkowy (approved):

Issue #495, opened 2023-08-23
Report that the domain will soon expire


kazet:

kazet:
e.g. if a domain has less than 7 days to expiration, a report gets generated

Subtasks:
- a module that detects expired domains using whois
- a module that creates human-readable reports about such cases using the reporting mechanism
- an e2e test, maybe mocking the actual whois call (inspired e.g. by https://github.com/CERT-Polska/Artemis/blob/main/test/reporting/test_bruter_autoreporter_integration.py)

PR #544, opened 2023-09-26
Report that the domain will soon expire #495

anna1492:


artemis/config.py
```diff
@@ -404,6 +404,11 @@ class WordPressBruter:
                 "www123, when testing www.projectname.example.com.",
             ] = get_config("WORDPRESS_BRUTER_STRIPPED_PREFIXES", default="www", cast=decouple.Csv(str))
 
+        class DomainExpirationScanner:
+            DOMAIN_EXPIRATION_ALERT_IN_DAYS: Annotated[
```
kazet, line 408: I would prefer `DOMAIN_EXPIRATION_TIMEFRAME_DAYS` (for other variables we write the unit as a direct suffix, not 'in days')

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,52 @@
...
+
+    identity = "domain_expiration_scanner"
+    filters = [{"type": TaskType.DOMAIN.value}]
+
+    def run(self, current_task: Task) -> None:
+        domain = current_task.get_payload(TaskType.DOMAIN)
+        if is_main_domain(domain):
+            now = datetime.datetime.now()
+            result: Dict[str, Any] = {}
+            domain_data = query(domain)
```
kazet, line 27: Did you check:

- what are the rate limits for whois? 
- do they differ between registries?
- what pywhois does if limits are hit?

artemis/reporting/modules/domain_expiration_scanner/template_close_domain_expiration_scanner.jinja2
```diff
@@ -0,0 +1,13 @@
+{% if "close_domain_expiration_date" in data.contains_type %}
+    <li>{% trans %}The following addresses will soon expire: {% endtrans %}
```
kazet, line 2: domains

artemis/reporting/modules/domain_expiration_scanner/template_close_domain_expiration_scanner.jinja2
```diff
@@ -0,0 +1,13 @@
...
+        <ul>
+            {% for report in data.reports %}
+                {% if report.report_type == "close_domain_expiration_date" %}
+                    <li>
+                        <p> {{ report.target }} - {{ report.additional_data }} {% trans %}days{% endtrans %}</p>
```
kazet, line 7: additional_data is a dict for other modules, let's keep it consistent (key can be e.g. "expiry_date")

artemis/reporting/modules/domain_expiration_scanner/reporter.py
```diff
@@ -0,0 +1,45 @@
...
+            return []
+
+        additional_data = task_result["result"]
+
+        return [
+            Report(
+                top_level_target=get_top_level_target(task_result),
+                target=f"http://{task_result['payload']['domain']}",
+                report_type=DomainExpirationScannerReporter.CLOSE_DOMAIN_EXPIRATION_DATE,
+                additional_data=additional_data["days_to_expire"],
```
kazet, line 33: don't write days to expiry, as there may be a large difference betweeen scanning and sending the e-mail. Instead write expiry date

artemis/reporting/modules/domain_expiration_scanner/reporter.py
```diff
@@ -0,0 +1,45 @@
...
+
+        if not isinstance(task_result["result"], dict):
+            return []
+
+        additional_data = task_result["result"]
+
+        return [
+            Report(
+                top_level_target=get_top_level_target(task_result),
+                target=f"http://{task_result['payload']['domain']}",
```
kazet, line 31: the report is about a domain, let's make the target a domain

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,52 @@
...
+                result["close_expiry_date"] = True
+                result["days_to_expire"] = days_to_expire
+                status = TaskStatus.INTERESTING
+                status_reason = (
+                    f"Scanned domain will expire in {days_to_expire} days."
+                    if days_to_expire != 1
+                    else f"Scanned domain will expire in {days_to_expire} day."
+                )
+            else:
+                status = TaskStatus.OK
```
kazet, line 46: save TaskStatus.OK always if you don't save INTERESTING. Currently you save this status only if `is_main_domain` returns true.

artemis/modules/domain_expiration_scanner.py
anna1492: - Whois limits are not oficially disclosed. 
- They may differ between registries - for .com there is no confirmed information, for .pl - https://www.dns.pl/en/whois/FAQ (from a single IP address - 100 per day)
- I assume that you're asking about library used here. It is python-whois, which is a wrapper around system's whois. This library does not add its own limits - it just passes on all requests to whois.

artemis/reporting/modules/domain_expiration_scanner/reporter.py
anna1492: I think we should fix this (the way the project code works in this regard).

kazet: what do you mean - what should be fixed?

artemis/reporting/modules/domain_expiration_scanner/translations/pl_PL/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,7 @@
+#: artemis/reporting/modules/domain_expiration_scanner/template_close_domain_expiration_scanner.jinja2:2
+msgid "The following domains will soon expire: "
+msgstr "Poniższe domeny wkrótce wygasną: "
+
+#: artemis/reporting/modules/domain_expiration_scanner/template_close_domain_expiration_scanner.jinja2:7
+msgid " will expire on "
```
kazet, line 6: you translate a version with spase to a version without spaces, that will cause typos

artemis/reporting/modules/domain_expiration_scanner/reporter.py
```diff
@@ -0,0 +1,74 @@
...
+        data = task_result["result"]
+        exp_date = data["expiration_date"]
+        expiration_date = exp_date.strftime("%d-%m-%Y")
+
+        return [
+            Report(
+                top_level_target=get_top_level_target(task_result),
+                target=task_result["payload"]["domain"],
+                report_type=DomainExpirationScannerReporter.CLOSE_DOMAIN_EXPIRATION_DATE,
+                additional_data={"exp_date": expiration_date},
```
kazet, line 40: IMO exp_date -> expiry_date, no reason to shorten it

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,52 @@
...
+            now = datetime.datetime.now()
+            domain_data = query(domain)
+            expiry_date = domain_data.expiration_date
+            days_to_expire = None
+            if expiry_date:
+                days_to_expire = (expiry_date - now).days
+            result["expiration_date"] = expiry_date
+            if (
+                days_to_expire
+                and days_to_expire <= Config.Modules.DomainExpirationScanner.DOMAIN_EXPIRATION_TIMEFRAME_DAYS
```
kazet, line 35: what will be the value of status if days_to_Expire > TIMEFRAME but is_main_domain?

artemis/modules/domain_expiration_scanner.py
kazet: If for .pl the limit is 100 per day, how can we handle the case of timeouts?

artemis/reporting/modules/domain_expiration_scanner/reporter.py
```diff
@@ -0,0 +1,74 @@
...
+            return []
+
+        if not task_result["status"] == "INTERESTING":
+            return []
+
+        if not isinstance(task_result["result"], dict):
+            return []
+
+        data = task_result["result"]
+        exp_date = data["expiration_date"]
```
kazet, line 32: if the variables differ by type, that should be singified not by their shortened name, but by e.g. suffix

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,52 @@
...
+            ):
+                result["close_expiry_date"] = True
+                result["days_to_expire"] = days_to_expire
+                status = TaskStatus.INTERESTING
+                status_reason = (
+                    f"Scanned domain will expire in {days_to_expire} days - {expiry_date}."
```
kazet, line 41: `- {expiry_date}` -> `(on {expiry_date})`

artemis/reporting/modules/domain_expiration_scanner/reporter.py
anna1492: We cannot say we can have URL or domain as target if we cannot handle both cases. Both cases should work.

artemis/config.py
```diff
@@ -404,6 +404,11 @@ class WordPressBruter:
                 "www123, when testing www.projectname.example.com.",
             ] = get_config("WORDPRESS_BRUTER_STRIPPED_PREFIXES", default="www", cast=decouple.Csv(str))
 
+        class DomainExpirationScanner:
+            DOMAIN_EXPIRATION_TIMEFRAME_DAYS: Annotated[
+                int, "The scanner warns if the domain's expiration date falls within this time frame from now."
+            ] = get_config("DOMAIN_EXPIRATION_ALERT_IN_DAYS", default=5, cast=int)
```
kazet, line 410: you changed in only one place

artemis/reporting/modules/domain_expiration_scanner/reporter.py
kazet: does https://github.com/CERT-Polska/Artemis/pull/544/commits/a1eb8936cf394d99a41b90bc9ff30ab8b40ba7b3 solve the problem?

artemis/modules/domain_expiration_scanner.py
anna1492: We have two things to manipulate - 1. querying cache (available directly in python-whois) - it defaults to 48 h 2. waiting time - to reverse a ban - after using all requests, we get a particular exception, so we can wait a certain amount of time - in this case probably 24h + 30 min (they grant credits every ca.15 min so it's safe to add 30 min). But it seems that they count time after 00:00 as a day.

artemis/reporting/modules/domain_expiration_scanner/reporter.py
anna1492: We cannot say we can have URL or domain as target if we cannot handle both cases. Both cases should work.

kazet: you didn't respond whether a1eb893 solves the problem ;)

anna1492: Workaround is still a workaround even if it works and is still worth fixing.

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,62 @@
...
+                result["expiration_date"] = expiry_date
+                if (
+                    days_to_expire
+                    and days_to_expire <= Config.Modules.DomainExpirationScanner.DOMAIN_EXPIRATION_TIMEFRAME_DAYS
+                ):
+                    result["close_expiry_date"] = True
+                    result["days_to_expire"] = days_to_expire
+                    status = TaskStatus.INTERESTING
+                    status_reason = (
+                        f"Scanned domain will expire in {days_to_expire} days - {expiry_date}."
```
kazet, line 43: (on {expiry_date})

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,62 @@
...
+                        if days_to_expire != 1
+                        else f"Scanned domain will expire in {days_to_expire} day - (on {expiry_date})."
+                    )
+                else:
```
kazet, line 47: This code can be made less complex if you set status and status_reason at the beginning and save_task_Result at the end - therefore you wouldn't have to repeat l48-l50

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,62 @@
...
+                    status = TaskStatus.OK
+                    status_reason = None
+                    self.db.save_task_result(task=current_task, status=status, status_reason=status_reason, data=result)
+
+            except WhoisQuotaExceeded:
+                time.sleep(24 * 60 * 60)
```
kazet, line 53: you sleep but you don't repeat whois query ;)

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,76 @@
...
+                expiry_date = domain_data.expiration_date
+                result = self._prepare_expiration_data(expiration_date=expiry_date, result=result)
+                status = TaskStatus.INTERESTING
+                status_reason = self._prepare_expiration_status_reason(
+                    days_to_expire=result["days_to_expire"], expiration_date=result["expiration_date"]
+                )
+
+            except WhoisQuotaExceeded:
+                time.sleep(24 * 60 * 60)
+                domain_data = self._query_whois(domain=domain)
```
kazet, line 40: now you repeat lines  30-36, IMO that's too much code to be repeated

artemis/modules/domain_expiration_scanner.py
```diff
@@ -0,0 +1,76 @@
...
+        domain = current_task.get_payload(TaskType.DOMAIN)
+        result: Dict[str, Any] = {}
+        status = TaskStatus.OK
+        status_reason = None
+        if is_main_domain(domain):
+            try:
+                domain_data = self._query_whois(domain=domain)
+                expiry_date = domain_data.expiration_date
+                result = self._prepare_expiration_data(expiration_date=expiry_date, result=result)
+                status = TaskStatus.INTERESTING
```
kazet, line 33: now you set TaskStatus.INTERESTING always, not only if the domain is close to expiry date

kazet:
testy Ci nie przechozą

kazet (approved):

PR #565, opened 2023-10-06
Fixing time-based sqli template

kazet:


msm-code (approved):
Both versions look good, and I thought the first version was to ensure that not every requests takes 10+ seconds? But I guess you've checked it works.

PR #585, opened 2023-10-18
Blocklist blocks scanning, not only reporting

kazet:


artemis/blocklist.py
```diff
@@ -0,0 +1,182 @@
...
+def load_blocklist(file_path: Optional[str]) -> List[BlocklistItem]:
+    if not file_path:
+        return []
+
+    with open(file_path, "r") as file:
+        data = yaml.safe_load(file)
+
+    for item in data:
+        # Assert there are no additional or missing keys
+        unexpected_keys = set(item.keys()) - {
```
bulkowy, line 56: put differenced set as a variable, no need for it to be initialized every item

artemis/module_base.py
```diff
@@ -172,6 +186,33 @@ def _consume_random_routed_task(self, identity: str) -> Optional[Task]:
                 return task
         return None
 
+    def _is_blocklisted(self, task: Task) -> bool:
+        host = get_target_host(task)
+
+        if is_domain(host):
+            domain = host
```
bulkowy, line 193: why though, I would use same arg, as it is the same

artemis/module_base.py
```diff
@@ -172,6 +186,33 @@ def _consume_random_routed_task(self, identity: str) -> Optional[Task]:
...
+            if ip_addresses:
+                for ip in ip_addresses:
+                    if should_block_scanning(
+                        domain=domain, ip=ip, karton_name=self.identity, blocklist=self._blocklist
+                    ):
+                        return True
+            else:
+                if should_block_scanning(domain=domain, ip=None, karton_name=self.identity, blocklist=self._blocklist):
+                    return True
+        else:
```
bulkowy, line 209: `elif is_ip_address(host)`, and then delete assert below

bulkowy (commented):
Generally, ok. But I would opt for putting blocklist in some form of DB with easy key retrieval. This would allow code to be simplified, eg `blocklist_reports` could use only one for loop, now it is really O(N*M)

bulkowy (dismissed):

bulkowy (approved):

PR #600, opened 2023-10-31
Changing the logic in the requests configuration

anna1492:


artemis/config.py
```diff
@@ -129,12 +129,12 @@ class Limits:
             "E.g. when set to 100, Artemis will send no more than 100 port scanning packets per seconds per port scanner instance.",
         ] = get_config("SCANNING_PACKETS_PER_SECOND", default=100, cast=int)
 
-        SECONDS_PER_REQUEST: Annotated[
+        REQUEST_PER_SECOND: Annotated[
```
kazet, line 132: requests_per_second

artemis/config.py
```diff
@@ -129,12 +129,12 @@ class Limits:
...
             int,
             """
-            E.g. when set to 2, Artemis will make sure no HTTP/MySQL connect/... request takes less than 2 seconds, sleeping if needed.
+            E.g. when set to 2, Artemis will make sure no more HTTP/MySQL connect/... requests take place per second, sleeping if needed.
```
kazet, line 135: no more than ...

artemis/utils.py
```diff
@@ -43,13 +44,27 @@ def is_directory_index(content: str) -> bool:
...
+        average_time_per_request = 1 / request_per_second
+        f_start = time.time()
+        try:
+            return f()
+        finally:
+            func_time = time.time() - f_start
+            if func_time < average_time_per_request:
+                time.sleep(average_time_per_request - func_time)
+    elif request_per_second < 1:
+        seconds_for_req = math.floor(1 / request_per_second)
```
kazet, line 60: czemu floor? jak ma byc 0.75 requesta per second to sleep 1/0.75 i zalatwione

artemis/modules/nuclei.py
```diff
@@ -104,7 +104,9 @@ def run_multiple(self, tasks: List[Task]) -> None:
             "-headless-bulk-size",
             str(len(tasks_filtered)),
             "-milliseconds-per-request",
-            str(int(Config.Limits.SECONDS_PER_REQUEST * 1000.0 / len(tasks_filtered))),
+            str(int((1 / Config.Limits.REQUEST_PER_SECOND) * 1000.0 / len(tasks_filtered)))
+            if Config.Limits.REQUEST_PER_SECOND != 0
+            else str(int(0)),
```
kazet, line 109: why int(

artemis/utils.py
```diff
@@ -43,13 +43,27 @@ def is_directory_index(content: str) -> bool:
...
-    try:
+    request_per_second = Config.Limits.REQUESTS_PER_SECOND
+    if request_per_second == 0:
         return f()
-    finally:
-        time_elapsed = time.time() - time_start
-        if time_elapsed < Config.Limits.SECONDS_PER_REQUEST:
-            time.sleep(Config.Limits.SECONDS_PER_REQUEST - time_elapsed)
+    elif request_per_second >= 1:
+        average_time_per_request = 1 / request_per_second
```
kazet, line 50: now you have two duplicate if branches

artemis/modules/nuclei.py
kazet: bump

artemis/utils.py
anna1492: Assuming int, now is float - so it can be that way.

artemis/modules/nuclei.py
anna1492: I changed the value here, not the structure. It was earlier here, and I assumed it had its purpose. But if you change the structure, it won't work (nuclei takes int), and tests will (obviously) not pass in one of our config conditions. So - do you propose something specific here, or you just asked?

kazet: I meant that 'else str(int(0))' can be reduced to 'else str(0)'

kazet (approved):

PR #615, opened 2023-11-05
more sql dumps

kazet:


artemis/reporting/modules/bruter/classifier.py
```diff
@@ -92,7 +92,15 @@ def is_sql_dump(found_url: FoundURL) -> bool:
     if "sql dump" in found_url.content_prefix.lower():
         return True
 
-    if "\ncreate table" in found_url.content_prefix.lower():
+    if any(
+        [
+            # strip() to allow lines starting with whitespace
+            line.strip().startswith("create table")
+            or line.strip().startswith("alter table")
+            or line.strip().startswith("insert into")
```
bulkowy, line 100: I would put identifier strings in list variable and put these `line.strip()` conditions in `any([line.strip().startswith(s) for s in strings])`

PR #650, opened 2023-11-24
Now we can block only subdomains

kazet:


artemis/blocklist.py
```diff
@@ -91,10 +94,16 @@ def should_block_scanning(
...
-        if item.domain:
+        if item.domain_and_subdomains:
             if not domain:
                 continue
-            if not is_subdomain(domain, item.domain):
+            if not is_subdomain(domain, item.domain_and_subdomains):
+                continue
+
+        if item.subdomains:
+            if not domain:
```
bulkowy, line 104: you can flatten 3 if statements
`if not domain`
`if item.domain_and_subdomains and not is_subdomain(...)`
`if item.subdomains and not is_subdomain(...)`

PR #658, opened 2023-12-01
Drupal check

kazet:


artemis/modules/base/base_newer_version_comparer.py
```diff
@@ -0,0 +1,58 @@
...
+
+class BaseNewerVersionComparerModule(ArtemisBase):
+    software_name: str
+
+    def __init__(self, *args, **kwargs):  # type: ignore
+        super().__init__(*args, **kwargs)
+
+        self.endoflife_data_folder = tempfile.mkdtemp()
+        subprocess.call(
+            ["git", "clone", "https://github.com/endoflife-date/endoflife.date", self.endoflife_data_folder]
```
bulkowy, line 20: I would put this as submodule, not explicitely cloning this everytime

artemis/modules/base/base_newer_version_comparer.py
```diff
@@ -0,0 +1,58 @@
...
+            return semver.VersionInfo.parse(version)
+
+    def is_version_obsolete(
+        self,
+        version: str,
+    ) -> bool:
+        version_parsed = self._parse_version(version)
+
+        for release in self.endoflife_data["releases"]:
+            if not (version == release["releaseCycle"] or version.startswith(release["releaseCycle"] + ".")):
```
bulkowy, line 40: wouldnt `version.startswith(release["releaseCycle"])` be enough?

artemis/modules/base/base_newer_version_comparer.py
```diff
@@ -0,0 +1,58 @@
...
+                continue
+
+            # Semver compare returns 1 if the latter version is greater, 0 if they are equal, and -1 if
+            # the latter version is smaller.
+            comparison_result = semver.VersionInfo.parse(release["latest"]).compare(version_parsed)
+            if comparison_result > 0:
+                return True
+            elif comparison_result == 0:
+                return release["eol"] <= datetime.datetime.now().date()  # type: ignore
```
bulkowy, line 49: should not it be in other way? EOL should be latter than now()

artemis/modules/drupal_scanner.py
```diff
@@ -0,0 +1,66 @@
...
+            if "/core/misc/drupal.js?v=" in script.get("src", ""):
+                version = script["src"].split("=")[1]
+
+        if version:
+            is_version_obsolete = super().is_version_obsolete(
+                version,
+            )
+        else:
+            is_version_obsolete = None
+        result = {
```
bulkowy, line 44:
```suggestion

        result = {
```

artemis/modules/drupal_scanner.py
```diff
@@ -0,0 +1,66 @@
...
+            "version": version,
+            "is_version_obsolete": is_version_obsolete,
+        }
+
+        if version and is_version_obsolete:
+            found_problems = [f"Drupal version {version} on {url} is obsolete"]
+        elif not version:
+            found_problems = [f"Unable to obtain Drupal version for {url}"]
```
bulkowy, line 52: not sure if it is something to log, it always can be not a drupal app

artemis/reporting/modules/drupal_scanner/template_old_drupal.jinja2
```diff
@@ -0,0 +1,21 @@
...
+                        {{ report_meta(report) }}
+                    </li>
+                {% endif %}
+            {% endfor %}
+        </ul>
+        <p>
+            {% trans trimmed %}
+                If a site is no longer used, we recommend shutting it down to eliminate the risk
+                of exploitation of known vulnerabilities in older Drupal versions. Otherwise, we
+                recommend regular Drupal core and plugin updates.
```
bulkowy, line 17:
```suggestion
                recommend following regular Drupal core and plugin updates.
```

artemis/modules/base/base_newer_version_comparer.py
kazet: added comment

artemis/modules/base/base_newer_version_comparer.py
kazet: nope, because the function says whether the version is obsolete

artemis/modules/drupal_scanner.py
kazet: made this an exception as this may mean that the check is broken. This karton takes only identified Drupal apps

kazet: made this an exception as this may mean that the check is broken. This karton takes only identified Drupal apps

artemis/reporting/modules/drupal_scanner/template_old_drupal.jinja2
kazet: why is this one better? IMO `update` has two meanings, "what you do on your server" and "what appears on drupal release list" and both are OK and what it's now in the repo is simpler

artemis/modules/base/base_newer_version_comparer.py
kazet: we do want to do this every container start so that we always have the latest version

bulkowy: so put it in Dockerfile

kazet: if we put it in Dockerfile, we would re-download this on all container **rebuilds**, not restarts - and currently the containers restart every 200 tasks, so e.g. Nuclei templates get redownloaded - I want the same behavior with endoflife

bulkowy (approved):

PR #668, opened 2023-12-07
Brute force tool for administration panels

anna1492:


artemis/modules/admin_panel_login_bruter.py
```diff
@@ -0,0 +1,205 @@
...
+        if not is_root_url and not any([item in url.lower() for item in ["login", "admin", "cms", "backend", "panel"]]):
+            self.db.save_task_result(
+                task=task,
+                status=TaskStatus.OK,
+                status_reason=f"URL {url} doesn't look like an admin panel URL",
+                data=None,
+            )
+            return
+
+        if "admin" in url and not url.endswith("/admin") and "admin" in task.get_payload("found_urls"):
```
bulkowy, line 69: I don't get this check and task result reason doesn't give clear indication what it is about

artemis/modules/admin_panel_login_bruter.py
```diff
@@ -0,0 +1,205 @@
...
+            user_input.send_keys(username)
+        if password_input:
+            password_input.send_keys(password)
+            password_input.send_keys(Keys.ENTER)
+
+    @staticmethod
+    def _get_logging_in_result(driver: WebDriver, login_failure_msgs: list[str]) -> Optional[list[str]]:
+        try:
+            web_content = driver.find_element(By.XPATH, "html/body").text
+            print(web_content)
```
bulkowy, line 197:
```suggestion
```

artemis/passwords.py
```diff
@@ -0,0 +1,41 @@
...
+        while domain_items and domain_items[0] in Config.Modules.WordPressBruter.WORDPRESS_BRUTER_STRIPPED_PREFIXES:
+            domain_items = domain_items[1:]
+
+        if domain_items:
+            site_name = domain_items[0]
+
+            passwords.append(site_name + "123")
+            passwords.append(site_name + "1")
+            for year_relative in [0, -1, -2, -3]:
+                passwords.append(site_name + str(datetime.datetime.now().year + year_relative))
```
bulkowy, line 40: Why not `[0, 1, 2, 3]` and `- year_relative`, seems more intuitive for me

artemis/modules/admin_panel_login_bruter.py
```diff
@@ -0,0 +1,247 @@
...
+                except TimeoutException:
+                    self.log.info(
+                        "Timeout occured when waiting for the URL to match, let's try "
+                        f"to login even if the url doesn't match, url={driver.current_url}"
+                    )
+
+                # Ignore alerts
+                driver.execute_script("window.alert = function() {};")  # type: ignore
+                driver.implicitly_wait(Config.Modules.AdminPanelLoginBruter.WAIT_TIME_SECONDS)
+                self._sleep_after_performing_requests(driver)
```
bulkowy, line 126: we are waiting and sleeping in the same time? I think that waiting would be sufficient

artemis/modules/admin_panel_login_bruter.py
```diff
@@ -0,0 +1,247 @@
...
+                data=credentials,
+            )
+        else:
+            self.db.save_task_result(task=task, status=TaskStatus.OK, status_reason=None, data=None)
+
+    def _brute(self, url: str) -> Optional[Tuple[str, str]]:
+        working_credentials = []
+        for username in self.USERNAMES:
+            for password in get_passwords_for_url(url):
+                driver = AdminPanelLoginBruter._get_webdriver()
```
bulkowy, line 110: Would not it be better to get only one webdriver for whole `_brute` function and just drop old cookies every 
credential? Now it will be resource consuming

artemis/passwords.py
```diff
@@ -0,0 +1,41 @@
...
+
+from artemis import utils
+from artemis.config import Config
+
+PASSWORDS = [
+    # Let's keep the most popular password near the top, as some admin panels
+    # block logins after a given number of tries.
+    "123456",
+    "admin",
+    "admin1",
```
bulkowy, line 13: unnecessary entry, as `get_passwords_for_url` appends `1`

artemis/utils.py
```diff
@@ -85,6 +86,25 @@ def get_host_from_url(url: str) -> str:
...
+        port = getservbyname(url_parsed.scheme)
+        url_parsed_dict["netloc"] = url_parsed_dict["netloc"] + ":" + str(port)
+    return urllib.parse.urlunparse(urllib.parse.ParseResult(**url_parsed_dict))
+
+
+def remove_standard_ports_from_url(url: str) -> str:
+    url_parsed = urllib.parse.urlparse(url)
+    url_parsed_dict = url_parsed._asdict()
+    if url_parsed.scheme == "http" and url_parsed.port == 80:
+        url_parsed_dict["netloc"] = url_parsed_dict["netloc"].split(":")[0]
```
bulkowy, line 102: just to note: sometimes URL might have credentials (ik, 99.999% won't but), so it might be better to change it to `[-1]`

PR #705, opened 2024-01-08
Added Humble

cyberamt:
This pull request introduces a new scanning modul to artemis called humble.

humble is an open source HTTP-Header security scanner ([repo](https://github.com/rfc-st/humble))

I have strongly oriented myself to the implementation of dnsreaper. humble.py has the class [Humble](https://github.com/cyberamt/Artemis/blob/33b9ed1f8bf3b6382fcf4dafb2268b947edc311a/artemis/modules/humble.py#L74) which is responsible for the call and additionally the function [process_json_data](https://github.com/cyberamt/Artemis/blob/33b9ed1f8bf3b6382fcf4dafb2268b947edc311a/artemis/modules/humble.py#L13C8-L13C8) which parses the output.

The parser tries to parse missing security headers so that artemis can display them as interesting findings. If necessary, the parser needs a little refinement so that depricated headers and the like are correctly recognized as "interesting".

This pull request was first made for the [artemis-extra-module repo](https://github.com/CERT-Polska/Artemis-modules-extra/pull/63).

Please review the changes and let me know if there are any concerns or suggestions for improvement.


artemis/modules/humble.py
```diff
@@ -0,0 +1,155 @@
...
+            # Check if the key is not in the excluded categories and there are relevant values
+            if (
+                category.lower() != "info"
+                and key
+                not in [
+                    "[2. Fingerprint HTTP Response Headers]",
+                    "[3. Deprecated HTTP Response Headers/Protocols and Insecure Values]",
+                    "[4. Empty HTTP Response Headers Values]",
+                    "[5. Browser Compatibility for Enabled HTTP Security Headers]",
+                ]
```
bulkowy, line 45: This key check could be done before looking directly at category, at LOC 29

artemis/modules/humble.py
```diff
@@ -0,0 +1,155 @@
...
+                        isinstance(value, list)
+                        and any(
+                            subvalue
+                            for subvalue in value
+                            if subvalue
+                            and subvalue
+                            not in ["Nothing to report, all seems OK!", "No HTTP security headers are enabled."]
+                        )
+                    )
+                )
```
bulkowy, line 58:
```suggestion
```
These checks might be unnecessary, as you check for them in if-block

artemis/modules/humble.py
```diff
@@ -0,0 +1,155 @@
...
+        # strip boilerplatetext from the output to get the location and filename of the output file
+        filename = (
+            data.decode("ascii", errors="ignore")
+            .removeprefix("\n Analyzing URL and saving the report, please wait ...\n\n\n Report saved to ")
+            .removesuffix("\n")
+        )
+        data_str = open(filename, "r").read()
+
+        # cleanup file
+        subprocess.run(["rm", filename])
```
bulkowy, line 124: Please use `.unlink()` function from [Pathlib python library](https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink)

bulkowy (approved):

kazet:
@cyberamt : thank you for your contribution <3

cyberamt:
Hi, really cool that you were able to improve the code and have memorized it. I would be happy if you could include the account as "cyberamt" in the contributers. I will try to write more merge requests with tools if I like.

kazet:
Hello,

added you as a contributor, thanks again for the code!

PR #729, opened 2024-01-21
Known exploited vulnerabilities are checked

kazet:


artemis/modules/nuclei.py
```diff
@@ -31,8 +31,14 @@ class Nuclei(ArtemisBase):
 
     def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
+        subprocess.call(["git", "clone", "https://github.com/Ostorlab/KEV/", "/known-exploited-vulnerabilities/"])
```
bulkowy, line 34: Put this repo cloning in docker file, as it should not be done on class initializaiton, I guess

bulkowy (approved):

PR #742, opened 2024-01-24
WordPress plugin version identifier

kazet:


artemis/modules/wordpress_plugins.py
```diff
@@ -0,0 +1,300 @@
...
+                    line.replace("(", " ")
+                    .replace("*", " ")
+                    .replace("[", " ")
+                    .replace("]", " ")
+                    .replace("'", " ")
+                    .replace(":", " ")
+                    .replace(",", " ")
+                    .replace("-", " ")
+                    .replace("<h4>", " ")
+                    .replace("</h4>", " ")
```
bulkowy, line 74: to simplify this, I would use regex

artemis/modules/wordpress_plugins.py
```diff
@@ -0,0 +1,300 @@
...
+
+    # Some plugins have broken "stable tag"
+    if slug in PLUGINS_TO_SKIP_STABLE_TAG:
+        return changelog_version
+
+    tag_lines = [line for line in readme_content.lower().split("\n") if line.strip("* ").startswith("stable tag:")]
+    if len(tag_lines) > 1:
+        return changelog_version
+
+    if len(tag_lines):
```
bulkowy, line 95: I don't get this, before this if statement you check if there are at least 2 tag_lines, then you again check for length of tag_lines. Should not it be just `if tag_lines` or `if len(tag_lines) == 1`?

artemis/modules/wordpress_plugins.py
```diff
@@ -0,0 +1,300 @@
...
+    soup = BeautifulSoup(response.text)
+    links = []
+    for tag in soup.find_all():
+        new_url = None
+        for attribute in ["src", "href"]:
+            if attribute not in tag.attrs:
+                continue
+
+            new_url = urllib.parse.urljoin(url, tag[attribute])
+            new_url_parsed = urllib.parse.urlparse(new_url)
```
bulkowy, line 127: this part was used in finding related links to scanned page. I would make it metafunction, so there won't be repetitions

artemis/modules/wordpress_plugins.py
```diff
@@ -186,6 +173,29 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+
+        plugin_data = []
+        for link in links:
+            pattern = r"\/wp-content\/plugins\/([a-z-0-9]*)\/"
+            match = re.search(pattern, link)
+            if match:
+                slug = match.group(1)
+                data = json.loads(
+                    self.cached_get(f"https://api.wordpress.org/plugins/info/1.0/{slug}.json", "version-" + slug),
+                    timeout=3600,
```
bulkowy: Isn't this timeout too big? 1 hour timeout seems over necessary

kazet: I don't think plugin will get updated that frequently and I don't want to get banned on wp api

bulkowy (dismissed):

artemis/modules/wordpress_plugins.py
```diff
@@ -62,6 +64,28 @@
 ]
 
 
+def _is_version_larger(v1: str, v2: str) -> bool:
```
bulkowy, line 80: what if version has character or string in it, like `v6.1.0a`

artemis/modules/wordpress_plugins.py
```diff
@@ -64,14 +69,30 @@
 ]
 
 
+def strip_trailing_zeros(version: Optional[str]) -> Optional[str]:
+    if not version:
+        return None
+
+    version_split = version.split(".")
+    while len(version_split) > 0 and version_split[-1] == "0":
```
bulkowy, line 75: what if version has more than 1 digit eg. `v6.1.00`

artemis/modules/wordpress_plugins.py
```diff
@@ -64,14 +69,30 @@
...
+        version_split = version_split[:-1]
```
bulkowy, line 76: list.pop() will be better, mainly in readibility

artemis/modules/wordpress_plugins.py
kazet: supporting 000

artemis/modules/wordpress_plugins.py
kazet: we compare numbers and fall back to lexicographical compare if the values aren't numeric

bulkowy (dismissed):

bulkowy (dismissed):

PR #744, opened 2024-01-24
Humble user-agent

kazet:


artemis/modules/humble.py
```diff
@@ -83,6 +84,12 @@ class Humble(ArtemisBase):
     ]
 
     def run(self, current_task: Task) -> None:
+        if Config.Miscellaneous.CUSTOM_USER_AGENT:
+            with open("/humble/additional/user_agents.txt", "w") as f:
+                f.write(f"1.- {Config.Miscellaneous.CUSTOM_USER_AGENT}\n")
+        else:
+            subprocess.call(["git", "checkout", "additional/user_agents.txt"], cwd="/humble")
```
bulkowy, line 91: I don't get it at all, why this list won't be imported into project without need to use git checkout on runtime

bulkowy (approved):

PR #749, opened 2024-01-26
Build docker images for linux/arm64

es1o:
I added support for linux/arm64 architecture used on Apple Silicon.
This image is used by Dockerfiles in https://github.com/CERT-Polska/Artemis-modules-extra that fails on Mac and requires manual, local build.


BonusPlay (approved):
LGTM

PR #784, opened 2024-02-12
not starting shodan if no key

kazet:


artemis/modules/shodan_vulns.py
```diff
@@ -103,5 +103,6 @@ def run(self, current_task: Task) -> None:
 
             with open(no_api_key_message_printed_filename, "w"):
                 pass
+        return
```
bulkowy, line 106: we are not in function, so return doesn't make sense here

bulkowy (approved):

PR #803, opened 2024-02-22
CSRF and removing pending tasks

kazet:


artemis/frontend.py
```diff
@@ -122,18 +136,49 @@ async def post_add(
...
+        csrf_protect,
+    )
+
+
+@router.post("/analysis/remove-pending-tasks/{analysis_id}", include_in_schema=False)
+@csrf.validate_csrf
+async def post_remove_pending_tasks(
+    request: Request, analysis_id: str, csrf_protect: CsrfProtect = Depends()
+) -> Response:
+    db.mark_analysis_as_stopped(analysis_id)
```
bulkowy, line 157: I would swap places, firstly I would stop tasks in queue, then mark analysis as stopped

kazet: why? That would be a race condition - If then after stopping tasks in queue an existing task would finish and add new tasks to queue, we would end up with non-finished analysis

kazet: marking analysis as stopped prevents adding new tasks as well

bulkowy (approved):

PR #815, opened 2024-02-29
Migrating Mongo->PostgreSQL

kazet:


artemis/db.py
```diff
@@ -28,6 +48,72 @@ def as_dict(self) -> Dict[str, Any]:
...
+Session = sessionmaker(bind=engine)
+
+
+class TSVector(TypeDecorator):  # type: ignore
+    impl = TSVECTOR
+
+
+class ScheduledTask(Base):  # type: ignore
+    __tablename__ = "scheduled_task"
+    created_at = Column(DateTime, server_default=text("NOW()"))
```
bulkowy, line 64:
```suggestion
    created_at = Column(DateTime, default=datetime.datetime.now())
```

artemis/db.py
```diff
@@ -28,6 +48,72 @@ def as_dict(self) -> Dict[str, Any]:
...
+    analysis_id = Column(String, primary_key=True)
+    deduplication_data = Column(String, primary_key=True)
+    deduplication_data_original = Column(String)
```
bulkowy, line 67: why some data is primary_key to table, and why there are two fields related to similar thing?

artemis/db.py
```diff
@@ -28,6 +48,72 @@ def as_dict(self) -> Dict[str, Any]:
...
+    task_id = Column(String)
+
+
+class Analysis(Base):  # type: ignore
+    __tablename__ = "analysis"
+    id = Column(String, primary_key=True)
+    created_at = Column(DateTime, server_default=text("NOW()"))
```
bulkowy, line 74: like above - use datetime.now()

artemis/db.py
```diff
@@ -28,6 +48,72 @@ def as_dict(self) -> Dict[str, Any]:
...
+    )
+
+    __table_args__ = (Index("analysis_fulltext", fulltext, postgresql_using="gin"),)
+
+
+class TaskResult(Base):  # type: ignore
+    __tablename__ = "task_result"
+    id = Column(String, primary_key=True)
+    analysis_id = Column(String, index=True)
+    created_at = Column(DateTime, server_default=text("NOW()"))
```
bulkowy, line 92: like above, use `datetime.now()`

artemis/db.py
```diff
@@ -1,14 +1,34 @@
...
+    Boolean,
+    Column,
+    Computed,
+    DateTime,
+    Index,
+    String,
+    create_engine,
+)
+from sqlalchemy.dialects.postgresql import TSVECTOR
+from sqlalchemy.dialects.postgresql import insert as postgres_upsert
```
bulkowy, line 23: nah, leave it as `insert` because PSQL doesnt have `upsert` method

artemis/db.py
```diff
@@ -60,62 +146,77 @@ def get_task_target(task: Task) -> str:
...
-                )
-                if result.upserted_id:  # If the record has been created, set creation date
-                    result = self.task_results.update_one(
-                        {"_id": created_task_result["uid"]}, {"$set": {"created_at": datetime.datetime.now()}}
-                    )
+            to_save["result"] = data
+
+        statement = postgres_upsert(TaskResult).values([copy.copy(to_save)])
+        del to_save["id"]
+        statement = statement.on_conflict_do_update(index_elements=[TaskResult.id], set_=to_save)
```
bulkowy, line 203: I am not sure about this bit: I know you want to provide upsert capability to postgres as per docs, but I don't get why you are deleting `id` key

artemis/db.py
```diff
@@ -125,23 +226,28 @@ def get_paginated_analyses(
...
+        with Session() as session:
+            records_count_total = session.query(Analysis).count()
+
+            if search_query:
+                query = select(Analysis.c.fulltext.match(self._to_postgresql_query(search_query)))
+            else:
+                query = session.query(Analysis)
+
+            records_count_filtered: int = query.count()  # type: ignore
+            results_page = [item.__dict__ for item in query.order_by(*ordering_postgresql)[start : start + length]]  # type: ignore
```
bulkowy, line 245: either use `.pagination()` or implement `after id` pagination

artemis/db.py
```diff
@@ -154,33 +260,47 @@ def get_paginated_task_results(
...
+
+            if analysis_id:
+                query = query.filter(TaskResult.analysis_id == analysis_id)  # type: ignore
+
+            if task_filter:
+                for key, value in task_filter.as_dict().items():
+                    query = query.filter(getattr(TaskResult, key) == value)  # type: ignore
+
+            records_count_filtered = query.count()
+            results_page = [item.__dict__ for item in query.order_by(*ordering_postgresql)[start : start + length]]  # type: ignore
```
bulkowy, line 286: like above

artemis/main.py
```diff
@@ -17,7 +17,7 @@
 
 # We run it here so that it will get executed even when importing from main,
 # which will happen when running the app via `uvicorn artemis.main:app`
-db.initialize_database()
+db_migration.migrate_and_start_thread()
```
bulkowy, line 20: There should some check whether migration from mongo has already happened

artemis/db.py
kazet: This will evaluate now() only once, when the class is created. What I proposed will create a timestamp once a row is added. This will be converted to "DEFAULT NOW()" in SQL, so we won't need to provide the date on INSERT.

artemis/db.py
kazet: This will evaluate now() only once, when the class is created. What I proposed will create a timestamp once a row is added. This will be converted to "DEFAULT NOW()" in SQL, so we won't need to provide the date on INSERT.

artemis/main.py
kazet: This  call will start migrating all unmigrated rows. It is needed as there might be a race condition - new web is started but old kartons keep writing to MongoDB. These subsequent migrations will happen every 20 seconds and will be very fast if nothing is to be migrated

artemis/db.py
kazet: I don't want to set the ID, as it's already set.

artemis/db.py
kazet: I've added a comment

artemis/db.py
kazet: right, done

bulkowy (approved):

PR #855, opened 2024-03-20
Added FptPortScanner

christinathucanh:
* I have named the module FtpPortScanner to reflect its purpose of scanning for open FTP ports.
* The run method performs the scanning logic. It retrieves the target IP from the current task payload, scans for open FTP ports using the scan_ftp_ports method, and creates tasks for each detected open port.
* The scan_ftp_ports method represents the hypothetical port scanning logic for FTP ports. In a real scenario, this method would use a port scanning tool or library to identify open FTP ports on the target IP.
* I have defined a list of open FTP ports detected by the scanner and returned it as the scan result.
* Task status and status reasons are updated based on the scan results, and the results are saved using the db.save_task_result method.

### DIFF
--- a/artemis/modules/FptPortScanner.py
+++ b/artemis/modules/FptPortScanner.py
@@ -1,0 +1,50 @@
+# FtpPortScanner.py
+
+from typing import List
+from artemis.module_base import ArtemisBase
+from artemis.task import Task
+from artemis.binds import Service, TaskStatus, TaskType
+
+class FtpPortScanner(ArtemisBase):
+    identity = "ftp_port_scanner"
+    filters = [
+        {"type": TaskType.IP.value},
+    ]
+
+    def run(self, current_task: Task) -> None:
+        target_ip = current_task.payload.get("ip")
+        
+        # Perform FTP port scanning logic here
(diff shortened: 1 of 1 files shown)


kazet:
Hello,

we already have port scanning capabilities, which detect (among others) FTP ports (https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/port_scanner.py). We can also bruteforce easy FTP passwords: https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/ftp_bruter.py

kazet:
Hello,

we already have port scanning capabilities, which detect (among others) FTP ports (https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/port_scanner.py). We can also bruteforce easy FTP passwords: https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/ftp_bruter.py

kazet:
Hello,

we already have port scanning capabilities, which detect (among others) FTP ports (https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/port_scanner.py). We can also bruteforce easy FTP passwords: https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/ftp_bruter.py

PR #856, opened 2024-03-20
Validating entered data, not silently stripping port numbers or URLs

kazet:


artemis/modules/classifier.py
```diff
@@ -30,33 +29,27 @@ class Classifier(ArtemisBase):
...
-                "Defanged URL detected. If you really want to scan it, please provide it as a standard one."
-            )
-
-        # strip URL schemes
-        if "://" in data:
-            hostname = urllib.parse.urlparse(data).hostname
-            assert hostname is not None
-            data = hostname
+        # Strip leading and trailing dots (e.g. domain.com.)
+        data = data.strip(".")
```
bulkowy, line 33: isn't it URL that is being passed there? If yes, you should retrieve hostname eg using urlparse

kazet: nope

previously Artemis silently converter URLs and host:port strings to domains

now Artemis explicitely says that these aren't allowed and you need to provide ip/domain

bulkowy (approved):

PR #861, opened 2024-03-22
Added smtp scanner

christinathucanh:
SMTP Scanner scans common port ranges (1-1024) on a given target IP for open SMTP ports. For each open SMTP port found, it creates a new SERVICE task with the appropriate payload. Finally, it saves the scan results and task status to the database.


kazet:
Hello! As I said earlier, Artemis already has port scanning capabilities - would you describe what improvement this module provides over existing tools?

PR #865, opened 2024-03-24
Nuclei automation

kshitijk4poor:
In the updated code:

- The links_per_task and links variables are only populated if the -as flag is not set (NUCLEI_AUTOMATIC_SCAN is False).
- The scanning of homepage links (NUCLEI_TEMPLATES_TO_RUN_ON_HOMEPAGE_LINKS) is only performed if the -as flag is not set.
- When determining the relevant findings for each task, the links_per_task is only considered if the -as flag is not set.
- Modified the `_scan` method in `artemis/modules/nuclei.py` to accept an `auto_scan` boolean parameter. When `auto_scan` is `True`, the `-as` flag is added to the Nuclei command to enable automatic scanning based on technology detection.

With these changes, when the -as flag is set (NUCLEI_AUTOMATIC_SCAN is True), the module will only scan the target URLs directly, without retrieving and scanning additional links from the homepage.

You can then set the NUCLEI_AUTOMATIC_SCAN option to True in the configuration file or through an environment variable to enable the automatic scan behavior.


kazet:
Let me know as soon as the code is ready for review

Issue #307, opened 2023-04-19
Check outgoing links: whether they work (no 404), whether the destination uses SSL, whether no bad reputation sites are linked


kazet:

kshitijk4poor:
Can be fixed with httpx.

github.com/projectdiscovery/httpx

kazet:
This solves only a part of the task - downloading the content. How would you e.g. solve the reputation problem?

RasenRhino:
cant we use something like [this](https://urlscan.io/docs/api/) ?

kazet:
I think a good first step would be to collect a couple of sites that provide reputation information (and check what they consider bad websites) and only then compare and choose the best one

Sahityaaryan:
Hi @kazet , I've been working on issue #307 (outgoing links checker). Before opening a PR, I wanted to get your input on a few design decisions:


1. Reputation provider

In the issue discussion you mentioned "a good first step would be to collect a couple of sites that provide reputation information and then compare and choose the best one." Here's what I found:

    Google Safe Browsing — Free API key, 10k lookups/day. Covers malware, phishing, social engineering, unwanted software. Well-documented, widely used.
    VirusTotal — Free API key, but only 500 lookups/day (4/min rate limit). Aggregates 70+ scanners — great coverage but too rate-limited for production scanning.
    urlscan.io — Free API key, 5k lookups/day. Good for phishing/malware, provides detailed page analysis but slower per-request.
    PhishTank — Free API key, unlimited lookups. Phishing only — too narrow in scope.

I went with Google Safe Browsing because it has the broadest threat coverage, a generous free tier, and is well-documented. The API key is optional — if not configured, reputation checking is simply skipped.

Would you prefer a different provider, or should we support multiple?


2. Crawling depth for link extraction

Currently the module only checks the root page for outgoing links. For the leak_scanner we added BFS crawling — should I do the same here, or is single-page extraction sufficient since outgoing links tend to be on the homepage/main pages?

3. Non-SSL reporting

The issue mentions checking "whether the destination uses SSL." I've implemented this, but in practice most sites have many HTTP outgoing links, which could generate a lot of noise. Should non-SSL links:

    (a) Be reported as a separate finding (current implementation, severity LOW)
    (b) Be included only as informational data without generating individual reports
    (c) Be skipped for now and added later if needed

4. Severity levels

[...]

PR #878, opened 2024-03-27
URL reputation module 

RasenRhino:
I have opened this PR for code review and further improvements. This is an abstract idea on how I plan to proceed with this [issue](https://github.com/CERT-Polska/Artemis/issues/307) .  
testing it alongside. 

TODOS:

- [ ]  add the data to the database part 
- [ ]  check for the urls (the no 404 is relatively easy so have left it for last) 

The test script that has been translated to this module can be looked at [here](https://gist.github.com/RasenRhino/5e9327cbc544d5d591f5eee6996dc1ed)
one can probably use [this](https://github.com/RasenRhino/bad_url_website) for testing purposes. just a simple html website with a few bad urls which I was using while developing this script. just serve it through any live server and point it to the base url.


kazet:
The overall logic seems fine. The main two things to improve would be:
- add reporting capabilities, so that malicious outgoing links would be included in automatic reports generated by Artemis
- research what methods of identifying bad URLs are available - the one you chose detects only direct malware download links as malicious, which looks insufficient to me

PR #882, opened 2024-03-27
possible error with the /add route , throws internal server error

RasenRhino:
Fixes issue #883 
So if you write a custom module and you don't add your doc string, the `info` key in the `KartonBind` object is empty and hence when we try to access the `/add` route to add new targets, the method [dedent](https://github.com/CERT-Polska/Artemis/blob/a04663d992581a53e6bf4fbaa8dc16758b942fb5/artemis/templating.py#L19) and [render_markdown](https://github.com/CERT-Polska/Artemis/blob/a04663d992581a53e6bf4fbaa8dc16758b942fb5/artemis/templating.py#L23) in template.py throw an exception. This error was realised while I was writing custom modules and didn't bother to add the doc string. The line of interest in the jinja template where the input is piped to these methods is [this](https://github.com/CERT-Polska/Artemis/blob/a04663d992581a53e6bf4fbaa8dc16758b942fb5/templates/add.jinja2#L74). 
A simple check should solve the issue.


artemis/templating.py
```diff
@@ -17,10 +17,14 @@
 
 
 def dedent(text: str) -> str:
+    if(text==None):
```
kazet, line 20: `if not text`.

BTW, please run linter (according to https://github.com/cert-polska/artemis?tab=readme-ov-file#code-formatting)

kazet (changes requested):

artemis/templating.py
RasenRhino: done so.

PR #942, opened 2024-04-23
Exporting reports from the web UI

kazet:


artemis/db.py
```diff
@@ -376,6 +409,67 @@ def _get_task_deduplication_data(self, task: Task) -> str:
             }
         ).replace("\x00", " ")
 
+    def take_single_report_generation_task(self) -> Optional[ReportGenerationTask]:
+        with self.session() as session:
+            query = session.query(ReportGenerationTask).filter(
+                ReportGenerationTask.status == ReportGenerationTaskStatus.PENDING.value
+            )
+            if query.count() > 0:
```
bulkowy, line 417: you can use return `query.first()`, as it gives either first row or None when query returns no row

artemis/db.py
```diff
@@ -376,6 +409,67 @@ def _get_task_deduplication_data(self, task: Task) -> str:
...
+                return query[0]  # type: ignore
+        return None
+
+    def mark_report_generation_task_as_completed(
+        self,
+        task: ReportGenerationTask,
+        status: ReportGenerationTaskStatus,
```
bulkowy, line 424: if task status is known from function name, should we not just have `ReportGenerationTaskStatus.done` in function directly? I mean, I dont see reason to pass it here

artemis/db.py
```diff
@@ -376,6 +409,67 @@ def _get_task_deduplication_data(self, task: Task) -> str:
...
+                skip_previously_exported=skip_previously_exported,
+                status=ReportGenerationTaskStatus.PENDING,
+            )
+            session.add(task)
+            session.commit()
+
+    def get_report_generation_task(self, id: int) -> Optional[ReportGenerationTask]:
+        with self.session() as session:
+            try:
+                return session.query(ReportGenerationTask).get(id)  # type: ignore
```
bulkowy, line 454: As above, you are returning None, so for code simplicity you can just pass `.first(<filter>)` which will reduce unnecesary code block

artemis/reporting/task_handler.py
```diff
@@ -0,0 +1,67 @@
...
+            report_generation_task.custom_template_arguments,  # type: ignore
+            silent=True,
+        )
+    finally:
+        if previous_reports_directory:
+            shutil.rmtree(previous_reports_directory)
+
+
+def main() -> None:
+
```
bulkowy, line 39:
```suggestion
```

artemis/db.py
kazet: changed function name to better reflect its purpose

PR #951, opened 2024-05-05
Logging to file

kazet:


artemis/karton_logger.py
```diff
@@ -0,0 +1,26 @@
+import datetime
+from pathlib import Path
+from typing import Any, Dict
+
+from karton.core.karton import LogConsumer
+
+
+class FileLogger(LogConsumer):
```
bulkowy, line 8: This whole class seems race-condition prone with accessing I/O - eg. 2 tasks would try to open file as there was no handler before, one of them would fail as it would be too late. IK it won't happen often, but it is possibility

bulkowy (approved):

PR #998, opened 2024-05-19
Ftp creating files

kazet:


bulkowy:
I am not convinced to check for FTP Write Access. Firstly, it might depend on directory you want to upload file to. Secondly, simply having FTP open to internet should be sufficient to report it to interested party. IMO, it should be on them to verify perms and visibility of resource

bulkowy (dismissed):
After discussion, we agreed to allow checking for FTP write access

bulkowy (approved):

PR #1021, opened 2024-05-24
rough fix

kshitijk4poor:
## Changes Made
- Added a check for machine architecture before running Docker containers.
- Skips the SSL-checks container if the machine architecture is ARM.
- Provides a hint to download the Artemis-Extra-Modules if `karton-sqlmap` container is not found.

## Related Issues
Code is pretty rough, needs refactoring


scripts/start
```diff
@@ -1,23 +1,27 @@
...
 
-cd $(dirname $0)/..
-
-exists() {
-    [ -e "$1" ]
-}
-
-if exists Artemis-modules-extra; then
-    ADDITIONAL_OPTIONS=--scale=karton-sqlmap=5
+if [[ $(uname -m) != "arm"* ]]; then
```
kazet, line 3: too big duplicate - instead of copying the whole run_docker_compose command you can write it only once and add --profile $PROFILE, only changing the variable

scripts/start
```diff
@@ -1,23 +1,27 @@
 #!/bin/bash
 
-cd $(dirname $0)/..
```
kazet, line 3: this was important - why did you remove the code to set proper path in case scripts/starts was called from a ddiferent directory?

scripts/start
```diff
@@ -1,23 +1,27 @@
...
-
-exists() {
-    [ -e "$1" ]
-}
-
-if exists Artemis-modules-extra; then
```
kazet, line 9: this was important too - we don't want to scale sqlmap if the user is nto using sqlmap

scripts/start
```diff
@@ -1,23 +1,27 @@
...
-    ADDITIONAL_OPTIONS=--scale=karton-sqlmap=5
+if [[ $(uname -m) != "arm"* ]]; then
+    echo "Artemis-Extra-Modules are only x86 compatible yet"
```
kazet, line 4: the correct name is Artemis-modules-extra (even if it's not the best one possible, it is as it is now and let's use proper version and capitalization everywthere)

scripts/start
```diff
@@ -1,23 +1,27 @@
...
-# We don't use `-d` here so that the user sees the logs by default.
-. scripts/run_docker_compose up --build \
-    --scale=karton-wp_scanner=3 \
-    --scale=karton-identifier=3 \
-    --scale=karton-port_scanner=5 \
-    --scale=karton-wordpress_plugins=5 \
-    --scale=karton-nuclei=5 \
-    --scale=karton-bruter=5 \
-    $ADDITIONAL_OPTIONS
+    if [[ $(. scripts/run_docker_compose up -d --build) == *"karton-sqlmap: not found"* ]]; then
```
kazet, line 24: this won't run ;)

scripts/start
```diff
@@ -5,19 +5,26 @@ cd $(dirname $0)/..
 exists() {
     [ -e "$1" ]
 }
+if [[ $(uname -m) != "arm"* ]]; then
+    PROFILE=--profile x86
```
kazet, line 9: if architecture is not x86 it doesn't have to be arm ;)

scripts/start
```diff
@@ -5,19 +5,26 @@ cd $(dirname $0)/..
...
+else
+    echo "some modules are not ARM compatible yet so skipping them."
+    PROFILE=
+fi
 
 if exists Artemis-modules-extra; then
     ADDITIONAL_OPTIONS=--scale=karton-sqlmap=5
 else
+    echo "you do not have Artemis-modules-extra installed. please refer to https://artemis-scanner.readthedocs.io/en/latest/quick-start.html"
```
kazet, line 18: use the language of benefits - why should anyone want to install these modules? what do they bring?

tl;dr: CERT pl finds tens of thousands of misconfigured SSL and hundreds of SQL injections using these modules ;)

scripts/start
```diff
@@ -5,19 +5,26 @@ cd $(dirname $0)/..
 exists() {
     [ -e "$1" ]
 }
+if [[ $(uname -m) != "arm"* ]]; then
+    PROFILE=--profile x86
+else
+    echo "some modules are not ARM compatible yet so skipping them."
```
kazet, line 11: more descriptive - why do we commend these modules?

scripts/start
kazet: I think we may even show what command to run (just tell that the modules have different licenses) so that the burden on the user is as low as possible, they just need to copy a comamnd

PR #1129, opened 2024-07-07
Open-sourcing slow pusher

kazet:


utils/slow_pusher/slow_pusher.py
```diff
@@ -0,0 +1,91 @@
...
+            [
+                item["target"]
+                for item in requests.get(
+                    artemis_url + "api/analyses",
+                    headers={"X-API-Token": api_token},
+                ).json()
+            ]
+        ) & set(all_targets)
+
+        print(f"Num queued tasks: {num_queued_tasks}")
```
bulkowy, line 58: use logging

utils/slow_pusher/slow_pusher.py
```diff
@@ -0,0 +1,91 @@
...
+    variable in .env.
+    """
+    with open(targets_file_name, "r") as f:
+        all_targets = set([line.strip() for line in f])
+
+    while True:
+        num_queued_tasks = int(
+            requests.get(
+                artemis_url + "api/num-queued-tasks",
+                headers={"X-API-Token": api_token},
```
bulkowy, line 37: you can set `requests.Session` with this header, so you would not repeat it

bulkowy (approved):

PR #1135, opened 2024-07-08
Caching of APIs with fallback

kazet:


artemis/fallback_api_cache.py
```diff
@@ -0,0 +1,93 @@
...
+
+    CACHE = CachedSession(
+        backend=FileCache("/cache"),
+        cache_control=False,
+        expire_after=timedelta(seconds=3600),
+        allowable_codes=[200],
+        stale_if_error=True,
+    )
+
+    URL_JOOMLA_LATEST_RELEASE = CachedURL(
```
bulkowy, line 37: I am not fan of using cls fields and then iterating over them via `dir(cls)`. Change these cls fields into one cls field typed `Dict[CachedURL]` and iterate over it

artemis/fallback_api_cache.py
```diff
@@ -0,0 +1,93 @@
...
+    @classmethod
+    def get(cls, url: str, allow_unknown: bool = False) -> Response:
+        headers = {}
+        if Config.Miscellaneous.CUSTOM_USER_AGENT:
+            headers["User-Agent"] = Config.Miscellaneous.CUSTOM_USER_AGENT
+
+        response = FallbackAPICache.CACHE.get(url, headers=headers)
+
+        found = False
+        for item in cls.urls():
```
bulkowy, line 65: If we care about what URLs to be accessed with this class, check should be done before sending GET request

artemis/fallback_api_cache.py
```diff
@@ -0,0 +1,93 @@
...
+                found = True
+                if not item.validator(response.json()):
+                    raise InvalidResponseException()
+        if not allow_unknown:
+            assert found
+
+        return response
+
+    @classmethod
+    def urls(cls) -> List[CachedURL]:
```
bulkowy, line 76: having `Dict[CachedURL]` makes this method not needed

artemis/modules/joomla_scanner.py
```diff
@@ -40,7 +40,8 @@ def run(self, current_task: Task) -> None:
             joomla_version = match.group(1)
             result["joomla_version"] = joomla_version
             # Get latest release in repo from GitHub API
-            gh_api_response = requests.get("https://api.github.com/repos/joomla/joomla-cms/releases/latest")
+            gh_api_response = FallbackAPICache.URL_JOOMLA_LATEST_RELEASE.get()
```
bulkowy, line 43: inb4 - I see that you wanted cls fields solely for this reason, but still, you might just use magic methods when having dict

artemis/modules/wordpress_plugins.py
```diff
@@ -244,13 +243,9 @@ def _get_plugins_from_homepage(self, url: str) -> List[Dict[str, Any]]:
...
             if match:
                 slug = match.group(1)
-                data = json.loads(
-                    self.cached_get(
-                        f"https://api.wordpress.org/plugins/info/1.0/{slug}.json",
-                        "version-" + slug,
-                        timeout=3600,
-                    )
-                )
+                data = FallbackAPICache.get(
```
bulkowy, line 246: but then what's the point of having predefined URLs in class fields when you access other with defined url

artemis/modules/wordpress_plugins.py
```diff
@@ -379,8 +374,17 @@ def run(self, current_task: Task) -> None:
         messages = []
         closed_plugins = []
         for plugin_slug in plugins.keys():
-            existed = len(requests.get("https://api.wordpress.org/stats/plugin/1.0/" + plugin_slug).json()) > 0
-            plugin_data = requests.get(f"https://api.wordpress.org/plugins/info/1.0/{plugin_slug}.json").json()
+            existed = (
+                len(
+                    FallbackAPICache.get(
+                        "https://api.wordpress.org/stats/plugin/1.0/" + plugin_slug, allow_unknown=True
```
bulkowy, line 380: `{plugin_slug}.json`?

artemis/modules/wordpress_plugins.py
kazet: because the predefined ones will be warmed up so in case the first request after starting the container doesn't work, a stale response will be used

bulkowy (approved):

PR #1146, opened 2024-07-13
Domain Scanner

### LINKED ISSUE #1094: If a domain doesn't exist, don't scan it

kshitijk4poor:
# 🚀 Implement domain validation and task addition for efficient scanning

## 📝 Description
This PR introduces two complementary methods in the `ArtemisBase` class: `check_domain_exists` for domain validation and `add_valid_domains_task` for adding tasks for valid domains. The primary purpose of these additions is to ensure that only valid domains are sent for further scanning, improving efficiency and reducing unnecessary processing.

## 🔧 Changes
- ✨ Added `check_domain_exists` and `add_valid_domains_task` method to `ArtemisBase` class
- 🔍 Implemented domain validation logic using NS and A record lookups
- 🔄 Created a method to filter and add tasks only for valid domains
- Fix mock setup in `ArtemisModuleTestCase` to handle multiple arguments in lookup function calls, resolving TypeError in tests.
  - This change is important because:
      - It fixes a breaking error in the test suite, allowing tests to run successfully.
      - It more accurately mimics the real lookup function's signature, improving test fidelity.
      - It prevents false negatives in tests that could mask real issues in the code being tested.
      - It ensures that the mock can handle various lookup calls throughout the codebase, increasing test coverage and reliability.

## 🌟 Benefits
- 🎯 Ensures only valid domains proceed to further scanning stages
- 🚀 Improves overall system efficiency by reducing unnecessary scans
- 🧩 Provides reusable domain validation logic across modules
- 🔄 Centralizes the process of adding tasks for valid domains
- 🛡️ Reduces potential false positives from invalid or non-existent domains

## 🤝 How the methods complement each other
1. `check_domain_exists` verifies the validity of a domain by checking for NS or A records
2. `add_valid_domains_task` uses `check_domain_exists` to filter out invalid domains
3. Only domains that pass the validation are added as new tasks for further scanning

This combination ensures a streamlined and efficient domain processing pipeline.

## 📝Testing
- [x] Updated unit tests for the `check_domain_exists` method to ensure all scenarios are covered, including:
  - Domains with only NS records
  - Domains with only A records
  - Domains with both NS and A records
  - Non-existent domains
  - Scenarios where DNS resolution fails

## TODO
- remove NS records for subdomains maybe? since most of them will inherit it from the parent domains and for subdomains only A records will be checked.

## 📚 Related Issues
Closes #1094


artemis/modules/domain_scanner.py
```diff
@@ -0,0 +1,60 @@
...
+                task=current_task,
+                status=TaskStatus.OK,
+                data={
+                    "domain": domain,
+                    "exists": True
+                }
+            )
+            # Create a new task for the next module
+            new_task = Task(
+                {"type": TaskType.DOMAIN},
```
kazet, line 33: how do you plan to make other modules distinguish between filtered and unfiltered domains? maybe create a new task type for unfiltered domain and keep domain for filtered?

.gitignore
```diff
@@ -147,3 +147,4 @@ cython_debug/
 docker-compose.dev.yaml
 karton-logs
 output
+docker-compose.yaml
```
kazet, line 150: why is that needed?

docker-compose.yaml
```diff
@@ -1,6 +1,8 @@
 x-artemis-build-or-image:
   &artemis-build-or-image
-  image: certpl/artemis:latest
```
kazet, line 3: don't commit such local changes ;)

artemis/modules/port_scanner.py
```diff
@@ -214,7 +214,7 @@ def run(self, current_task: Task) -> None:
                     },
                 )
 
-                self.add_task(current_task, new_task)
+                self.add_valid_domains_task(current_task, new_task)
```
kazet, line 217: what do you think about add_task_if_domain_exists ?

artemis/modules/port_scanner.py
```diff
@@ -214,7 +214,7 @@ def run(self, current_task: Task) -> None:
                     },
                 )
 
-                self.add_task(current_task, new_task)
+                self.add_task_if_domain_exists(current_task, new_task)
```
kazet, line 217: if we have open ports, we can be sure the domain exists ;) but checking in classifier is a good idea as well as it processes incoming tasks

michalkrzem (approved):

michalkrzem (approved):

kazet:
waits for https://github.com/CERT-Polska/Artemis/issues/1261 for the tests to pass

PR #1237, opened 2024-08-27
Exports can be filtered by prefix

kazet:


artemis/db.py
```diff
@@ -472,9 +472,15 @@ def get_report_generation_task(self, id: int) -> Optional[ReportGenerationTask]:
         with self.session() as session:
             return session.query(ReportGenerationTask).filter(ReportGenerationTask.id == id).first()  # type: ignore
 
-    def list_report_generation_tasks(self) -> List[ReportGenerationTask]:
+    def list_report_generation_tasks(self, tag_prefix: Optional[str] = None) -> List[ReportGenerationTask]:
         with self.session() as session:
-            return list(session.query(ReportGenerationTask).order_by(ReportGenerationTask.created_at.desc()))
+            query = session.query(ReportGenerationTask)
+            if tag_prefix:
+                if "%" in tag_prefix:
```
bulkowy, line 479: I don't get it, why are making `%` operator as not implemented and then using it for query :thinking:

kazet: I mark % as not implemented if someone provided % in their tag prefix

bulkowy (approved):

PR #1241, opened 2024-08-29
new Nuclei

kazet:


.github/workflows/test.yml
```diff
@@ -6,7 +6,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
-    timeout-minutes: 40
+    timeout-minutes: 120
```
bulkowy, line 9: isn't 2 hours runtime for jub a little bit too much? Hour won't be sufficient?

artemis/modules/nuclei.py
```diff
@@ -247,16 +256,36 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
         )
 
+        findings_per_task = collections.defaultdict(list)
+        findings_unmatched = []
+        for finding in findings:
+            found = False
+            for task in tasks:
+                if finding["url"] in [get_target_url(task)] + links_per_task[task.uid]:
+                    findings_per_task[task.uid].append(finding)
+                    found = True
```
bulkowy, line 266: if found, then I guess safe to break from loop

artemis/modules/nuclei.py
```diff
@@ -247,16 +256,36 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
+
+        if findings_unmatched:
+            self.log.info("Findings unmatched: %s", repr(findings_unmatched))
+
+            for finding in findings_unmatched:
+                found = False
+                for task in tasks:
+                    if finding["host"].split(":")[0] == get_target_host(task).split(":")[0]:
+                        findings_per_task[task.uid].append(finding)
+                        found = True
```
bulkowy, line 278: as above

bulkowy (dismissed):

PR #1247, opened 2024-09-03
Faster scanning when locking

kazet:


artemis/config.py
```diff
@@ -124,6 +124,17 @@ class Locking:
             "see LOCK_SLEEP_MIN_SECONDS.",
         ] = get_config("LOCK_SLEEP_MAX_SECONDS", default=0.5, cast=float)
 
+        QUEUE_LOCATION_MAX_AGE_SECONDS: Annotated[
+            int,
+            """
+            Requires LOCK_SCANNED_TARGETS to be enabled.
+
+            Sometimes the task queue is very long and e.g. the first n tasks can't be taken because they concern IPs that
```
bulkowy, line 132:
```suggestion
            Sometimes the task queue is very long and e.g. the first N tasks can't be taken because they concern IPs that
```

artemis/module_base.py
```diff
@@ -204,9 +213,28 @@ def _take_and_lock_tasks(self, num_tasks: int) -> Tuple[List[Task], List[Optiona
...
             tasks = []
             locks: List[Optional[ResourceLock]] = []
-            for queue in self.backend.get_queue_names(self.identity):
-                self.log.debug(f"[taking tasks] Taking tasks from queue {queue}")
-                for i, item in enumerate(self.backend.redis.lrange(queue, 0, -1)):
+
+            if self.queue_location_timestamp < time.time() - self.queue_location_max_age_seconds:
+                self.queue_id = 0
+                self.queue_position = 0
+                self.queue_location_timestamp = time.time()
```
bulkowy, line 220: I guess, best would be if `time.time()` was assigned to variable before if statement, then it will match timestamp used in conditional

artemis/module_base.py
```diff
@@ -265,6 +293,10 @@ def _take_and_lock_tasks(self, num_tasks: int) -> Tuple[List[Task], List[Optiona
         finally:
             self.taking_tasks_from_queue_lock.release()
 
+        if len(tasks) < num_tasks:
+            self.queue_id = 0
+            self.queue_position = 0
+            self.queue_location_timestamp = time.time()
```
bulkowy, line 299: same comment as above, best if timestamp is the same in this funcblock

artemis/module_base.py
kazet: I don't see any reason to do that, I want to update the timestamp simply to current time, no problem if the two time..time() calls diverge

artemis/module_base.py
kazet: I don't see any reason to do that, I want to update the timestamp simply to current time, no problem if the two time..time() calls diverge, even better to have the current one if we spent some time in the function

bulkowy (approved):

PR #1273, opened 2024-09-16
1261 switch to linux dns resolving to allow scanning internal networks

michalkrzem:
First look


artemis/resolvers.py
```diff
@@ -16,61 +15,51 @@ class NoAnswer(Exception):
...
 
 
-def _results_from_answer(domain: str, answer: List[Dict[str, Any]], result_type: int) -> Set[str]:
+def _results_from_answer(domain: str, answer: Answer, result_type: int) -> Set[str]:
     found_results = set()
-    for entry in answer:
-        if entry["name"] != domain:
+    response = answer.response
+
+    for entry in response.answer:
```
kazet, line 22: this one returns a different answer for NS servers:

before
```
>>> lookup("wp.pl", "NS")
{'ns1.task.gda.pl', 'ns2.wp.pl', 'ns1.wp.pl'}
```
after
```
>>> lookup("wp.pl", "NS")
{'ns2.wp.pl'}

```

kazet (dismissed):

PR #1304, opened 2024-10-04
Kshitijk4poor/main

michalkrzem:


artemis/placeholder_page_analyzer.py
```diff
@@ -0,0 +1,84 @@
...
+            "trwa konserwacja",
+            "dostawcą domeny jest",
+            "domena utrzymywana",
+            "domen dostępnych w",
+            "automatycznie wygenerowane przez",
+            "w trakcie wykonywania operacji wystąpił błąd",
+            "wróć do bezpieczeństwa",
+            "utrzymywana jest na serwerach",
+            "parametry wspólne serwerów",
+            "forbidden",
```
kazet, line 21: Some of these keywords are too generic and may match too mch (e.g. "forbidden")

image

IMO let's take 5-10 most popular website providers and just match their "no website" pages (even matching with html so that we don't match some text inside)

BTW, make this configurable

artemis/config.py
```diff
@@ -698,6 +698,29 @@ class Nuclei:
                 "NUCLEI_TEMPLATE_CHUNK_SIZE is 200, three calls will be made with 200 templates each.",
             ] = get_config("NUCLEI_TEMPLATE_CHUNK_SIZE", default=200, cast=int)
 
+        class PlaceholderPagesHtmlElements:
+            PLACEHOLDER_PAGE_HTML_ELEMENTS: Annotated[
```
kazet, line 702: IMO this is a good idea to extract this to a file (have config named "PLACEHOLDER_PAGE_CONTENT_FILENAME" and these contents being loaded from file)

artemis/config.py
```diff
@@ -698,6 +698,29 @@ class Nuclei:
...
+                List[str],
+                "A list of html elements that appear on placeholder pages",
```
kazet, line 704: maybe write that exact string matching will be performed, no parsing, no nothing. And write what will happen with such pages

artemis/module_base.py
```diff
@@ -153,6 +154,11 @@ def check_domain_exists(self, domain: str) -> bool:
             bool: True if the domain exists, False otherwise.
         """
         try:
+            manager = AnalyzerManager(domain)
```
kazet, line 157: name this class more verbosely, not AnalyzerManager, maybe PlaceholderPageDetector?

artemis/module_base.py
```diff
@@ -153,6 +154,11 @@ def check_domain_exists(self, domain: str) -> bool:
...
+            result = manager.run_analysis()
```
kazet, line 158: mabe just is_placeholder(domain)

artemis/config.py
```diff
@@ -698,6 +698,29 @@ class Nuclei:
                 "NUCLEI_TEMPLATE_CHUNK_SIZE is 200, three calls will be made with 200 templates each.",
             ] = get_config("NUCLEI_TEMPLATE_CHUNK_SIZE", default=200, cast=int)
 
+        class PlaceholderPagesHtmlElements:
```
kazet, line 701: IMO add a configuration variable to enable this behavior, default false - by default people may not want random webpages to be skipped

artemis/placeholder_page_detector.py
```diff
@@ -0,0 +1,53 @@
...
+                    return False
+        response.encoding = "utf-8"
+        return response
+
+    def is_placeholder(self) -> bool:
+        response = self.check_response()
+        if response:
+            html_content = response.content
+            with open(
+                Config.Modules.PlaceholderPageContent.PLACEHOLDER_PAGE_CONTENT_FILENAME, "r", encoding="utf-8"
```
kazet, line 37: this will cause the file to be read on every check, this can be made more optimal by moving to __init__ and constrycting a singleton instance (and calling is_placeholder on different urls)

artemis/placeholder_page_detector.py
```diff
@@ -0,0 +1,53 @@
...
+            ) as file:
+                for keyword in file:
+                    if keyword.strip() in html_content:
+                        return False
+        return True
+
+
+class PlaceholderPageDetector:
```
kazet, line 45: I don't think having two classes adds anything in terms of readability

artemis/module_base.py
```diff
@@ -153,6 +154,12 @@ def check_domain_exists(self, domain: str) -> bool:
             bool: True if the domain exists, False otherwise.
         """
         try:
+            if Config.Modules.PlaceholderPageContent.PLACEHOLDER_PAGE_DETECTOR_ENABLE:
```
kazet, line 157: PLACEHOLDER_PAGE_DETECTOR_ENABLED or ENABLE_PLACEHOLDER_PAGE_DETECTOR

artemis/config.py
```diff
@@ -698,6 +698,28 @@ class Nuclei:
                 "NUCLEI_TEMPLATE_CHUNK_SIZE is 200, three calls will be made with 200 templates each.",
             ] = get_config("NUCLEI_TEMPLATE_CHUNK_SIZE", default=200, cast=int)
 
+        class PlaceholderPageContent:
+            PLACEHOLDER_PAGE_DETECTOR_ENABLE: Annotated[
+                bool,
+                "Enable or disable placeholder pages detector."
```
kazet, line 704: space or newline between dot and A

artemis/config.py
```diff
@@ -698,6 +698,28 @@ class Nuclei:
...
+                "A strict string matching will be performed without any parsing or modifications. The string will be "
```
kazet, line 705: two sentences with the same meaning: "A strict string matching will be performed without any parsing or modifications." and "The string will ...".

artemis/config.py
```diff
@@ -698,6 +698,28 @@ class Nuclei:
...
+                "matched exactly as provided, without applying any transformations or processing. If the page exists "
+                "but the specified string is found within it, the page will not be scanned for vulnerabilities. "
```
kazet, line 707: imo "and the specified"

artemis/config.py
kazet, line 704: add maybe "using this feature you may skip vulnerability scanning for websites that aren't built yet, but e.g. contain a hosting provider placeholder page"

artemis/placeholder_page_detector.py
```diff
@@ -0,0 +1,48 @@
...
+
+import requests
+
+from artemis import http_requests
+from artemis.config import Config
+
+PLACEHOLDER_PAGE_CONTENT_FILENAME = Config.Modules.PlaceholderPageContent.PLACEHOLDER_PAGE_CONTENT_FILENAME
+
+
+placeholder_page_content = []
```
kazet, line 11: this is constant as well, upcase

artemis/placeholder_page_detector.py
```diff
@@ -0,0 +1,48 @@
...
+            url = "http://" + domain
+            try:
+                response = http_requests.get(url)
+            except requests.RequestException:
+                url = "https://" + domain
+                try:
+                    response = http_requests.get(url)
+                except requests.RequestException:
+                    return False
+        response.encoding = "utf-8"
```
kazet, line 38: why is this needed?

artemis/placeholder_page_detector.py
michalkrzem: But after read the file, list will be modified.

artemis/placeholder_page_detector.py
michalkrzem: it's a remnant of beautiful soup - probably

PR #1389, opened 2024-11-13
The possibility to skip suspicious reports via API, normal form in output.json

kazet:


bulkowy (approved):

bulkowy (dismissed):

bulkowy (dismissed):

artemis/reporting/export/export_data.py
```diff
@@ -51,6 +52,15 @@ def build_export_data(
     for reporter in get_all_reporters():
         alerts.extend(reporter.get_alerts(reports))
 
+    for top_level_target in list(reports_per_top_level_target.keys()):
+        reports_per_top_level_target[top_level_target] = [
```
bulkowy, line 56: it won't be better to make temporary variable to store list result and if len != 0 then add to dict? It will make it cleaner and make del redundant

bulkowy (approved):

bulkowy (approved):

PR #1396, opened 2024-11-15
Dalfox deleted

michalkrzem:


kazet:
by just removing the module it will still be registered and the queue will grow very long. Maybe do something along the lines of https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/subdomain_enumeration.py#L63 ?

PR #1422, opened 2024-11-29
Cleanup dalfox queue

michalkrzem:
Added option to remove modules that have been removed (dalfox or others) from the queue.


michalkrzem:
#1396

PR #1467, opened 2025-01-04
Removed domain, existing vhost: bugfix, filtering domains that have similar content to parent. Piggyback: Nuclei translation

kazet:


artemis/modules/removed_domain_existing_vhost.py
```diff
@@ -57,15 +60,19 @@ def _obtain_past_target_ips(self, domain: str) -> Set[str]:
         return result
 
     @staticmethod
-    def _request_with_patched_ip(url: str, patch_ip_to: str) -> http_requests.HTTPResponse:
+    def _request_with_patched_ip(url: str, patch_domain: str, patch_ip_to: str) -> http_requests.HTTPResponse:
```
bulkowy, line 63: Is it really needed? Defaulting patch_ip_to to None and verifying it by checking whether it is empty or not should produce same result, but without need to scramble `patch_domain` outside

bulkowy (approved):

PR #1478, opened 2025-01-17
Added Moodle scanner

kshitijk4poor:
# 🔍 Add Moodle Scanner Module

Adds a new scanner module to detect and analyze Moodle installations.

## Key features:
- 🎯 Version detection & server fingerprinting
- 🔒 Vulnerability scanning via moodlescan
- 📊 Structured output parsing
- ⚡ Error handling & logging

## 📋TODO
- [ ] Add test cases


docker/Dockerfile
```diff
@@ -21,7 +21,8 @@ RUN apk add --no-cache --virtual .build-deps gcc git libc-dev make libffi-dev li
     GOBIN=/usr/local/bin/ go install -modfile go.mod github.com/projectdiscovery/nuclei/v3/cmd/nuclei && \
     cd / && \
     git clone https://github.com/rfc-st/humble.git --branch master /humble && \
-    pip install --no-cache-dir -r requirements.txt -r /humble/requirements.txt $ADDITIONAL_REQUIREMENTS && \
+    git clone https://github.com/inc0d3/moodlescan.git /moodle_scanner && \
```
kazet, line 24: this repository is licensed under GPL, and as such should be part of https://github.com/cert-polska/artemis-modules-extra package, as it's not compatible with BSD, the license of https://github.com/cert-polska/artemis

kazet:
Thank you so much for the contribution! Remember to add tests (e.g. a small Moodle instance with obsolete version of moodle that we can check against) and autoreporter module.

artemis/modules/moodle_scanner.py
```diff
@@ -0,0 +1,149 @@
...
+@dataclasses.dataclass
+class MoodleMessage:
+    category: str
+    problems: List[str]
+
+    @property
+    def message(self) -> str:
+        return f"{self.category}: {', '.join(self.problems)}"
+
+def process_moodle_json(result: Dict[str, Any]) -> List[MoodleMessage]:
```
kazet, line 22: this is a good candidate for an unit test ;)

artemis/modules/moodle_scanner.py
```diff
@@ -0,0 +1,149 @@
...
+            elif version_info and version_info != "Version not found":
+                status = TaskStatus.INTERESTING
+                status_reason = f"Found version: {version_info}"
+            else:
+                status = TaskStatus.OK
+                status_reason = "Version not found" if version_info == "Version not found" else None
+
+            self.db.save_task_result(task=current_task, status=status, status_reason=status_reason, data=result)
+
+        except subprocess.CalledProcessError as e:
```
kazet, line 135: the try...except block may wrap only the subprocess call in lines 77...84, right?

artemis/modules/moodle_scanner.py
```diff
@@ -0,0 +1,149 @@
...
+            if process.stderr:
+                self.log.warning(f"Moodlescan stderr: {process.stderr}")
+
+            # Parse the output for relevant information
+            output_lines = process.stdout.splitlines()
+            server_info = None
+            version_info = None
+            vulnerabilities = []
+
+            for i, line in enumerate(output_lines):
```
kazet, line 96: the output processing can be factored into a separate method and unittested ;)

artemis/modules/moodle_scanner.py
```diff
@@ -0,0 +1,171 @@
...
+from karton.core import Task
+
+from artemis import load_risk_class
+from artemis.binds import Service, TaskStatus, TaskType
+from artemis.module_base import ArtemisBase
+from artemis.task_utils import get_target_url
+
+
+@dataclasses.dataclass
+class MoodleMessage:
```
kazet, line 15: maybe MoodleProblems? as it contains multiple problems

kshitijk4poor: Actually, in this case it would be more appropriate to use message rather than problem. The latter narrows down the scope unnecessarily.

kazet:
hello, can we close this one or you need it for something?

PR #1486, opened 2025-01-24
Artemis supports ASNs

kazet:


artemis/modules/classifier.py
```diff
@@ -1,17 +1,40 @@
...
+
+def get_ip_prefixes_for_asn(asn: str) -> List[str]:
+    url = f"https://stat.ripe.net/data/announced-prefixes/data.json?resource={asn.upper()}"
+
+    try:
+        response = http_requests.get(url)
+        if response.status_code == 200:
+            data = response.json()
+            prefixes = [item["prefix"] for item in data["data"]["prefixes"]]
+            return prefixes
```
bulkowy, line 31: I would use setting var to limit prefixes just in case, someone uses ASN with 10k+ addresses, which might clog pipeline

bulkowy (approved):

PR #1551, opened 2025-02-25
Add periodic scanning 

### LINKED ISSUE #1286: Add periodic scanning
Currently, to have periodic scanning, you need to periodically call the API and add targets. Can it be done easier?

motypi:
resolves #1286
* Added fields in the add task form
* Used APScheduler in new scheduler.py backend to schedule periodic scans
* Modified frontend.py post /add route to include logic for checking if a periodic scan is set

### DIFF
--- a/templates/edit_scheduled_scan.jinja2
+++ b/templates/edit_scheduled_scan.jinja2
@@ -1,44 +1,121 @@
 {% extends "components/base.jinja2" %}
 {% block main %}
+
+    <script>
+        function setSelectionForAllModules(value) {
+            document.querySelectorAll(".enabled-modules input[type='checkbox']").forEach(
+                function(item) {
+                    if (item.getAttribute("name") != "module_enabled_example") {
+                        item.checked = value;
+                    }
+                }
+            );
+        }
+    </script>
+
   <h1>Edit Scheduled Scan</h1>
   <form action="/scheduled_scans/edit/{{ job_id }}" method="post">
-    <input type="hidden" name="csrf_token" value="{{ csrf_token }}" />
+ {% if validation_message %}
+            <div class="alert alert-danger" role="alert">
+                {{ validation_message }}
+            </div>
+        {% endif %}
 
-    <div class="form-group">
-      <label for="targets">Targets (one per line):</label>
-      <textarea class="form-control" name="targets" id="targets" rows="5"></textarea>
-    </div>
+        <input type="hidden" name="csrf_token" value="{{ csrf_token }}" />
 
-    <div class="form-group">
-      <label for="tag">Tag:</label>
-      <input type="text" class="form-control" name="tag" id="tag" />
-    </div>
+        <div class="form-group mb-3">
+            <label class="form-label">Targets (separated with newlines)</label>
+            <textarea class="form-control" name="targets" rows="10" required>{% for task in tasks %}{{ task }}
+{% endfor %}</textarea>
+        </div>
+        <div class="form-group mb-3">
+            <label class="form-label">Enter your tag{% if tag_names %} or select existing one{% endif %}</label>
+            <input list="object-list" class="form-control" name="tag" autocomplete="off" {% if tag %}value="{{ tag }}"{% endif %}>
+            <datalist id="object-list">
+                 {% for tag in tag_names %}
+                    <option>{{ tag.tag_name }}</option>
+                 {% endfor %}
+            </datalist>
+        <label>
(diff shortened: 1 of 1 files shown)


kazet:
Hello,

thank you for the PR and sorry in the delay in getting back to you!

artemis/frontend.py
```diff
@@ -122,6 +124,10 @@ async def post_add(
     targets: str = Form(),
     tag: Optional[str] = Form(None),
     priority: Optional[str] = Form(None),
+    schedule_enabled: Optional[bool] = Form(False),
+    schedule_interval: Optional[int] = Form(1),
```
kazet, line 128: add unit to the interval

artemis/frontend.py
```diff
@@ -174,8 +180,28 @@ async def post_add(
...
+            targets=total_list,
+            tag=tag,
+            disabled_modules=disabled_modules,
+            priority=TaskPriority(priority),
+            interval_minutes=schedule_interval,
+            start_time=start_time,
+            end_time=end_time
+        )
+
+    else:
```
kazet, line 200: what do you think about adding the possiblity to change and cancel the scheduled san?

artemis/scheduler.py
```diff
@@ -0,0 +1,60 @@
+from datetime import datetime
+import pytz
+
+from apscheduler.schedulers.background import BackgroundScheduler
```
kazet, line 4: As Artemis can get restarted, it's crucial to store the tasks in a sensible database. Would you mind adding a backend to APScheduler to ensure the jobs are persisted?

artemis/scheduler.py
```diff
@@ -0,0 +1,60 @@
...
+    :param targets: List[str] - targets to scan.
+    :param tag: Optional[str] - a tag associated with the scan.
+    :param disabled_modules: List[str] - modules to disable for the scan.
+    :param priority: TaskPriority - priority of the scan tasks.
+    :param interval_minutes: int - interval (in minutes) between each scan.
+    :param start_time: datetime - when the first run should occur (UTC recommended).
+    """
+    def scan_job():
+        # This function will be called every interval.
+        print(f"[{datetime.utcnow().isoformat()}] Running periodic scan for tag '{tag}'.")
```
kazet, line 27: use logging instead of printing

artemis/scheduler.py
```diff
@@ -0,0 +1,60 @@
...
+from apscheduler.jobstores.base import JobLookupError
+from karton.core.task import TaskPriority
+
+from artemis.producer import create_tasks
+
+scheduler = BackgroundScheduler(timezone=pytz.utc)
+scheduler.start()
+
+
+def schedule_periodic_scan(targets, tag, disabled_modules, priority, interval_minutes, start_time, end_time):
```
kazet, line 14: use Python typing ;)

templates/add.jinja2
```diff
@@ -35,12 +35,35 @@
...
                  {% endfor %}
             </datalist>
+        <label>
+            <input type="checkbox" name="schedule_enabled" id="schedule_enabled"
+                   onchange="toggleScheduleOptions(this.checked)" />
+            Schedule this scan periodically
+        </label>
+        
+        <div id="schedule-options" style="display: none;">
+            <label for="schedule_interval">Interval (in minutes):</label>
```
kazet, line 45: I think the user shouldn't see the posssibility to schedule using minutes interval as the standard scans take hours or days, what do you think about changing the field e.g. to a dropdown with possible options starting with 1 day or even a field to provide the number of days?

templates/add.jinja2
```diff
@@ -35,12 +35,35 @@
...
+            <input type="number" name="schedule_interval" id="schedule_interval" min="1" />
+        
+            <label for="schedule_start">Start Time:</label>
+            <input type="datetime-local" name="schedule_start" id="schedule_start" />
+
+            <label for="schedule_end">End Time:</label>
```
kazet, line 51: It would be nice if the task list contained the scheule data

artemis/scheduler.py
motypi: what mechanism should I use? thinking of just using python logging

evolution536:
Hello, this would be a great feature for me. Has there been any progress on this pull request? :)

Thanks in advance!

motypi:
Hi, I haven't been able to dedicate enough time to the PR. If someone else would like to take over they're welcome.

PR #1555, opened 2025-02-27
Using readme.txt to identify Joomla version

dr4g0n369:
Fixes #1634

- Made changes to `artemis/modules/joomla_scanner.py`
    - Using readme path `{url}/svr/www/htdocs/testing/README.txt`


dr4g0n369:
Hello! Thank you for giving me the issue and sorry for the delay. Please let me know if this is correct and whether any changes are needed. I will implement them as soon as possible.

kazet:
The delay is not a problem - nothing to be sorry about!

artemis/modules/joomla_scanner.py
```diff
@@ -47,6 +47,18 @@ def run(self, current_task: Task) -> None:
                 found_problems.append(f"Joomla version is too old: {joomla_version}")
                 result["joomla_version_is_too_old"] = True
 
+        # Check version via README.txt
+        response_readme = self.http_get(f"{url}/svr/www/htdocs/testing/README.txt")
```
kazet, line 51: this doesn't look as a correct path ;)

artemis/modules/joomla_scanner.py
```diff
@@ -47,6 +47,18 @@ def run(self, current_task: Task) -> None:
...
+        if response_readme.status_code == 200:
+            match_readme = re.search("([0-9]+\\.[0-9]+\\.[0-9]+)", response_readme.text) # Regex for checking version
```
kazet, line 53: I don't think such precise version is present in the README - you will have major.minor onlyy

artemis/modules/joomla_scanner.py
```diff
@@ -47,6 +47,18 @@ def run(self, current_task: Task) -> None:
...
+            if match_readme:
+                joomla_version_readme = match_readme.group(1)
+                result["joomla_version_readme"] = joomla_version_readme
+                # Check if the version is obsolete
+                if self.is_version_obsolete(joomla_version_readme):
+                    found_problems.append(f"Joomla version (from readme.txt) is too old: {joomla_version_readme}")
```
kazet, line 59: I think there could be a single version number - if xml files are present, then from xml, if they're not, then as a last resort from readme

artemis/modules/joomla_scanner.py
```diff
@@ -47,6 +47,18 @@ def run(self, current_task: Task) -> None:
                 found_problems.append(f"Joomla version is too old: {joomla_version}")
```
kazet, line 47: BTW, you may also iterate over some other xml files, as they also contain versions and might be present: not only joomla.xml, but also e.g. api/language/en-GB/install.xml or installation/language/en-GB/langmetadata.xml

dr4g0n369:
Thanks for the review, I will implement the changes as soon as possible.

dr4g0n369:
I have implemented the changes. Please review them.

Issue #433, opened 2023-07-13
Extract config into YAML


bulkowy:
Extends #371

As config.py is getting more robust, I would like to see extracting config values into YAML file, so it would be a lot easier to edit proposed values, eg. `NUCLEI_TEMPLATES_TO_SKIP`. It would also reduce LOC in config.py and make it a lot cleaner to read through, without need to access documentation.

kunalsz:
Hi @kazet , is this issue still open ? I would like to work on this.

PR #1562, opened 2025-03-01
Improved Configurations

kunalsz:
In reference to issue #433 

## Changes made
- A new file `config.yaml` is added containing all the configurations stored inside `config.py`
- `config.py` is updated to import data from the yaml file

This ensures better readability and editing of configurations ,reducing load on `config.py`

@kazet  Looking forward to your insights !

### DIFF
--- a/artemis/config.py
+++ b/artemis/config.py
@@ -1,11 +1,23 @@
 import os
+import yaml
 from typing import Annotated, Any, List, get_type_hints
 
 import decouple
 
 DEFAULTS = {}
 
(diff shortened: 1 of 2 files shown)


kazet:
Hello,

sorry in the delay in getting back to you!

The PR is going in a good direction. What I would change is that currently, the configuration is done by modifying config.yaml - this is hard to manage, a more natural way would be to have a **separate** configuration yaml file with only the settings you want to change.

PR #1563, opened 2025-03-02
Added new module lfi_detector

### LINKED ISSUE #247: Replace LFI check with an existing tool

dr4g0n369:
Resolves #247 

# Description

Implements a new module `lfi_detector`

# Progress

- Using `get_links_and_resources_on_same_domain` to get links from the given url
- Detecting lfi using the same mechanism as given in [LFI-FINDER](https://github.com/capture0x/LFI-FINDER)
    - Define some common parameters vulnerable to lfi (eg: `file`, `page`, `load`)
    - Give the payloads as inputs to these parameters in different urls obtained from first step -> `http://{url}?file={payload1}&page={payload1}&load={payload1}`
    - Check if any data from some common files are present (like `root:x` from `/etc/passwd`)

I am using [sql_injection_detector.py](https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/sql_injection_detector.py) as a reference for making this

Please let me know what changes are needed.


artemis/modules/data/lfi_detector_data.py
```diff
@@ -0,0 +1,7 @@
+# Common parameters vulnerable to LFI
+LFI_PARAMS = ["file", "page", "load"]
```
kazet, line 2: i think we might take all parameters sql_injection_detector uses, because why not?

artemis/modules/data/lfi_detector_payloads.txt
```diff
@@ -0,0 +1,861 @@
+/etc/passwd
```
kazet, line 1: Are they taken from somewhere? If yes, can you provide the source and its license as a comment in this file? (and lfi_detector_data can strip comments)

artemis/modules/lfi_detector.py
```diff
@@ -0,0 +1,119 @@
...
+        """Scan URLs for LFI vulnerabilities."""
+        message: List[Dict[str, Any]] = []
+
+        for current_url in urls:
+            for param_batch in more_itertools.batched(LFI_PARAMS, 50):  # Example parameters
+                for payload in LFI_PAYLOADS:
+                    url_with_payload = self.create_url_with_batch_payload(current_url, param_batch, payload)
+                    response = self.http_get(url_with_payload)
+
+                    if indicator := self.contains_lfi_indicator(response):
```
kazet, line 72: I would also check whether indicator is not present without injecting anything, to decrease the amount of false positives

artemis/modules/lfi_detector.py
```diff
@@ -0,0 +1,119 @@
...
+
+            links = [
+                link.split("#")[0]
+                for link in links
+                if not any(link.split("?")[0].lower().endswith(extension) for extension in STATIC_EXTENSIONS)
+            ]
+
+            random.shuffle(links)
+
+            message = self.scan(urls=links[:25], task=current_task)
```
kazet, line 104: the 25 can be configurable - BTW, it could be a single variable for all modules that crawl: this one, Nuclei, sq_injection_detector

artemis/modules/lfi_detector.py
```diff
@@ -0,0 +1,119 @@
...
+
+            if message:
+                status = TaskStatus.INTERESTING
+                status_reason = ", ".join([m["statement"] for m in message])
```
kazet, line 108: IMO `message` should be plural, `messages` ;)

artemis/reporting/modules/lfi_detector/reporter.py
```diff
@@ -0,0 +1,59 @@
...
+        if not isinstance(task_result["result"], dict):
+            return []
+
+        url_parsed = urllib.parse.urlparse(task_result["result"]["result"][0]["url"])
+
+        return [
+            Report(
+                top_level_target=get_top_level_target(task_result),
+                target=f"{url_parsed.scheme}://{url_parsed.netloc}",
+                report_type=LFIDetectorReporter.LFI_CORE,
```
kazet, line 34: _CORE is not needed here - for sql_injection_detector we've added that to distinguish from sql_injection from sqlmap, but here we have only one LFI module

artemis/reporting/modules/lfi_detector/reporter.py
```diff
@@ -0,0 +1,59 @@
...
+
+        if not isinstance(task_result["result"], dict):
+            return []
+
+        url_parsed = urllib.parse.urlparse(task_result["result"]["result"][0]["url"])
+
+        return [
+            Report(
+                top_level_target=get_top_level_target(task_result),
+                target=f"{url_parsed.scheme}://{url_parsed.netloc}",
```
kazet, line 33: could we provide the full URL here?

kazet:
Thanks for the PR - I have added some minor comments but overall the pull request is very good!

dr4g0n369:
Thanks for the review, I will start working on the changes

artemis/modules/data/lfi_detector_payloads.txt
```diff
@@ -0,0 +1,863 @@
+# Copied from GNU General Public License v3.0 (https://github.com/capture0x/LFI-FINDER/blob/main/LICENSE)
```
kazet, line 1: hmm, if it's GPL, we can't use it in Artemis, which is licensed under BSD

we have two options for now:
- finding a payload list which is licensed under BSD or MIT
- moving the module to https://github.com/CERT-Polska/Artemis-modules-extra, a repository with modules separate from Artemis with varying licenses

I would suggest going with the first one, as it will allow us to have a LFI module in core Artemis. What do you think?

dr4g0n369: Yes, definitely the first option would be better, I think we can use a wordlist from https://github.com/danielmiessler/SecLists

I will change it right now

dr4g0n369:
I have changed the wordlist. Please see if it's all good now.

artemis/modules/data/lfi_detector_data.py
kazet: did I miss something or this one isn't actually resolved?

artemis/modules/data/lfi_detector_payloads.txt
```diff
@@ -0,0 +1,931 @@
...
+/etc/ftphosts
+/etc/group
+/etc/hostname.bge
+/etc/hostname.ce0
+/etc/hostname.ce1
+/etc/hostname.ce2
+/etc/hostname.ce3
+/etc/hostname.dcelx0
+/etc/hostname.dcelx1
+/etc/hostname.dcelx2
```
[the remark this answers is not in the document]
dr4g0n369: True, that many payloads would put an unnecessary load on the target. I will modify the wordlist to remove the redundant payloads.

dr4g0n369:
I have decreased the wordlist size to `179` payloads. Since our aim is just to detect LFI (and not look for sensitive files), I think its sufficient to just verify it using `/etc/passwd` file. Thus, I have removed the indicator for `php` and only left indicators for `/etc/passwd` and `Windows .ini file`. 

This should hopefully not put a lot of burden on a target that we scan.

artemis/modules/data/lfi_detector_data.py
dr4g0n369: No, this was resolved, I am using the `URL_PARAMS` from sql_injection_detector. I will remove this line in order to avoid confusion.

artemis/modules/lfi_detector.py
```diff
@@ -0,0 +1,119 @@
...
+
+from artemis import load_risk_class
+from artemis.binds import Service, TaskStatus, TaskType
+from artemis.config import Config
+from artemis.crawling import get_links_and_resources_on_same_domain
+from artemis.http_requests import HTTPResponse
+from artemis.module_base import ArtemisBase
+from artemis.modules.data.lfi_detector_data import LFI_PAYLOADS
+from artemis.modules.data.static_extensions import STATIC_EXTENSIONS
+from artemis.sql_injection_data import URL_PARAMS
```
kazet, line 18: if we use this in two modules, it can be extracted to e.g. artemis/modules/data/

kazet:
Nice PR! Two minor comments and I think we can get it merged

kazet:
Thanks again for your contribution <3

Issue #1036, opened 2024-05-29
Detect wildcards during reporting


kazet:
so that if the same problem exists on a.example.com and b.example.com and anydomain.example.com, only one instance is reported

Nazihbenbrahim:
Hi @kazet, I'd love to work on this issue! Is it available for me to take?

Nazihbenbrahim:
Hi @kazet,

I hope you're doing well. I'm working on the Artemis project and facing multiple import-related issues. I've tried several approaches to resolve them, but I'm still stuck. Could you please help me figure out what's going wrong?
Problem Description

I’m encountering the following errors when trying to run my script (test_report.py):

    ModuleNotFoundError: No module named 'artemis.reporting.base.normal_form'

    ImportError: cannot import name 'ReportType' from 'artemis.reporting.base.report_type'

    AttributeError: module 'artemis.reporting.base.reporters' has no attribute 'get_all_reporters'

These errors occur when I try to import modules in report.py.

PR #1580, opened 2025-03-05
Add wildcard detection and deduplication logic to report.py

Nazihbenbrahim:
# Pull Request: Add Wildcard Detection and Deduplication Logic to `report.py`

## Description

This PR introduces **wildcard detection** and **deduplication logic** to the `report.py` file in the Artemis project. The goal is to ensure that vulnerabilities found on multiple subdomains of the same root domain (e.g., `a.example.com` and `b.example.com`) are reported only once under the wildcard domain (`*.example.com`). This improves the efficiency and readability of vulnerability reports by reducing redundancy.

This PR addresses **Issue #1036: Detect wildcards during reporting**.

---

### 1. Wildcard Detection
- Added the `get_root_domain` method to the `Report` class. This method extracts the root domain (wildcard) from a given target (domain or URL). For example:
  - `a.example.com` → `*.example.com`
  - `sub.a.example.com` → `*.example.com`
  - `example.com` → `example.com` (no wildcard applied).

### 2. Deduplication Logic
- Added the `group_reports_by_root_domain` function to group reports by their root domain (wildcard).
- Added the `deduplicate_reports` function to consolidate reports by selecting the best report (based on score) for each root domain group.

### 3. Integration with Existing Code
- Modified the `get_normal_form` method to include the root domain in the normal form of a report. This ensures that reports with the same root domain are treated as duplicates during deduplication.
- Updated the `get_score` method to prioritize reports based on their relevance (e.g., HTTPS over HTTP, non-www over www).

---

## Why This Change Is Important

- **Reduces Redundancy**: Vulnerabilities on multiple subdomains of the same root domain are now reported only once, reducing clutter in the output.
- **Improves Readability**: Consolidated reports make it easier for users to focus on the most relevant information.
- **Enhances Efficiency**: The deduplication logic reduces the number of reports generated, making the system more efficient.

---

### DIFF
--- a/artemis/reporting/base/report.py
+++ b/artemis/reporting/base/report.py
@@ -4,6 +4,7 @@ import urllib.parse
 import uuid
 from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional
+from collections import defaultdict
 
 from artemis.domains import is_domain
 from artemis.reporting.severity import Severity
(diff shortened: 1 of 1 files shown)


kazet:
Sorry for the delay in getting back to you. I think a better approach would be to detect wildcards during subdomain enumeration. For example, if we enumerate subdomains of example.com and we found sub1.example.com, if we detect that it has identical content to sub-[random-suffix].example.com, we can skip scanning of that subdomain. what do you think?

kazet:
Closing for now, let me know if you still want to work on that one!

PR #1601, opened 2025-03-13
Adding nuclei workflows and 403-bypass workflow

dr4g0n369:
# Changes

- Added new command in `nuclei.py` to use nuclei workflows
- Added a new workflow for `403 bypass` in `nuclei_workflows_custom`

Please let me know what you think of this and whether we should go with this idea or not.


artemis/modules/data/nuclei_workflows_custom/templates/403_nullbyte_check.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - |
+        GET {{Path}}%00 HTTP/1.1
+        Host: {{Hostname}}
+
+      - |
+        POST {{Path}}%00 HTTP/1.1
+        Host: {{Hostname}}
+
+      - |
+        PUT {{Path}}%00 HTTP/1.1
```
kazet, line 21: what about also a nonexistent HTTP method and version without nullbytes?

artemis/modules/data/nuclei_workflows_custom/templates/testing_403.yaml
```diff
@@ -0,0 +1,30 @@
+id: 403-bypass-using-headers
```
kazet, line 1: IMO the file name should correspond to the template name

artemis/modules/nuclei.py
```diff
@@ -295,6 +297,44 @@ def _scan(self, templates: List[str], targets: List[str]) -> List[Dict[str, Any]
                 stdout_utf8_lines = stdout_utf8.split("\n")
                 stderr_utf8_lines = stderr_utf8.split("\n")
 
+                # command for using nuclei workflows
+                command = [
```
kazet, line 301: some common code can be extracted :)

artemis/modules/nuclei.py
```diff
@@ -295,6 +297,44 @@ def _scan(self, templates: List[str], targets: List[str]) -> List[Dict[str, Any]
                 stdout_utf8_lines = stdout_utf8.split("\n")
                 stderr_utf8_lines = stderr_utf8.split("\n")
```
kazet, line 298: here you set stdout_utf8_lines and later override it

dr4g0n369:
I have made the changes. 
Mainly moving the `workflow` portion in a separate loop outside of the `template_chunk` as that would make the workflows run again and again). I have defined a base `command` and then I am using that to generate `template_command` and `workflow_command`. 

Please let me know if there are any issues in this.

bulkowy (approved):

PR #1616, opened 2025-03-19
Module config

### LINKED ISSUE #1585: Allow configuring modules before they're run - e.g. when adding targets, checking which Nuclei templates should be run - e.g. all? only medium-severity and up?

kshitijk4poor:
aims to fix #1585


artemis/api.py
```diff
@@ -84,12 +91,20 @@ def add(
     elif not disabled_modules:
         disabled_modules = Config.Miscellaneous.MODULES_DISABLED_BY_DEFAULT
 
+    # Get module configurations from Redis
```
kazet, line 94: I don't think this is a good idea to store the configs in Redis, because that way there is one, global configuration, and to add some tasks with a custom one you need to change the global settings and then add tasks. What do you think about configs being just a parameter to add()?

artemis/api.py
```diff
@@ -345,3 +360,43 @@ def _get_search_query(request: Request) -> Optional[str]:
         raise NotImplementedError("Regex search is not yet implemented")
 
     return request.query_params["search[value]"]
+
+
+@router.post("/v1/configure_module", dependencies=[Depends(verify_api_token)])
+async def configure_module(request: ModuleConfigurationRequest) -> Dict[str, Any]:
```
kazet, line 366: As in the above comment, this could be a part of add()

artemis/config.py
```diff
@@ -21,6 +23,43 @@ def get_config(name: str, **kwargs) -> Any:  # type: ignore
     )
 
 
+class SeverityThreshold(enum.Enum):
```
kazet, line 26: what do you think about a separate folder, module_configurations, so that config.py doesn't grow too much?
that way this one could reside e.g. in artemis/modules_config/nuclei.py

artemis/modules/base/module_configuration.py
```diff
@@ -0,0 +1,56 @@
...
+        self.enabled = enabled
+
+    def serialize(self) -> Dict[str, Any]:
+        """
+        Serialize the configuration to a dictionary.
+
+        Returns:
+            Dict[str, Any]: Dictionary representation of the configuration.
+        """
+        return {"enabled": self.enabled}
```
kazet, line 33: the decision whether modules are enables is made earlier, so "enabled" shouldn't be there, IMO you can return an empty dictionary

docs/user-guide/configuration.rst
```diff
@@ -12,6 +12,186 @@ Additionally, you can configure modules from the ``Artemis-modules-extra`` repos
 the configuration variables from https://github.com/CERT-Polska/Artemis-modules-extra/blob/main/extra_modules_config.py. The file to put them
 in (``.env``) and the syntax (``VARIABLE_NAME=VARIABLE_VALUE``) is the same as for the core Artemis configuration.
 
+Module Configuration
+-------------------
+
+The ``ModuleConfiguration`` class serves as the base for all module-specific configurations in Artemis. It provides a standardized way to handle module configurations with serialization, deserialization, and validation capabilities.
```
kazet, line 18: I think we need to underscore, maybe in multiple places, that it's runtime configuration, for stuff you want to change during scanning (how aggressive the scanning is)

artemis/modules/bruter.py
```diff
@@ -54,6 +56,32 @@ class BruterResult:
...
+class BruterConfiguration(ModuleConfiguration):
+    """Configuration for the bruter module."""
+
+    def __init__(self, enabled: bool = True, max_attempts: int = 1000) -> None:
+        super().__init__(enabled=enabled)
+        self.max_attempts = max_attempts
+
+    def serialize(self) -> Dict[str, Any]:
+        result = super().serialize()
+        result["max_attempts"] = self.max_attempts
```
kazet, line 68: what is max_attempts? I don't see that implemented

If you want to provide something other than nuclei as well, whast do you think about including warnings in mail_DNS_scanner reporter?

artemis/modules/nuclei.py
```diff
@@ -203,7 +236,11 @@ def _scan(
         if not targets:
             return []
 
-        templates_or_workflows_filtered = []
+        # Apply max_templates limit from configuration if set
+        if self.configuration and self.configuration.max_templates:
+            templates = templates[: self.configuration.max_templates]
```
kazet, line 241: I don't think this is a good idea as it will skip the tempaltes quite randomly - I would vote for keeping severity as the only means of controlling how aggressive the module is

artemis/api.py
kshitijk4poor: I agree that storing configs in Redis is problematic. It creates a global configuration state, forcing you to change settings for all tasks at once. Making configs a parameter to add() would allow per-task configurations, enabling parallel scanning with different settings without modifying global state, and creating a more flexible, explicit API design.

artemis/api.py
```diff
@@ -84,12 +96,32 @@ def add(
...
+        for module_name, config in module_configs.items():
+            config_class = ConfigurationRegistry().get_configuration_class(module_name)
+            if config_class:
+                try:
+                    config_instance = config_class.deserialize(config)
+                    if not config_instance.validate():
+                        raise ValueError(f"Invalid configuration for module {module_name}")
+                except ValueError as e:
+                    raise HTTPException(status_code=400, detail=f"Invalid configuration for {module_name}: {str(e)}")
+    # If no module_configs provided, get configs from Redis as fallback
```
kazet, line 110: I don't think this fallback is particularily needed, as it adds a dependency on the global state

test/unit/test_nuclei_configuration.py
```diff
@@ -0,0 +1,109 @@
...
+from artemis.modules.nuclei_configuration import NucleiConfiguration
+
+
+class TestNucleiConfiguration(unittest.TestCase):
+    """Test suite for the NucleiConfiguration class."""
+
+    def test_init_with_default_values(self) -> None:
+        """Test initialization with default values."""
+        config = NucleiConfiguration()
+        self.assertTrue(config.enabled)
```
kazet, line 15: here you still have `enabled` configuration

docs/features.rst
```diff
@@ -56,3 +56,38 @@ REST API
 
 REST API documentation is auto-generated by the FastAPI framework in the form of
 Swagger and is available at your Artemis instance under ``/docs`` URL.
+
+Module Configuration API
```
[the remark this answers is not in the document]
kazet: this one is not yet solved ;)

artemis/reporting/modules/mail_dns_scanner/template_misconfigured_email.jinja2
```diff
@@ -28,4 +28,36 @@
...
+                            {% endif %}
+
+                            {{ report.additional_data.message_translated }}
+                        {{ report_meta(report) }}
+                    </li>
+                {% endif %}
+            {% endfor %}
+        </ul>
+        <p>
+            {{ _("While these issues are not critical, fixing them can improve the reliability of your email delivery and security.")|safe }}
```
kazet, line 59: I don't think |safe is needed here

artemis/modules/bruter.py
kazet: I don't think this one is resolved ;)

artemis/config.py
```diff
@@ -329,6 +330,20 @@ class Nuclei:
                 cast=decouple.Csv(str, delimiter=","),
             )
 
+            NUCLEI_SEVERITY_THRESHOLD: Annotated[
```
kazet, line 333: NUCLEI_TEMPLATE_LISTS can also be updated :)

artemis/module_base.py
```diff
@@ -559,6 +606,17 @@ def process_multiple(self, tasks: List[Task]) -> None:
         if len(tasks) == 0:
             return
 
+        # Load configuration from first task's payload if present
```
kazet, line 609: IMO if there are multiple tasks with separate configurations, we should process them separately, not just take the config for the first one and call it a day ;)

artemis/config.py
kshitijk4poor: should NUCLEI_TEMPLATE_LISTS be list instead of str? as it actually returns a List[str] via decouple.Csv(str, delimiter=",") cast

PR #1632, opened 2025-03-24
Limitting dns_scanner scope when ip range scanning

kazet:


artemis/modules/dns_scanner.py
```diff
@@ -37,6 +38,11 @@ def run(self, current_task: Task) -> None:
         nameservers = list(set(soa_result + ns_result))
         result["nameservers"] = nameservers
 
+        # If the task originated from scanning an IP range, that means, that we only want to
+        # check the nameservers that belong to that IP range, not random ones.
+        if has_ip_range(current_task):
+            nameservers = [nameserver for nameserver in nameservers if nameserver in get_ip_range(current_task)]
```
bulkowy, line 44: it should not be appended to existing `nameservers` arg? and `result["nameservers"]` should not be updated as well?

bulkowy (approved):

artemis/modules/dns_scanner.py
kazet: the first sentence - right, the second - no, we do want to filter and skip

bulkowy (approved):

PR #1753, opened 2025-05-30
Offat based api hacking module

dr4g0n369:
This is an API Hacking module that uses `openapi` specification files in order to test APIs. Largely based on [OWASP/OFFAT](https://github.com/OWASP/OFFAT/).

### DIFF
--- a/artemis/reporting/modules/api_scanner/reporter.py
+++ b/artemis/reporting/modules/api_scanner/reporter.py
@@ -22,11 +22,13 @@ class APIScannerReporter(Reporter):
         if task_result["status"] != "INTERESTING":
             return []
 
-        if not isinstance(task_result.get("data"), dict) or not isinstance(task_result["data"].get("results"), list):
+        if not isinstance(task_result.get("result"), dict) or not isinstance(
+            task_result["result"].get("results"), list
+        ):
             return []
 
         reports = []
-        for result in task_result["data"]["results"]:
+        for result in task_result["result"]["results"]:
             url = result["url"]
 
             reports.append(
--- a/artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
+++ b/artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
@@ -1,42 +1,31 @@
 {% if "api_vulnerability" in data.contains_type %}
-    <li>{% trans %}We identified potential API security vulnerabilities:{% endtrans %}
+    <li>{% trans %}We identified following API vulnerabilities:{% endtrans %}
         <ul>
             {% for report in data.reports %}
                 {% if report.report_type == "api_vulnerability" %}
                     <li>
-                        {{ report.target }}
-                        <ul>
-                            <li>{% trans %}Method:{% endtrans %} {{ report.additional_data.method|upper }}</li>
-                            <li>{% trans %}Endpoint:{% endtrans %} {{ report.additional_data.endpoint }}</li>
-                            <li>{% trans %}Status Code:{% endtrans %} {{ report.additional_data.status_code }}</li>
-                            <li>
-                                {% trans %}Details:{% endtrans %}
-                                {% set raw_detail = report.additional_data.details %}
-                                {% if "HTTP verb" in raw_detail %}
-                                    {% trans %}Unsupported HTTP Method - The endpoint accepts HTTP methods that should not be allowed, potentially enabling unintended operations.{% endtrans %}
-                                {% elif "vulnerable to SQL Injection" in raw_detail or "vulnerable to SQli" in raw_detail %}
-                                    {% trans %}SQL Injection - Malicious input can manipulate database queries, leading to data leaks or unauthorized access.{% endtrans %}
-                                {% elif "time-based SQL Injection" in raw_detail %}
-                                    {% trans %}Time-based SQL Injection - An attacker can infer data using timing-based payloads, exploiting database behavior.{% endtrans %}
-                                {% elif "OS Command Injection" in raw_detail %}
-                                    {% trans %}OS Command Injection - Arbitrary system commands can be executed on the server, leading to full compromise.{% endtrans %}
-                                {% elif "XSS/HTML Injection" in raw_detail %}
-                                    {% trans %}XSS/HTML Injection - Untrusted input is rendered in the browser, enabling attacks such as session hijacking or defacement. In the case of defacement, reflected XSS allows an attacker to inject and execute arbitrary HTML or JavaScript, altering the appearance or content of the affected page for the victim, effectively impersonating changing site behavior.{% endtrans %}
-                                {% elif "SSTI" in raw_detail %}
-                                    {% trans %}SSTI - Server-Side Template Injection allows attackers to execute code on the server by injecting into templates.{% endtrans %}
-                                {% elif "fails to implement security authentication" in raw_detail %}
-                                    {% trans %}Missing Authentication - Endpoints lack proper access control, allowing unauthorized access.{% endtrans %}
-                                {% else %}
-                                    {{ raw_detail }}
-                                {% endif %}
-                            </li>
-                            {% if report.additional_data.data_leak %}
-                                <li>{% trans %}Data leaks:{% endtrans %} {{ report.additional_data.data_leak }}</li>
-                            {% endif %}
-                            {% if report.additional_data.curl_command %}
-                                <li>{% trans %}CURL Command:{% endtrans %} <code>{{ report.additional_data.curl_command }}</code></li>
-                            {% endif %}
-                        </ul>
+                        <tt>{{ report.target }}</tt> -
+                        {% set raw_detail = report.additional_data.details %}
+                        {% if "HTTP verb" in raw_detail %}
+                            {% trans %}the endpoint accepts HTTP methods that is not documented as allowed, potentially enabling unintended operations{% endtrans %}
+                        {% elif "vulnerable to SQL Injection" in raw_detail or "vulnerable to SQli" in raw_detail %}
+                            {% trans %}SQL Injection - malicious input can manipulate database queries, leading to data leaks or unauthorized access{% endtrans %}
+                        {% elif "time-based SQL Injection" in raw_detail %}
+                            {% trans %}time-based SQL Injection - an attacker can infer data using timing-based payloads, exploiting database behavior{% endtrans %}
+                        {% elif "OS Command Injection" in raw_detail %}
+                            {% trans %}OS Command Injection - arbitrary system commands can be executed on the server, leading to full compromise{% endtrans %}
+                        {% elif "XSS/HTML Injection" in raw_detail %}
+                            {% trans %}XSS/HTML Injection - untrusted input is rendered in the browser, enabling attacks such as session hijacking or defacement. In the case of defacement, reflected XSS allows an attacker to inject and execute arbitrary HTML or JavaScript, altering the appearance or content of the affected page for the victim, effectively impersonating the administrator or changing site behavior{% endtrans %}
+                        {% elif "SSTI" in raw_detail %}
+                            {% trans %}SSTI - Server-Side Template Injection allows attackers to execute code on the server by injecting into templates{% endtrans %}
+                        {% elif "fails to implement security authentication" in raw_detail %}
+                            {% trans %}missing authentication - endpoints lack proper access control, allowing unauthorized access{% endtrans %}
+                        {% else %}
+                            {{ raw_detail }}
+                        {% endif %}
+                        ({% trans %}method:{% endtrans %}
+                        {{ report.additional_data.method|upper }},
+                        {% trans %}status code:{% endtrans %} {{ report.additional_data.status_code }}{% if report.additional_data.data_leak %}, {% trans %}data leaks:{% endtrans %} {{ report.additional_data.data_leak }}{% endif %}).
                     </li>
                 {% endif %}
             {% endfor %}
--- a/artemis/reporting/modules/api_scanner/translations/pl_PL/LC_MESSAGES/messages.po
+++ b/artemis/reporting/modules/api_scanner/translations/pl_PL/LC_MESSAGES/messages.po
@@ -1,30 +1,38 @@
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:2
 msgid "We identified following API vulnerabilities:"
-msgstr ""
+msgstr "Wykryliśmy następujące podatności w końcówkach API:"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:10
 msgid ""
 "the endpoint accepts HTTP methods that is not documented as allowed, "
 "potentially enabling unintended operations"
 msgstr ""
+"końcówka akceptuje metody HTTP które nie są udokumentowane, potencjalnie "
+"umożliwiając nieuprawnione operacje"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:12
 msgid ""
 "SQL Injection - malicious input can manipulate database queries, leading "
 "to data leaks or unauthorized access"
 msgstr ""
+"SQL Injection - atakujący może manipulować zapytaniami do bazy danych, co"
+" umożliwia pobranie danych lub nieuprawniony dostęp do systemu"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:14
 msgid ""
 "time-based SQL Injection - an attacker can infer data using timing-based "
 "payloads, exploiting database behavior"
 msgstr ""
+"SQL Injection - atakujący może manipulować zapytaniami do bazy danych, co"
+" umożliwia pobranie danych lub nieuprawniony dostęp do systemu"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:16
 msgid ""
 "OS Command Injection - arbitrary system commands can be executed on the "
 "server, leading to full compromise"
 msgstr ""
+"wstrzykiwanie poleceń systemowych - atakujący może wykonać dowolne "
+"polecenia na serwerze, uzyskując pełny dostęp do systemu"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:18
 msgid ""
@@ -35,27 +43,36 @@ msgid ""
 "affected page for the victim, effectively impersonating the administrator"
 " or changing site behavior"
 msgstr ""
+"Cross-Site Scripting - niezaufana treść jest wyświetlana/wykonywana w "
+"przeglądarce internetowej, co umożliwia atakującemu spreparowanie linku, "
+"który, po kliknięciu przez administratora, wykona dowolną akcję z jego "
+"uprawnieniami (taką jak np. modyfikacja treści czy kradzież danych)"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:20
 msgid ""
 "SSTI - Server-Side Template Injection allows attackers to execute code on"
 " the server by injecting into templates"
 msgstr ""
+"Server-Side Template Injection - atakujący może wstrzykiwać własną "
+"zawartość szablonów HTML a dzięki temu wykonywać kod na serwerze"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:22
 msgid ""
 "missing authentication - endpoints lack proper access control, allowing "
 "unauthorized access"
 msgstr ""
+"brak uwierzytelnienia - końcówka nie weryfikuje, czy użytkownik jest "
+"uprawniony do jej uruchomienia, co może prowadzić do nieuprawnionego "
+"dostępu do systemu"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:26
 msgid "method:"
-msgstr ""
+msgstr "metoda:"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:28
 msgid "status code:"
-msgstr ""
+msgstr "kod odpowiedzi:"
 
 #: artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2:28
 msgid "data leaks:"
-msgstr ""
+msgstr "wyciek danych:"
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -94,6 +94,20 @@ services:
       - ./output/autoreporter/:/opt/output/autoreporter
       - ${DOCKER_COMPOSE_ADDITIONAL_SHARED_DIRECTORY:-./shared}:/shared/
 
+  karton-api_scanner:
+    <<: [*logging, *artemis-build-or-image]
+    command: "python3 -m artemis.modules.api_scanner"
+    env_file: .env
+    restart: always
+    volumes: ["./docker/karton.ini:/etc/karton/karton.ini", "${DOCKER_COMPOSE_ADDITIONAL_SHARED_DIRECTORY:-./shared}:/shared/"]
+
+  karton-admin_panel_login_bruter:
+    <<: [*logging, *artemis-build-or-image]
+    command: "python3 -m artemis.modules.admin_panel_login_bruter"
+    env_file: .env
+    restart: always
+    volumes: ["./docker/karton.ini:/etc/karton/karton.ini", "${DOCKER_COMPOSE_ADDITIONAL_SHARED_DIRECTORY:-./shared}:/shared/"]
+
   karton-bruter:
     <<: [*logging, *artemis-build-or-image]
     command: "python3 -m artemis.modules.bruter"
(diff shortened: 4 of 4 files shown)


test/modules/test_api_scanner.py
```diff
@@ -0,0 +1,35 @@
...
+        self.run_task(task)
+        (call,) = self.mock_db.save_task_result.call_args_list
+        self.assertEqual(call.kwargs["status"], TaskStatus.INTERESTING)
+        self.assertEqual(call.kwargs["data"]["results"][0]["url"], "http://test-flask-vulnerable-api:5000/api/login")
+        self.assertEqual(call.kwargs["data"]["results"][0]["method"], "POST")
+        self.assertEqual(call.kwargs["data"]["results"][0]["vulnerable"], True)
+        self.assertEqual(call.kwargs["data"]["results"][0]["vuln_details"], "One or more parameter is vulnerable to SQL Injection Attack")
+        self.assertEqual(call.kwargs["data"]["results"][0]["status_code"], 500)
+
+        self.assertEqual(call.kwargs["data"]["results"][0]["url"], "http://test-flask-vulnerable-api:5000/api/user/' AND (SELECT * FROM (SELECT(SLEEP(5)))abc)")
```
kazet: for now let's only use GETs, as we don't want to be intrusive. If you want to, you may make it configurable

artemis/config.py
```diff
@@ -302,6 +302,13 @@ class AdminPanelLoginBruter:
                 "How many times to recheck whether the good password works, and the bad doesn't",
             ] = get_config("ADMIN_PANEL_LOGIN_BRUTER_NUM_RECHECKS", default=10, cast=int)
 
+        class APIScanner:
+            ONLY_GET_REQUESTS: Annotated[
+                bool,
+                "If set to True, api scanner will only use GET requests to scan. If to False, a more intrusive scan "
```
kazet: api -> API

artemis/config.py
```diff
@@ -302,6 +302,13 @@ class AdminPanelLoginBruter:
...
+                "can be carried (using other http methods like POST)",
```
kazet: carried -> carried out, http -> HTTP ;)

artemis/modules/api_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+from artemis.config import Config
+from artemis.module_base import ArtemisBase
+from artemis.task_utils import get_target_url
+
+COMMON_SPEC_PATHS = [
+    "/swagger.json",
+    "/v2/swagger.json",
+    "/v3/api-docs",
+    "/openapi.json",
+    "/api-docs",
```
kazet: add /docs, BTW is there by any chance a longer list? IMO we can use e.g. 50-100 requests to find swagger ;)

artemis/modules/api_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+                        self.log.info(f"Unable to validate spec at {try_url}: {e}")
+                        continue
+            except Exception as e:
+                self.log.debug(f"Error checking {try_url}: {e}")
+                continue
+        return None
+
+    def install_offat(self) -> None:
+        if not os.path.exists("/offat"):
+            os.system("git clone https://github.com/OWASP/OFFAT /offat")
```
kazet: I'd rather use subprocess.call, it's less prone to command injection if we start building these strings somehow

artemis/modules/api_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+            patch_file_path = os.path.join(os.path.dirname(__file__), "data/offat", "offat_artemis.patch")
+            os.system(f"git -C /offat apply {patch_file_path}")
+
+        os.system("pip install -e /offat/src")
```
kazet: ditto

artemis/modules/api_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def scan(self, target_api_specification: str) -> Dict[str, Any]:
+        output_file = "/tmp/output.json"
+
+        if Config.Modules.APIScanner.ONLY_GET_REQUESTS:
+            os.system(
```
kazet: ditto

artemis/modules/api_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+                        "--only-get-requests",
+                        "-o",
+                        output_file,
+                        "--format",
+                        "json",
+                    ]
+                )
+            )
+        else:
+            os.system(" ".join(["offat", "-f", target_api_specification, "-o", output_file, "--format", "json"]))
```
kazet: you can have one subprocess.call invocation and add --only-get-requests to the command, that way we won't make the mistake of updating only one call() ;)

artemis/modules/api_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+                            status_code=result.get("response_status_code"),
+                        )
+                    )
+
+            if results:
+                status = TaskStatus.INTERESTING
+                status_reason = "Found potential vulnerabilities in the api"
+            else:
+                status = TaskStatus.OK
+                status_reason = None
```
kazet: even if we don't find anything, I'd save the detected swagger path ("status=ok, reason="detected API on ..., no vulnerasbilities found")

artemis/modules/data/offat/offat_artemis.patch
```diff
@@ -0,0 +1,86 @@
...
+         required=True,
+     )
++    # parser.add_argument(
++    #     '-v', '--version', action='version', version=f"%(prog)s {get_package_version()}"
++    # )
+     parser.add_argument(
+-        '-v', '--version', action='version', version=f"%(prog)s {get_package_version()}"
++        '--only-get-requests',
++        dest='only_get_requests',
++        help="send only get requests while scanning",
```
kazet: get -> GET

artemis/reporting/modules/api_scanner/reporter.py
```diff
@@ -0,0 +1,80 @@
...
+
+        reports = []
+        for result in task_result["data"]["results"]:
+            try:
+                url = result["url"]
+                parsed_url = urllib.parse.urlparse(url)
+
+                # Resolve hostname to IP
+                ips = list(lookup(parsed_url.hostname)) if parsed_url.hostname else []
+                resolved_host = ips[0] if ips else parsed_url.hostname
```
kazet: I don't think we need resolved hos in the report, is there a particular reason you added it?

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
```diff
@@ -0,0 +1,30 @@
+{% if "api_vulnerability" in data.contains_type %}
+    <li>{% trans %}We identified potential API security vulnerabilities:{% endtrans %}
+        <ul>
+            {% for report in data.reports %}
+                {% if report.report_type == "api_vulnerability" %}
+                    <li>
+                        {{ report.target }}
+                        <ul>
+                            <li>{% trans %}Type:{% endtrans %} <strong>{{ report.additional_data.vulnerability_type|replace('_', ' ')|title }}</strong></li>
```
kazet: can you add a comment here what are the possible types?

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
```diff
@@ -0,0 +1,30 @@
...
+                            <li>{% trans %}Method:{% endtrans %} {{ report.additional_data.method|upper }}</li>
+                            <li>{% trans %}Host:{% endtrans %} {{ report.additional_data.resolved_host }}</li>
+                            <li>{% trans %}Status Code:{% endtrans %} {{ report.additional_data.status_code }}</li>
+                            <li>{% trans %}Details:{% endtrans %} {{ report.additional_data.details }}</li>
```
kazet: what form is details in? can it be passed directly to scanned entities?

artemis/reporting/modules/api_scanner/reporter.py
dr4g0n369: No, you are right, it is not necessary here.

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
dr4g0n369: The details are actually just a simple description of where the vulnerability was found. Here is an example
image

dr4g0n369: This comes straight from offat. The various `vuln_details` are given [here](https://github.com/OWASP/OFFAT/blob/dev/src/offat/tester/generator.py) for different vulnerabilities.

artemis/modules/api_scanner.py
dr4g0n369: I think we can use this one from seclists. [Swagger.txt](https://github.com/danielmiessler/SecLists/blob/master/Discovery/Web-Content/Service-Specific/Swagger.txt). It is MIT Licensed and also not too long.

kazet: awesome!

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
kazet: so, these details are basically True=we found something, False=we didn't? I think we may skip it then

dr4g0n369: Yes. But then for now, we will only have `method`, `curl_command` and `status_code` in the report. 

Here is an example output we receive from offat for a single result.
```json
{"test_name": "SQLi Test in URI Path with Fuzzed Params", 
"url": "http://localhost:5000/api/user/' AND (SELECT * FROM (SELECT(SLEEP(5)))abc)", 
"endpoint": "/api/user/' AND (SELECT * FROM (SELECT(SLEEP(5)))abc)", 
"method": "GET", 
"body_params": [], 
"query_params": [], 
"path_params": [{"name": "username", "value": "'m>oFaO<-U"}, {"name": "username", "value": "'m>oFaO<-U"}, {"name": "username", "value": "'m>oFaO<-U"}, {"name": "username", "value": "'m>oFaO<-U"}, {"name": "username", "value": "'m>oFaO<-U"}], 
"malicious_payload": "' AND (SELECT * FROM (SELECT(SLEEP(5)))abc)", 
"vuln_details": "Endpoint might be vulnerable to SQli", 
"success_codes": [500], 
"response_filter": "STATUS_CODE_FILTER", 
"request_headers": {"Host": "localhost:5000", "Accept": "*/*", "Accept-Encoding": "gzip, deflate", "User-Agent": "Python/3.12 aiohttp/3.12.11"}, 
"response_headers": {"Server": "Werkzeug/3.1.3 Python/3.10.18", "Date": "Mon, 09 Jun 2025 12:39:29 GMT", "Content-Type": "text/html; charset=utf-8", "Content-Length": "265", "Connection": "close"}, 
"response_body": "<!doctype html>\n<html lang=en>\n<title>500 Internal Server Error</title>\n<h1>Internal Server Error</h1>\n<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>\n", 
"response_status_code": 500, 
"redirection": "()", 
"error": false, 
"curl_command": "curl -X GET \"http://localhost:5000/api/user/' AND (SELECT * FROM (SELECT(SLEEP(5)))abc)?\" \"Host: localhost:5000\" -H \"Accept: */*\" -H \"Accept-Encoding: gzip, deflate\" -H \"User-Agent: Python/3.12 aiohttp/3.12.11\"", 
"vulnerable": true, 
"data_leak": {}}
```
 
 What other parameters do you think we should include in the report??

dr4g0n369:
I have implemented the requested changes. Please review it.

artemis/modules/api_scanner.py
```diff
@@ -0,0 +1,162 @@
...
+    def discover_spec(self, base_url: str) -> Tuple[str, ...]:
+        """Try to discover OpenAPI/Swagger specification from common paths."""
+        for path in COMMON_SPEC_PATHS:
+            try_url = base_url.rstrip("/") + "/" + path
+            try:
+                response = http_requests.get(try_url)
+                if response.status_code == 200 and (
+                    "openapi" in response.text.lower() or "swagger" in response.text.lower()
+                ):
+                    temp_file = f"/tmp/api_spec_{os.urandom(8).hex()}"
```
kazet, line 61: We need to remove the temp file so that we don't pollute the filesystem

artemis/modules/data/api_scanner_data.py
```diff
@@ -0,0 +1,18 @@
...
+# Common LFI payloads
+with open(f"{dirname(__file__)}/swagger.txt", "r") as f:
+    COMMON_SPEC_PATHS = f.read().splitlines()
+    COMMON_SPEC_PATHS = [payload.strip() for payload in COMMON_SPEC_PATHS if not payload.startswith("#")]
+
+VULN_DETAILS_MAP = {
+    "Endpoint performs HTTP verb which is not documented": "Unsupported HTTP Method",
+    "One or more parameter is vulnerable to SQL Injection Attack": "SQL Injection",
+    "Endpoint might be vulnerable to SQli": "SQL Injection",
+    "Endpoint might be vulnerable to BOLA": "BOLA",
```
kazet, line 11: I needed to google this acronym ;) so I think some explanation is needed - and since we use explanations, the VULN_DETAILS_MAP can be moved to the template so that we can use {% trans %} tag to translate the explanation

artemis/reporting/modules/api_scanner/reporter.py
```diff
@@ -0,0 +1,74 @@
...
+                            "endpoint": result.get("endpoint"),
+                            "data_leak": result.get("data_leak"),
+                            "details": result.get("vuln_details"),
+                            "curl_command": result.get("curl_command"),
+                            "status_code": result.get("status_code"),
+                        },
+                        timestamp=task_result["created_at"],
+                    )
+                )
+            except (KeyError, AttributeError):
```
kazet: why AttributeError?

artemis/modules/data/api_scanner_data.py
```diff
@@ -0,0 +1,18 @@
...
+    "Endpoint might be vulnerable to BOPLA": "BOPLA",
+    "One or more parameter is vulnerable to OS Command Injection Attack": "OS Command Injection",
+    "One or more parameter is vulnerable to XSS/HTML Injection Attack": "XSS/HTML Injection",
```
kazet, line 14: are you sure the check for XSS/HTML injection checks for the resulting content-type? ;)

test/data/flask_vulnerable_api/openapi.json
```diff
@@ -0,0 +1,125 @@
...
+        "description": "Retrieves user info based on username in URL path.\nSQL injection possible via path parameter.",
+        "parameters": [
+          {
+            "in": "path",
+            "name": "username",
+            "required": true,
+            "schema": {
+              "type": "string"
+            },
+            "example": "admin"
```
kazet: Just to be sure, can you change the examples to ones that don't actually exist in the app? ;)

artemis/modules/data/api_scanner_data.py
```diff
@@ -0,0 +1,18 @@
+from os.path import dirname
+
+# Common LFI payloads
+with open(f"{dirname(__file__)}/swagger.txt", "r") as f:
+    COMMON_SPEC_PATHS = f.read().splitlines()
+    COMMON_SPEC_PATHS = [payload.strip() for payload in COMMON_SPEC_PATHS if not payload.startswith("#")]
+
+VULN_DETAILS_MAP = {
+    "Endpoint performs HTTP verb which is not documented": "Unsupported HTTP Method",
+    "One or more parameter is vulnerable to SQL Injection Attack": "SQL Injection",
```
kazet, line 9: From my experience, just checking for 500 status code is too prone to false positives (e.g. even if there's no SQLi, the app might crash because there's no user with username `' or 1=1`). What do you think about slightly patching offat to check also for the response time for time-based SQLi? Or other method of filtering?

artemis/modules/data/api_scanner_data.py
```diff
@@ -0,0 +1,18 @@
...
+    "Endpoint might be vulnerable to SQli": "SQL Injection",
+    "Endpoint might be vulnerable to BOLA": "BOLA",
+    "Endpoint might be vulnerable to BOPLA": "BOPLA",
```
kazet, line 12: how confident are you that the tests for BOLA and BOPLA have a low risk of false positives?

artemis/modules/api_scanner.py
dr4g0n369: Already taken care of. It is there at line 156, it removes the `spec_file` once the scanning is done.

artemis/modules/api_scanner.py
```diff
@@ -0,0 +1,167 @@
...
+
+        try:
+            results = []
+            test_results = self.scan(spec_file)
+
+            for result in test_results.get("results", {}):
+                # Removing BOLA and BOPLA results as they are prone to False Positives
+                if (
+                    result.get("vulnerable", False)
+                    and "BOLA" not in result.get("vuln_details")
```
kazet: add a task to bring  it back and add a link to that task here ;)

artemis/modules/data/api_scanner_data.py
```diff
@@ -0,0 +1,6 @@
+from os.path import dirname
+
+# Common LFI payloads
```
kazet: misleading comment

artemis/modules/data/swagger.txt
```diff
@@ -0,0 +1,65 @@
+# Copied from MIT-licensed (https://github.com/danielmiessler/SecLists/blob/master/LICENSE)
+# file https://github.com/danielmiessler/SecLists/blob/master/Discovery/Web-Content/Service-Specific/Swagger.txt
+swagger/index.html
```
kazet, line 3: is thee a particular reason you changed the ordering?

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
```diff
@@ -0,0 +1,52 @@
...
+                        {{ report.target }}
+                        <ul>
+                            <li>{% trans %}Method:{% endtrans %} {{ report.additional_data.method|upper }}</li>
+                            <li>{% trans %}Endpoint:{% endtrans %} {{ report.additional_data.endpoint }}</li>
+                            <li>{% trans %}Status Code:{% endtrans %} {{ report.additional_data.status_code }}</li>
+                            <li>
+                                {% trans %}Details:{% endtrans %}
+                                {% set raw_detail = report.additional_data.details %}
+                                {% if "HTTP verb" in raw_detail %}
+                                    {% trans %}Unsupported HTTP Method{% endtrans %}
```
kazet: can you add some more text on why it's a vulnerabilityi?

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
```diff
@@ -0,0 +1,52 @@
...
+                                {% elif "vulnerable to SQL Injection" in raw_detail or "vulnerable to SQli" in raw_detail %}
+                                    {% trans %}SQL Injection{% endtrans %}
+                                {% elif "time-based SQL Injection" in raw_detail %}
+                                    {% trans %}Time-based SQL Injection{% endtrans %}
+                                {% elif "OS Command Injection" in raw_detail %}
+                                    {% trans %}OS Command Injection{% endtrans %}
+                                {% elif "XSS/HTML Injection" in raw_detail %}
+                                    {% trans %}XSS/HTML Injection{% endtrans %}
```
kazet: from what I see you don't check for the response code here, do you?

artemis/modules/data/swagger.txt
dr4g0n369: I did not change the ordering but I did add a few more payloads which were present in the initial commit but not in the wordlist at the end of this file. But it lead to some payloads being duplicated (eg: `api-docs`). Will fix this quickly.

artemis/modules/api_scanner.py
kazet: from what I see, it will be removed only if it's a real swagger file, it will be kept if it's something went wrong (e.g. validation failed). Or I'm missing somthing?

artemis/modules/data/offat/offat_artemis.patch
```diff
@@ -0,0 +1,224 @@
...
++        for sqli_payload in basic_sqli_payloads+sql_injection_sleep_payloads:
+             for request_obj in fuzzed_request_list:
+                 # handle body request params
+                 body_request_params = request_obj.get('body_params', [])
+@@ -305,6 +314,8 @@ class TestGenerator:
+                 request_obj[
+                     'response_filter'
+                 ] = PostTestFiltersEnum.STATUS_CODE_FILTER.name
++
++                request_obj["sleep_paylaod"] = sqli_payload in sql_injection_sleep_payloads
```
kazet: paylaod -> payload

artemis/modules/data/offat/offat_artemis.patch
```diff
@@ -0,0 +1,224 @@
...
++
++        for result in results:
++            test_name = result.get("test_name", "")
++            is_sql_test = "SQLi" in test_name or "sql" in test_name.lower()
++            sleep_payload = result.get("sleep_payload", False)
++            response_time = result.get("response_time")
++
++            new_result = deepcopy(result)
++
++            if is_sql_test and sleep_payload and isinstance(response_time, (int, float)):
```
kazet, line 137: I'm afraid this is not correct - this will not perform this check if it's not a sleep payload, so when entering e.g. `'` as the payload, we will still treat error 500 as a sign of vulnerability. Can we change this to have only sleep payloads?

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
```diff
@@ -0,0 +1,52 @@
...
+                            {% endif %}
+                            {% if report.additional_data.curl_command %}
+                                <li>{% trans %}CURL Command:{% endtrans %} <code>{{ report.additional_data.curl_command }}</code></li>
+                            {% endif %}
+                        </ul>
+                    </li>
+                {% endif %}
+            {% endfor %}
+        </ul>
+        <p>
```
kazet: I don't think this paragraph adds anything, it's too generic - I think the per-type descriptions are enough

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
```diff
@@ -0,0 +1,52 @@
...
+                                {% elif "vulnerable to SQL Injection" in raw_detail or "vulnerable to SQli" in raw_detail %}
+                                    {% trans %}SQL Injection - Malicious input can manipulate database queries, leading to data leaks or unauthorized access.{% endtrans %}
+                                {% elif "time-based SQL Injection" in raw_detail %}
+                                    {% trans %}Time-based SQL Injection - An attacker can infer data using timing-based payloads, exploiting database behavior.{% endtrans %}
+                                {% elif "OS Command Injection" in raw_detail %}
+                                    {% trans %}OS Command Injection - Arbitrary system commands can be executed on the server, leading to full compromise.{% endtrans %}
+                                {% elif "XSS/HTML Injection" in raw_detail %}
+                                    {% trans %}XSS/HTML Injection - Untrusted input is rendered in the browser, enabling session hijacking or defacement.{% endtrans %}
```
kazet: I think we may add a sentence or two describing the mechanism of defacement - why reflected xss leads to defacement ;)

artemis/reporting/modules/api_scanner/template_api_vulnerability.jinja2
```diff
@@ -0,0 +1,52 @@
...
+                                {% elif "SSTI" in raw_detail %}
+                                    {% trans %}SSTI - Server-Side Template Injection allows attackers to execute code on the server by injecting into templates.{% endtrans %}
+                                {% elif "fails to implement security authentication" in raw_detail %}
+                                    {% trans %}Missing Authentication - Critical endpoints lack proper access control, allowing unauthorized access.{% endtrans %}
```
kazet: critical ednpoints -> simply endpoints, as we don't know how critical they ae

test/data/flask_vulnerable_api/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Use an official Python base image
+FROM python:3.10-slim
+
+# Set working directory inside the container
+WORKDIR /app
+
+# Copy project files into the container
+COPY openapi.json openapi.json
+COPY app.py .
```
kazet: the file doesn't build ;)

docker-compose.test.yaml
```diff
@@ -69,6 +69,13 @@ services:
     environment:
       MYSQL_ROOT_PASSWORD: root
 
+  test-flask-vulnerable-api:
+    build:
+      context: .
+      dockerfile: test/data/flask_vulnerable_api/Dockerfile
+    ports:
+      - 5000:5000
```
kazet: I don't think this directive is needed, this will exposed the port on the host - if you skip it, the port will still be exposed on the container (but not on the host)

kazet:
image

I think the message (found potential vulnerabilities in the api) should describe the vulnerabilities at least briefly :)

kazet:
Artemis supports request rate limitting and user agent check, I think these need to be passed to offat (like here: https://github.com/CERT-Polska/Artemis-modules-extra/blob/main/karton_sqlmap/karton_sqlmap.py#L61 lines 61-80)

PR #1760, opened 2025-06-05
Assets

kazet:


artemis/modules/classifier.py
```diff
@@ -251,9 +251,7 @@ def run(self, current_task: Task) -> None:
                 },
             )
             self.add_task(current_task, new_task)
-            self.db.save_task_result(
```
bulkowy, line 254: going through files from up to bottom - are you sure that this will not create regression? It now exports different fields

artemis/modules/subdomain_enumeration.py
```diff
@@ -158,9 +158,12 @@ def get_subdomains_by_dns_brute_force(self, domain: str) -> Optional[Set[str]]:
...
         #
         # The number here (100) is not a mistake - we observed that there might be a large number of possible results.
-        results_for_random_subdomain = [
-            tuple(lookup(binascii.hexlify(os.urandom(5)).decode("ascii") + "." + domain)) for _ in range(100)
-        ]
+        try:
+            results_for_random_subdomain = [
+                tuple(lookup(binascii.hexlify(os.urandom(5)).decode("ascii") + "." + domain)) for _ in range(100)
+            ]
+        except ResolutionException:
```
bulkowy: hm, won't it break on first invalid resolution? Or is this exception only related on errors connecting to resolver?

kazet: done

artemis/modules/classifier.py
kazet: never was used

PR #1786, opened 2025-06-17
Tests for LFI, more strings to check

kazet:


artemis/modules/lfi_detector.py
```diff
@@ -63,24 +62,30 @@ def scan(self, urls: List[str], task: Task) -> List[Dict[str, Any]]:
...
-                                "headers": {},
-                                "matched_indicator": indicator,
-                                "statement": "It appears that this URL is vulnerable to LFI: " + url_with_payload,
-                                "code": LFIFindings.LFI_VULNERABILITY.value,
-                            }
-                        )
-                        if Config.Modules.LFIDetector.LFI_STOP_ON_FIRST_MATCH:
-                            return messages
+
+                    if len(url_with_payload) >= 1600 or i == len(URL_PARAMS) - 1:
```
bulkowy, line 77: this second boolean statement seems unnecessary? `i, _ := list(enumerate(x))[-1]; i == len(URL_PARAMS) -1`

bulkowy: dont get it as well why it is only prompted when url is long enough, you are sure it should not be reversed?

bulkowy (approved):

PR #1790, opened 2025-06-19
Asset discovry fix

kazet:


test/e2e/test_exporting.py
```diff
@@ -116,6 +116,7 @@ def test_exporting_gui(self) -> None:
 
             with export.open("advanced/output.json", "r") as f:
                 output_data = json.loads(f.read().decode("ascii"))
+                self.assertEqual(list(output_data["messages"].keys()), ["test-smtp-server.artemis"])
```
bulkowy, line 119: this is incorrect I guess

kazet: why?

bulkowy: should not `messages` keys contain more fields?

kazet: but this is the only top level target getting scanned, what other ones would you expect?

PR #1799, opened 2025-06-24
Removing Nuclei template as the fixes have been backported

kazet:


bulkowy (approved):

rxerium:
Hey @kazet @bulkowy, 

I’ve updated the template with improvements for version future-proofing; now relies on automatic DSL matchers as opposed to maintaining a list of vulnerable versions. Let me know your thoughts:
https://github.com/rxerium/CVE-2025-49113/blob/main/template.yaml

Many thanks,

Rishi

danielkalinka:
> Hey @kazet @bulkowy,
> 
> I’ve updated the template with improvements for version future-proofing; now relies on automatic DSL matchers as opposed to maintaining a list of vulnerable versions. Let me know your thoughts: https://github.com/rxerium/CVE-2025-49113/blob/main/template.yaml
> 
> Many thanks,
> 
> Rishi

@rxerium 

Thanks for your comment.
Several Linux distributions apply backported patches to Roundcube without updating the `rcversion` value. As a result, checking for vulnerabilities based solely on the software version often leads to false positives.

A good example is Debian, where the fix was released in the package `roundcube 1.4.15+dfsg.1-1+deb11u5`, but the code still contains `rcversion: 10415`. In some cases, a suffix may be added in the code to indicate a backport, but this is not always the case.

Thank you for your efforts - we're also exploring how to reliably detect vulnerable instances without generating so many false positives.

Best regards,
Daniel

Issue #552, opened 2023-09-28
Don't run all Nuclei templates, but only the ones related to the webapp we are testing


kazet:
Is `nuclei --automatic-scan` a solution?

PR #1813, opened 2025-07-01
Optimize nuclei

dr4g0n369:
Adds technology specific scanning to nuclei module. For now, it only filters for `wordpress`.

Related issue: #552


artemis/config.py
```diff
@@ -838,6 +838,12 @@ class Nuclei:
                 "NUCLEI_CHUNK_SIZE is 200, three calls will be made with 200 templates each.",
             ] = get_config("NUCLEI_CHUNK_SIZE", default=200, cast=int)
 
+            NUCLEI_RUN_TECH_DETECTION: Annotated[
```
kazet: I don't think this needs to be configurable - adding too many configuration options can also be a burden (for developers but also for users)

artemis/modules/data/wappalyzer/main.go
```diff
@@ -0,0 +1,79 @@
+package main
```
kazet: hmm, I don't think this belongs to modules/data as it's code, maybe modules/utils?

artemis/modules/data/wappalyzer/main.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
```
kazet: is this file copied from somewhere? if yes, I'd add link and license infromation

artemis/modules/data/wappalyzer/main.go
```diff
@@ -0,0 +1,79 @@
...
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Println("Usage: go run . <path_to_urls_file>")
+		os.Exit(1)
+	}
+	filePath := os.Args[1]
+
+	file, err := os.Open(filePath)
+	if err != nil {
+		log.Fatalf("FATAL: Could not open file %s: %v\n", filePath, err)
```
kazet: some minor discrepancy between this log (having \n) and the next one (not having)

artemis/modules/data/wappalyzer/main.go
```diff
@@ -0,0 +1,79 @@
...
+	result := make(map[string][]string)
+
+	resp, err := http.DefaultClient.Get(url)
+	if err != nil {
+		log.Printf("Error fetching %s: %v\n", url, err)
+		return result
+	}
+	defer resp.Body.Close()
+
+	data, _ := io.ReadAll(resp.Body) // You can handle error here if needed
```
kazet: I don't think the `// You can handle error here if needed` comment is rquired ;)

artemis/modules/nuclei.py
```diff
@@ -164,6 +166,38 @@ def __init__(self, *args: Any, **kwargs: Any):
             for line in open(Config.Modules.Nuclei.NUCLEI_TEMPLATES_TO_SKIP_PROBABILISTICALLY_FILE):
                 self._nuclei_templates_or_workflows_to_skip_probabilistically_set.add(line.strip())
 
+    def _run_tech_detection(self, urls: List[str]) -> Any:
```
kazet: I think the run_tech_detection can be a function oudside of Nuclei as I can already imagine some uses for that outside of Nuclei ;)

artemis/modules/nuclei.py
```diff
@@ -164,6 +166,38 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+            wappalyzer_output = subprocess.check_output(
+                ["go", "run", main_go_path, temp_file_name], cwd=wappalyzer_path
+            )
+            # The output is a mapping from URL to a list of detected app names
+            if os.path.exists(temp_file_name):
+                os.remove(temp_file_name)
+            return json.loads(wappalyzer_output)
+        except (subprocess.CalledProcessError, FileNotFoundError, json.JSONDecodeError) as e:
+            self.log.error(f"Error running technology detection: {e}")
+            if os.path.exists(temp_file_name):
```
kazet: you can extract into the finally: block and therefore you won't have to repeat yoursef ;)

artemis/modules/nuclei.py
```diff
@@ -412,10 +453,48 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
+
+                # Use a hashable frozenset as the dictionary key
+                scan_groups[frozenset(tags_to_exclude)].append(target_url)
+
+            findings: List[Dict[str, Any]] = []
+            for tags_frozen_set, group_targets in scan_groups.items():
+                extra_args = []
+                if tags_frozen_set:
+                    self.log.info(f"For {len(group_targets)} targets, excluding tags: {tags_frozen_set}")
+                    extra_args = ["-etags", ",".join(tags_frozen_set)]
```
kazet: hmm, are you sure -etags will be taken into account if the templates are provided manually as a list?

artemis/modules/data/wappalyzer/main.go
dr4g0n369: No, it was not copied. It was a simple script so I wrote it myself :)

artemis/modules/nuclei.py
dr4g0n369: Yes, I have checked that, `-etags` do work even when the templates have been manually listed.

Here is an example. It does not load the `wordpress` template due to the `etags` being specified.
image

artemis/module_base.py
```diff
@@ -309,6 +311,36 @@ def loop(self) -> None:
         else:
             self.log.info("Exiting loop, shutdown=%s", self.shutdown)
 
+    def _run_tech_detection(self, urls: List[str]) -> Any:
```
kazet: I would move this to a separate Python file so that module_base doesn't grow too big

artemis/module_base.py
```diff
@@ -309,6 +311,36 @@ def loop(self) -> None:
...
+        """
+        Run technology detection on a list of URLs using Wappalyzer.
+        """
+        wappalyzer_path = os.path.join(os.path.dirname(__file__), "modules", "utils", "wappalyzer")
+        main_go_path = os.path.join(wappalyzer_path, "main.go")
+        if not os.path.exists(main_go_path):
```
kazet: that can be an exception, as lack of important files is a problem we don't want to silently skip

artemis/module_base.py
```diff
@@ -309,6 +311,36 @@ def loop(self) -> None:
...
+            self.log.error(f"Wappalyzer main.go not found at {main_go_path}")
+            return {url: [] for url in urls}
+
+        try:
+            # Update the Wappalyzer package once
+            subprocess.run(["go", "get", "-u", "./..."], cwd=wappalyzer_path, check=True, capture_output=True)
+
+            temp_file_name = "/tmp/temp_urls.txt"
```
kazet: hm, why not Python tempfile module, in particular NamedTemporarFile?

artemis/modules/nuclei.py
```diff
@@ -412,10 +422,43 @@ def run_multiple(self, tasks: List[Task]) -> None:
 
         self.log.info(f"running {len(templates)} templates and {len(self._workflows)} workflow on {len(tasks)} hosts.")
 
-        targets = []
+        targets: List[str] = []
         for task in tasks:
             targets.append(get_target_url(task))
 
+        tech_results = run_tech_detection(targets, self.log)
+
```
kazet: lines 428-448 are quite complex, what do you think about extracting them to a function and adding a test for them?

in docker-compose.test.yml we already have multiple services - wordpress, joomla, ... to choose from for technlogy detection

dr4g0n369: Sure, I will start working on it

test/modules/test_nuclei.py
```diff
@@ -73,3 +74,18 @@ def test_links(self) -> None:
             "values in the server response via GET-requests., [medium] http://test-php-xss-but-not-on-homepage:80/xss.php: Fuzzing Parameters - Cross-Site Scripting Cross-site scripting "
             "was discovered via a search for reflected parameter values in the server response via GET-requests.\n",
         )
+
+    def test_group_targets_by_tech(self) -> None:
```
kazet: one minor comment: I would call it group_targets_by_missing_tech and precisely describe in the docstring that they describe the target they **lack**

artemis/web_technology_identification.py
```diff
@@ -0,0 +1,34 @@
+import json
+import logging
+import os
+import subprocess
+import tempfile
+from typing import Any, List
+
+
+def run_tech_detection(urls: List[str], logger: logging.Logger) -> Any:
```
kazet, line 9: can you also write a small unittest for this function?

artemis/web_technology_identification.py
```diff
@@ -0,0 +1,34 @@
...
+        raise FileNotFoundError(f"Wappalyzer main.go not found at {main_go_path}")
+
+    try:
+        # Update the Wappalyzer package once
+        subprocess.run(["go", "get", "-u", "./..."], cwd=wappalyzer_path, check=True, capture_output=True)
+
+        with tempfile.NamedTemporaryFile(mode="w") as temp_file:
+            for url in urls:
+                temp_file.write(url + "\n")
+            temp_file.flush()
```
kazet, line 25: I would add `os.fsync(temp_file.fileno())` just in case

test/modules/test_nuclei.py
```diff
@@ -73,3 +74,18 @@ def test_links(self) -> None:
...
+        targets = [
+            "http://test-old-wordpress",
+            "http://test-old-joomla",
+            "http://test-flask-vulnerable-api:5000",
+        ]
+        logger = logging.Logger("test_logger")
+
+        grouped_targets = group_targets_by_tech(targets, logger)
```
kazet: why not just check for equality?

PR #1837, opened 2025-07-10
Technology detection tags

kazet:


artemis/reporting/modules/webapp_identifier/reporter.py
```diff
@@ -0,0 +1,52 @@
...
+            if technology == "WordPress":  # we have separate modules for that
+                result.append(
+                    Asset(
+                        asset_type=AssetType.CMS,
+                        name=task_result["target_string"],
+                        additional_type="wordpress",
+                        version=version,
+                    )
+                )
+            elif technology == "Joomla":  # we have separate modules for that
```
bulkowy: dunno if we want to solve this in this if/elif/else tree. What I would like to see is that we have some dict or additional fields in task_results that holds `asset_type`, and then `additional_type` would be filled with `technology.lower()`. It would make it a lot easier to maintain

bulkowy (approved):

PR #1852, opened 2025-07-22
Speedup artemis - Nuclei module

dr4g0n369:
This moves the code for updating nuclei templates to `docker/Dockerfile` so that it is run only in the beginning. 
Nuclei is heavily used so updating the templates in every single run slows down the scanning process.

## Changes made:
- Moved `update-templates` to `docker/Dockerfile`
- Moved cloning `KEV` repo to `docker/Dockerfile`
- Running templates list command only once for `log_exposures` and `exposed_panels`


artemis/modules/nuclei.py
```diff
@@ -106,22 +103,10 @@ def get_default_configuration(self) -> NucleiConfiguration:
     def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
 
-        # We clone this repo in __init__ (on karton start) so that it will get periodically
-        # re-cloned when the container gets retarted every 𝑛 tasks. The same logic lies behind
-        # updating the Nuclei templates in __init__.
-        if os.path.exists("/known-exploited-vulnerabilities/"):
-            shutil.rmtree("/known-exploited-vulnerabilities/")
+        # We are cloning the KEV repository and updating templates in the Dockerfile
```
kazet, line 111: how much time we are saving with this PR?

(not a precise value, just a ballpark)

dr4g0n369: I ran the tests locally, once with this update and then without
These were the results:

With the update
```
Ran 4 tests in 819.036s
```

Without the update
```
Ran 4 tests in 875.879s
```

So that makes it `56.843s` difference, so rounding off, we can say it is about `1 min`. Considering there were 4 tests, we save nearly `15s` every time nuclei is run.

dr4g0n369: Well, regarding this update, I also had another implementation in mind. We could keep the previous code but before updating templates we just do a simple check to see if we are running an outdated version and update the templates or clone the `KEV` only then. 

The only drawback for this might be that we might have to make an extra request in order to check the latest version, although it should still help in improving the speed it will only be slightly less than the one in this PR right now.

Let me know if you would prefer this implementation.

dr4g0n369:
I have made the changes so that it keeps updating the templates at an interval of `7 days`. 

Also ran the tests and it was still close to the time I got in the previous implementation.

```
Ran 4 tests in 826.646s
```

PR #1869, opened 2025-08-06
Implementing DAST nuclei templates

dr4g0n369:


artemis/modules/nuclei_dast.py
```diff
@@ -0,0 +1,435 @@
+#!/usr/bin/env python3
+import collections
```
kazet: I don't think copying the whole file is a good idea - it will make it very easy to make mistakes, where a fix is applied in one but not the second one. What about inheritance or any other way of extracting common code into one place?

test/modules/test_nuclei_dast.py
```diff
@@ -0,0 +1,47 @@
...
+    # The reason for ignoring mypy error is https://github.com/CERT-Polska/karton/issues/201
+    karton_class = NucleiDast  # type: ignore
+
+    def test_lfi_dast_template(self) -> None:
+        task = Task(
+            {"type": TaskType.SERVICE.value, "service": Service.HTTP.value},
+            payload={
+                "host": "test-flask-vulnerable-api",
+                "path": "/files",
+                "params": {"filename": "randomfilename"},
```
kazet: that will fail in real production usage, as the 'params' variable will never be filled.

I think the module itself should be responsible for filling the variable (the code for generating params is present e.g. in the xss_scanner module, it can be extracted into a function)

artemis/modules/data/dast_params/lfi.txt
```diff
@@ -0,0 +1,2 @@
+# Only for testing for now
+filename
```
kazet, line 10: todo: add something (e.g. from seclists)

artemis/modules/data/dast_params/ssrf.txt
```diff
@@ -0,0 +1,27 @@
+# Taken from https://github.com/lutfumertceylan/top25-parameter/blob/master/ssrf-parameters.txt
+# Apache License: https://github.com/lutfumertceylan/top25-parameter/blob/master/LICENSE
+dest
```
kazet, line 3: hm, why not top50? we don't have space in the url?

test/data/flask_vulnerable_api/app.py
```diff
@@ -28,6 +29,9 @@ def init_db() -> None:
 
 @app.route("/")
 def index() -> Tuple[Response, int]:
+    filename = request.args.get("filename", "")
```
kazet: I think it would be cleaner to have a separate test app - I'm afraid this one can grow too big

artemis/utils.py
```diff
@@ -117,3 +118,21 @@ def get_host_from_url(url: str) -> str:
 def read_template(path: str) -> str:
     with open(Path(__file__).parent.parent / "templates" / path) as f:
         return f.read()
+
+
+def add_common_params_from_wordlist(url: str, params_wordlist: str, default_param_value: str = "") -> str:
```
kazet, line 123: this function can have an unittest or two ;)

artemis/modules/nuclei.py
```diff
@@ -457,15 +481,21 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
-            targets.append(get_target_url(task))
-
         scan_groups = group_targets_by_missing_tech(targets, self.log)
         found_targets_after_grouping = []
         for scan_group in scan_groups.values():
             found_targets_after_grouping.extend(scan_group)
         assert set(found_targets_after_grouping) == set(targets)
 
+        for template_name, template_data in DAST_SCANNING.items():
+            DAST_SCANNING[template_name]["targets"] = list()
```
kazet: I don't think modifying a global variable is a good idea, what about keeping DAST_SCANNING constant and collecting the target in a different one?

artemis/modules/nuclei.py
```diff
@@ -182,6 +194,18 @@ def __init__(self, *args: Any, **kwargs: Any):
...
             self._template_lists["custom"] = Config.Modules.Nuclei.NUCLEI_ADDITIONAL_TEMPLATES
 
+            dast_templates = check_output_log_on_error(["nuclei", "-dast", "-tl"], self.log).decode("ascii").split()
+            dast_templates = [
+                template
+                for template in dast_templates
+                if template.startswith("dast/") and template not in Config.Modules.Nuclei.NUCLEI_TEMPLATES_TO_SKIP
+            ]
+            self._dast_templates: Dict[str, List[str]] = {}
+            for template_name in DAST_SCANNING.keys():
```
kazet: the template_name variable could rather be called 'keyword' or something

artemis/modules/data/dast_params/ssrf.txt
dr4g0n369: No, that wasn't the issue. I simply didn't find any wordlist for top50 :(

kazet: you can mix it with a generic top50 list (sum these two lists, sort, uniq)

artemis/modules/nuclei.py
```diff
@@ -42,6 +43,17 @@
 
 TECHNOLOGY_DETECTION_CONFIG = {"wordpress": {"tags_to_exclude": ["wordpress"]}}
 
+DAST_SCANNING: Dict[str, Dict[str, Any]] = {
+    "ssrf": {
+        "params_wordlist": os.path.join(os.path.dirname(__file__), "data", "dast_params", "ssrf.txt"),
+        "param_default_value": "http://example.com",
```
kazet: use 127.0.0.1 so that we don't direct too much traffic to example.com

artemis/modules/nuclei.py
```diff
@@ -457,15 +479,22 @@ def run_multiple(self, tasks: List[Task]) -> None:
         self.log.info(f"running {len(templates)} templates and {len(self._workflows)} workflow on {len(tasks)} hosts.")
 
         targets: List[str] = []
-        for task in tasks:
```
kazet: hm, why this got removed?

artemis/modules/nuclei.py
```diff
@@ -497,6 +559,23 @@ def run_multiple(self, tasks: List[Task]) -> None:
                     [item for item in link_package if item],
                 )
             )
+            findings.extend(
+                self._scan(
+                    self._dast_templates["other"],
```
kazet: I don't think this one exists ;)

artemis/modules/nuclei.py
```diff
@@ -182,6 +210,19 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+            dast_templates = check_output_log_on_error(["nuclei", "-dast", "-tl"], self.log).decode("ascii").split()
+            dast_templates = [
+                template
+                for template in dast_templates
+                if template.startswith("dast/") and template not in Config.Modules.Nuclei.NUCLEI_TEMPLATES_TO_SKIP
+            ]
+            self._dast_templates: Dict[str, List[str]] = {}
+            for keyword in DAST_SCANNING.keys():
+                self._dast_templates[keyword] = [
+                    dast_templates.pop(i) for i in range(len(dast_templates)) if keyword in dast_templates[i]
```
kazet: Are you sure it's correct? the list will change during iteration

I think this can be writen in a simpler way `[item for item in dast_templates if keyword in item]`

and then set 'other' to `[item for item in dast_templates if item not in ...]`

test/modules/test_nuclei.py
```diff
@@ -110,3 +110,19 @@ def test_group_targets_by_missing_tech(self) -> None:
 
         self.assertIn(frozenset(["wordpress"]), grouped_targets)
         self.assertEqual(grouped_targets[frozenset(["wordpress"])], expected_results[frozenset(["wordpress"])])
+
+    def test_lfi_dast_template(self) -> None:
```
kazet: can you also add a test for one template from the 'other' group?

artemis/modules/nuclei.py
```diff
@@ -497,6 +559,34 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
+            for keyword, template_data in DAST_SCANNING.items():
+                dast_targets[keyword] = []
+                for item in link_package:
+                    if item:
+                        dast_targets[keyword].append(
+                            add_common_params_from_wordlist(
+                                item, template_data["params_wordlist"], template_data["param_default_value"]
+                            )
+                        )
+            for keyword in DAST_SCANNING:
```
kazet: here you have two loops, what about having one?

```
for keyword, template_data in DAST_SCANNING.items():
    collect targets for a given keyword
    run them
```

artemis/modules/nuclei.py
dr4g0n369: I could do that but then I would have to separately find templates for `other`. That is the reason I went with this as `pop` would add the templates to `keyword` at the same time removing them from the original one thus we are only left with `other` templates.

test/modules/test_nuclei.py
dr4g0n369: Sure, will add them

artemis/modules/nuclei.py
```diff
@@ -466,6 +507,16 @@ def run_multiple(self, tasks: List[Task]) -> None:
             found_targets_after_grouping.extend(scan_group)
         assert set(found_targets_after_grouping) == set(targets)
 
+        dast_targets: Dict[str, List[str]] = {}
+        for keyword, template_data in DAST_SCANNING.items():
```
kazet: do I see correctly that you don't run 'other' on the root site?

artemis/modules/nuclei.py
```diff
@@ -480,6 +531,17 @@ def run_multiple(self, tasks: List[Task]) -> None:
                 self._scan(self._workflows, ScanUsing.WORKFLOWS, group_targets, extra_nuclei_args=extra_args)
             )
 
+        # DAST scanning
```
kazet, line 544: here you have two loops, what about having one?

```
for keyword, template_data in DAST_SCANNING.items():
    collect targets for a given keyword
    run them
```

artemis/modules/nuclei.py
dr4g0n369: Ya, that is true. `other` includes templates without any params wordlist. So without any params, running them wouldn't help much.

kazet: can't we have a default wordlist for 'other'?

artemis/modules/nuclei.py
kazet: I think separately finding templates for other would be less prone to mistakes

dr4g0n369: Ya, you are right about this. I fixed the tests and when I ran them, this code ended up giving an error.

test/modules/test_nuclei.py
```diff
@@ -110,3 +111,26 @@ def test_group_targets_by_missing_tech(self) -> None:
...
+                "host": "test-dast-vuln-app",
+                "port": 5000,
+            },
+        )
+        self.run_task(task)
+        (call,) = self.mock_db.save_task_result.call_args_list
+        self.assertEqual(call.kwargs["status"], TaskStatus.INTERESTING)
+        self.assertEqual(
+            call.kwargs["status_reason"],
+            "[high] http://test-dast-vuln-app:5000: Joomla! Helpdesk Pro plugin <1.4.0 - Local File Inclusion Directory traversal vulnerability in the Helpdesk Pro plugin before 1.4.0 for Joomla! allows remote attackers to read arbitrary files via a .. (dot dot) in the filename parameter in a ticket.download_attachment task., "
```
kazet: can you mark this templte as suspicious in artemis/config.py? This is a standard flow if it trigges on a generic lfi

test/modules/test_nuclei.py
```diff
@@ -110,3 +111,26 @@ def test_group_targets_by_missing_tech(self) -> None:
...
...http%3A%2F%2F127.0.0.1&image_url=http%3A%2F%2F127.0.0.1&go=http%3A%2F%2F127.0.0.1&return=http%3A%2F%2F127.0.0.1&returnTo=http%3A%2F%2F127.0.0.1&return_to=http%3A%2F%2F127.0.0.1&checkout_url=http%3A%2F%2F127.0.0.1&continue=http%3A%2F%2F127.0.0.1&return_path=http%3A%2F%2F127.0.0.1: Reflected SSTI Arithmetic Based , [medium] http://test-dast-vuln-app:5000/ssti?template=Testing&q=testing&s=testing&search=testing&id=testing&lang=testing&keyword=testing&query=testing&page=testing&keywords=testing&year=testing&view=testing&email=testing&type=testing&name=testing&p=testing&month=testing&image=testing&list_type=testing&url=testing&terms=testing&categoryid=testing&key=testing&login=testing&begindate=testing&enddate=testing: Reflected Cross-Site Scripting , [medium] http://test-dast-vuln-app:5000/ssti?template=Testing&q=testing&s=testing&search=testing&id=testing&lang=testing&keyword=testing&query=testing&page=testing&keywords=testing&year=testing&view=testing&email=testing&type=testing&name=testing&p=testing&month=testing&image=testing&list_type=testing&url=testing&terms=testing&categoryid=testing&key=testing&login=testing&begindate=testing&enddate=testing: Reflected SSTI Arithmetic Based ",
```
kazet: shouldn't the url include the actual payload?

artemis/modules/nuclei.py
```diff
@@ -480,6 +521,36 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
+        dast_targets["other"] = []
+        for keyword, template_data in DAST_SCANNING.items():
+            dast_targets[keyword] = []
+            for task in tasks:
+                param_url = add_common_params_from_wordlist(
+                    get_target_url(task), template_data["params_wordlist"], template_data["param_default_value"]
+                )
+                dast_targets[keyword].append(param_url)
+                # Since there are no params for "other", we can just combine all the params from the different vulns into this
+                dast_targets["other"].append(param_url)
```
kazet: I'm afraid this wont work - look, there will be 6 times as much URLs for 'other' than for each of the keywords. Can it be made faster, by putting 'other' as a separate, single entry in DAST_SCANNING?

test/modules/test_nuclei.py
kazet: I'm afraid this one isn't already resolved

test/modules/test_nuclei.py
kazet: I'm afraid this one isn't already resolved

artemis/modules/nuclei.py
```diff
@@ -480,6 +521,48 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
+                    param_url, template_data["params_wordlist"], template_data["param_default_value"]
+                )
+            dast_targets["other"].append(param_url)
+        findings.extend(
+            self._scan(
+                self._dast_templates["other"], ScanUsing.TEMPLATES, dast_targets["other"], extra_nuclei_args=["-dast"]
+            )
+        )
+
+        findings.extend(
```
kazet, line 560: why do you run the DAST scan for other twice? ;)

artemis/modules/nuclei.py
```diff
@@ -480,6 +521,48 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
+                    extra_nuclei_args=["-dast"],
+                )
+            )
+
+        # Since there are no params for "other", we can just combine all the params from the different vulns into this
+        dast_targets["other"] = []
+        for task in tasks:
+            param_url = get_target_url(task)
+            for keyword, template_data in DAST_SCANNING.items():
+                param_url = add_common_params_from_wordlist(
```
kazet, line 549: this will create a very long url where the parameters repeat, am I right?

artemis/modules/nuclei.py
```diff
@@ -498,6 +581,44 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
+                    )
+                )
+
+            dast_targets["other"] = []
+            for item in link_package:
+                if item:
+                    param_url = item
+                    for keyword, template_data in DAST_SCANNING.items():
+                        param_url = add_common_params_from_wordlist(
+                            item, template_data["params_wordlist"], template_data["param_default_value"]
```
kazet: this will set param_url to the item with parameters added for the last dict entry only, right?

artemis/modules/nuclei.py
dr4g0n369: It will create a really long url but no, the parameters do not repeat

PR #1870, opened 2025-08-07
Add docker troubleshooting dns problem to docs

kadewu:
It can be encountered during tests run.
Cause of a lot of logs during test run it may be not easily noticeable - those I believe it's worth to be mentioned.

### DIFF
--- a/docs/user-guide/troubleshooting.rst
+++ b/docs/user-guide/troubleshooting.rst
@@ -63,3 +63,20 @@ Results disappeared after migrating from an old version of Artemis
 ------------------------------------------------------------------
 If you updated from an old version of Artemis and don't see the scanned targets or scan results anymore,
 contact us on Discord: https://discord.gg/GfUW4mZmy9.
+
+
+Running tests locally
+---------------------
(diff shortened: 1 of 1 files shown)


docs/user-guide/troubleshooting.rst
```diff
@@ -63,3 +63,20 @@ Results disappeared after migrating from an old version of Artemis
 ------------------------------------------------------------------
 If you updated from an old version of Artemis and don't see the scanned targets or scan results anymore,
 contact us on Discord: https://discord.gg/GfUW4mZmy9.
+
+
+Running tests locally
+---------------------
+If you encounter problem with DNS while running ``scripts/tests`` you may need to set it explicitly
```
kazet: can you provide an example of how these problems manifest?

docs/user-guide/troubleshooting.rst
```diff
@@ -63,3 +63,20 @@ Results disappeared after migrating from an old version of Artemis
...
+
+According to https://docs.docker.com/engine/network/; containers should use the same DNS servers as the host by default
```
kazet: I'd put a comma instead of a semicolon

PR #1871, opened 2025-08-07
bump aiohttp dependency in pre-commit 

kadewu:
I run into problem with wheel build on python 3.13.
This fixes it.


kazet:
what do you think about updating all of them?

kadewu:
> what do you think about updating all of them?

Yeah, good point, I will do that.
We may also consider if there are ways to bump those with dependanabot, but for now I will just bump it manually.

kazet (approved):

PR #1927, opened 2025-09-04
Add more information to docs about writing new modules

kadewu:
It's a starting point to increase documentation of Artemis.
We can start adding more docstrings while developing.


artemis/db.py
```diff
@@ -199,6 +199,13 @@ def save_task_logs(self, task_id: str, logs: bytes) -> None:
             session.commit()
 
     def create_analysis(self, analysis: Task) -> None:
+        """
+        Create a new :class:`~artemis.db.Analysis` entry in the database from a given Karton Task.
```
kazet: I would write what is an Analysis

artemis/db.py
```diff
@@ -215,6 +222,19 @@ def create_analysis(self, analysis: Task) -> None:
...
+        Save the result of a task execution to the database.
+
+        :param task: The Karton Task object for which the result is being saved.
+        :type task: Task
+        :param status: The status of the task (e.g., OK, INTERESTING, etc.).
+        :type status: TaskStatus
+        :param status_reason: Optional reason or message describing the status.
+        :type status_reason: Optional[str]
+        :param data: Optional result data, can be a Pydantic model, Exception, or any serializable object.
+        :type data: Optional[Any]
```
kazet, line 302: I would add an example how to use these fields - e.g. when to use INTERESTING and when OK

artemis/db.py
```diff
@@ -199,6 +199,13 @@ def save_task_logs(self, task_id: str, logs: bytes) -> None:
             session.commit()
 
     def create_analysis(self, analysis: Task) -> None:
```
kazet, line 268: I think documenting the core models is a good idea so that the user will be able to grasp the core concepts

artemis/task_utils.py
```diff
@@ -9,6 +9,25 @@
 
 
 def get_target_host(task: Task) -> str:
+    """
+    Extract the target host identifier from a Karton task based on its type.
+
+    This function extracts different identifiers depending on the task type:
```
kazet: IMO this paragraph is not needed, as it just describes the implementation. I would rather elaborate on the high-level purpose of the function - it takes the task and returns the host (domain/ip) it refers to

docs/user-guide/writing-a-module.rst
```diff
@@ -14,6 +14,8 @@ Feel free to copy this module to implement a real one. Remember to start the mod
 https://github.com/CERT-Polska/Artemis/blob/main/docker-compose.yaml.
 
 Since Artemis uses the Karton framework (https://github.com/CERT-Polska/karton) underneath, modules are Karton services.
+It is recommeneded to get familiar with (https://karton-core.readthedocs.io/en/latest/advanced_concepts.html) in order to
```
kazet, line 17: is this really needed before writing a new module? e.g. was it needed for you?

docs/user-guide/writing-a-module.rst
```diff
@@ -24,6 +26,12 @@ a new task.
 
 Since Artemis saves some additional task information in the database, you need to use a wrapper - ``self.add_task``.
 
+
+Utilities
+---------
+Modules often needs to obtain certain variables such as IP adresses, hostnames etc.
```
kazet, line 32: I would add more examples of what task_utils provide

artemis/db.py
kadewu: You mean to document Task model?

docs/user-guide/writing-a-module.rst
kadewu: It was helpful. Probably the most helpful was this paragraph: https://karton-core.readthedocs.io/en/latest/advanced_concepts.html#task-tree-analysis-and-task-life-cycle
But overall reading it helped me in understanding how the queue works etc.

artemis/db.py
kazet: the models we have in db, Task is a karton concept so we can link to the docs (is there a 'core concepts' karton page?)

artemis/db.py
kadewu: Added the reference to Artemis class.

artemis/db.py
kadewu: Added the reference to Enum

artemis/db.py
kadewu: It's now directly linked thanks to https://github.com/CERT-Polska/Artemis/pull/1927/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R81-R84

artemis/db.py
```diff
@@ -62,6 +62,16 @@ class TSVector(TypeDecorator):  # type: ignore
 
 
 class ScheduledTask(Base):  # type: ignore
+    """
+    Represents a scheduled task in the Artemis system.
```
kazet, line 66: write that this model is only for deduplication, not for real task scheduling

artemis/db.py
```diff
@@ -75,6 +85,18 @@ class ScheduledTask(Base):  # type: ignore
 
 
 class Analysis(Base):  # type: ignore
+    """
+    Represents an analysis entry in the Artemis system.
```
kazet, line 90: write what an analysis is (e.g. what it corresponds to in the UI)

artemis/db.py
```diff
@@ -125,6 +164,24 @@ class ReportGenerationTaskStatus(str, enum.Enum):
 
 
 class ReportGenerationTask(Base):  # type: ignore
+    """
+    Represents a report generation task in the Artemis system.
```
kazet, line 169: link to what a report is (we have docs for that)

artemis/db.py
```diff
@@ -92,6 +114,23 @@ class Analysis(Base):  # type: ignore
 
 
 class TaskResult(Base):  # type: ignore
+    """
+    Stores the result of a Karton :class:`~karton.core.task.Task` execution.
+
+    :ivar id: Unique identifier for the task result.
+    :ivar analysis_id: Identifier of the related analysis.
+    :ivar created_at: Timestamp when the result was created.
+    :ivar status: Status of the task (e.g., OK, ERROR, INTERESTING).
```
kazet: link to what a status is

artemis/db.py
```diff
@@ -215,6 +287,19 @@ def create_analysis(self, analysis: Task) -> None:
     def save_task_result(
         self, task: Task, *, status: TaskStatus, status_reason: Optional[str] = None, data: Optional[Any] = None
     ) -> None:
+        """
```
kazet, line 292: do I see correctly that these methods aren't put by Sphinx into the docs?

artemis/task_utils.py
```diff
@@ -9,6 +9,18 @@
 
 
 def get_target_host(task: Task) -> str:
+    """
+    Extract the target host identifier (domain, ip) from a Karton task based on its type.
```
kazet: I'd phrase it `target host (domain or ip)`

artemis/task_utils.py
```diff
@@ -9,6 +9,18 @@
...
+
+    :param task: A Karton task object
+    :type task: Task
+
+    :return: A string representing the target host identifier (hostname, domain name, or IP address)
+    :rtype: str
+
+    :raises ValueError: If the task type is unknown
+    :raises AssertionError: If the extracted payload is not of the expected type
```
kazet: maybe 'supported' not 'expected'?

docs/user-guide/writing-a-module.rst
```diff
@@ -24,6 +26,21 @@ a new task.
...
+The ``artemis.task_utils`` (see: :ref:`task_utils`) module provides such helpers for extracting and
+manipulating task-related data in Artemis modules.
+
+Example utilities:
+
+- ``artemis.task_utils.get_target_host``
+    Extracts the target host (domain, IP, or hostname) from a Karton task, based on its type.
```
kazet: just domain or ip ;)

artemis/db.py
kadewu: Yes, I wanted to avoid that annoying duplication in docs cause of :ivar and :undoc-members:
There is a way to provide strings for sphinx for attributes, but I'm not sure if that's the way, so for now I've added manually all the classes to `db.rst` file.

kazet (approved):

PR #1945, opened 2025-09-09
Robustness: if unable to get lins, no links

kazet:


artemis/modules/nuclei.py
```diff
@@ -251,7 +251,11 @@ def __init__(self, *args: Any, **kwargs: Any):
                 self._nuclei_templates_or_workflows_to_skip_probabilistically_set.add(line.strip())
 
     def _get_links(self, url: str) -> List[str]:
-        links = get_links_and_resources_on_same_domain(url)
+        try:
+            links = get_links_and_resources_on_same_domain(url)
+        except requests.exceptions.RequestException:
```
kadewu: I think it may be better to  catch that error in the method implementation? WDYT?
It's also used in other places in the code, so will solve potential problems in other places as well.

kadewu (dismissed):

PR #1958, opened 2025-09-15
Clean scheduledtask automatically

kadewu:
Scheduletask table is only used for deduplication. Once the analysis is done we can safely remove records from the DB.

### DIFF
--- a/artemis/cleanup.py
+++ b/artemis/cleanup.py
@@ -82,16 +82,20 @@ def _cleanup_scheduled_tasks() -> None:
         else:
             has_unfinished_analyses = True
 
-    if not has_unfinished_analyses and Config.Miscellaneous.CLEANUP_RAISE_ERR_ON_NON_UNFINISHED_ANALYSES:
+    if not has_unfinished_analyses and Config.Miscellaneous.CLEANUP_RAISE_ERROR_ON_NON_UNFINISHED_ANALYSES:
         raise AssertionError("Did not found unfinished analyses during cleanup.")
 
     if finished_analyses_ids:
-        # introducing batched to not overwhelm database
+        # introducing batches to not overwhelm database
         BATCH = 100
+        removed_rows = 0
         for i in range(0, len(finished_analyses_ids), BATCH):
             analysis_ids = finished_analyses_ids[i : i + BATCH]
-            db.delete_analysis_scheduled_tasks(analysis_ids)
-            logger.info("Cleaned up ScheduledTask table for analyses: %s", ",".join(analysis_ids))
+            rows = db.delete_scheduled_tasks_for_analyses(analysis_ids)
+            removed_rows += rows
+
+            logger.debug("Cleaned up ScheduledTask table for analyses: %s", ",".join(analysis_ids))
+        logger.info("Removed %d rows in ScheduleTask table.", removed_rows)
 
 
 def cleanup() -> None:
--- a/artemis/cleanup.py
+++ b/artemis/cleanup.py
@@ -8,6 +8,7 @@ from karton.core.config import Config as KartonConfig
 from karton.core.inspect import KartonState
 
 from artemis import utils
+from artemis.config import Config
 from artemis.db import DB
 
 logger = utils.build_logger(__name__)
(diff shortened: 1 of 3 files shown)


artemis/cleanup.py
```diff
@@ -67,9 +71,26 @@ def process(self, *args, **kwargs):  # type: ignore
...
 
+def _cleanup_scheduled_tasks() -> None:
+    karton_state = KartonState(backend=KartonBackend(config=KartonConfig()))
+    finished_analysis_ids = []
+    for analysis in db.list_analysis():
+        if analysis["id"] not in karton_state.analyses or len(karton_state.analyses[analysis["id"]].pending_tasks) == 0:
+            finished_analysis_ids.append(analysis["id"])
+
+    if finished_analysis_ids:
+        # introducing batched to not overwhelm database
```
kadewu: I'm not sure if the batch will be needed, but wanted to avoid situation when I'm sending too much id to the postgresql server resulting in too much cpu usage etc.

artemis/cleanup.py
```diff
@@ -67,9 +71,26 @@ def process(self, *args, **kwargs):  # type: ignore
         logger.info("Queue for %s is cleaned up", old_module)
 
 
+def _cleanup_scheduled_tasks() -> None:
+    karton_state = KartonState(backend=KartonBackend(config=KartonConfig()))
+    finished_analysis_ids = []
+    for analysis in db.list_analysis():
+        if analysis["id"] not in karton_state.analyses or len(karton_state.analyses[analysis["id"]].pending_tasks) == 0:
+            finished_analysis_ids.append(analysis["id"])
```
kadewu: Currently I'm not tracking if the analysis was already cleaned by the task - I believe due to archiving results it should not be needed?

kazet: yup, that's fine

artemis/cleanup.py
```diff
@@ -67,9 +71,26 @@ def process(self, *args, **kwargs):  # type: ignore
...
+
+    if finished_analysis_ids:
+        # introducing batched to not overwhelm database
+        BATCH = 100
+        for i in range(0, len(finished_analysis_ids), BATCH):
+            analysis_ids = finished_analysis_ids[i : i + BATCH]
+            db.delete_analysis_scheduled_tasks(analysis_ids)
+            logger.info("Cleaned up ScheduledTask table for analysis: %s", ",".join(analysis_ids))
```
kazet: for analyses

artemis/cleanup.py
kazet: IMO that's ok, let's err on the side of caution

artemis/cleanup.py
```diff
@@ -67,9 +71,26 @@ def process(self, *args, **kwargs):  # type: ignore
         logger.info("Queue for %s is cleaned up", old_module)
 
 
+def _cleanup_scheduled_tasks() -> None:
+    karton_state = KartonState(backend=KartonBackend(config=KartonConfig()))
+    finished_analysis_ids = []
+    for analysis in db.list_analysis():
+        if analysis["id"] not in karton_state.analyses or len(karton_state.analyses[analysis["id"]].pending_tasks) == 0:
+            finished_analysis_ids.append(analysis["id"])
+
```
kazet, line 84: add an assert we have also **unfinished** analyses ids, in case we made a mistake in lines 75-79 (e.g. the key format changed) ;)

kadewu: I've added additionaly a flag to control that behaviour, this way e.g: running it locally will not throw errors in case of zero activity.

artemis/config.py
```diff
@@ -315,6 +315,10 @@ class Miscellaneous:
             "Maximum number of URLs to scan per target for modules that crawl like lfi_detector, Nuclei, sq_injection_detector, etc.",
         ] = get_config("MAX_URLS_TO_SCAN", default=25, cast=int)
 
+        CLEANUP_RAISE_ERR_ON_NON_UNFINISHED_ANALYSES: Annotated[
```
kazet: "raise err" -> "raise error"

artemis/db.py
```diff
@@ -479,6 +480,12 @@ def save_scheduled_task(self, task: Task) -> bool:
             session.commit()
             return bool(result.rowcount)
 
+    def delete_analysis_scheduled_tasks(self, analysis_ids: list[str]) -> None:
```
kazet: why not delete_scheduled_tasks_for_analyses ?

artemis/cleanup.py
```diff
@@ -67,9 +72,32 @@ def process(self, *args, **kwargs):  # type: ignore
...
+    if not has_unfinished_analyses and Config.Miscellaneous.CLEANUP_RAISE_ERR_ON_NON_UNFINISHED_ANALYSES:
+        raise AssertionError("Did not found unfinished analyses during cleanup.")
+
+    if finished_analyses_ids:
+        # introducing batched to not overwhelm database
+        BATCH = 100
+        for i in range(0, len(finished_analyses_ids), BATCH):
+            analysis_ids = finished_analyses_ids[i : i + BATCH]
+            db.delete_analysis_scheduled_tasks(analysis_ids)
+            logger.info("Cleaned up ScheduledTask table for analyses: %s", ",".join(analysis_ids))
```
kazet: after full cleanup we may log the count, BTW I think  logging precise analyses IDs is more of a debug log

PR #2013, opened 2025-10-07
Decreasing ram usage cleanup

kazet:


artemis/cleanup.py
```diff
@@ -73,18 +72,23 @@ def process(self, *args, **kwargs):  # type: ignore
...
-    karton_state = KartonState(backend=KartonBackend(config=KartonConfig()))
-    finished_analyses_ids = []
+    karton_backend = KartonBackend(config=KartonConfig())
+    finished_analyses_ids_set = set()
     has_unfinished_analyses = False
     for analysis in db.list_analysis():
-        if analysis["id"] not in karton_state.analyses or len(karton_state.analyses[analysis["id"]].pending_tasks) == 0:
-            finished_analyses_ids.append(analysis["id"])
-        else:
+        finished_analyses_ids_set.add(analysis["id"])
```
bulkowy, line 81: there was change that now every analysis is taken into consideration, without checking for it's state. Will it work the same now, as it was before?

kazet: first we take the set of all analyses, and then remove the ones that have tasks (i.e. unfinished). That way we result in the set of finished analyses.

bulkowy (approved):

PR #2046, opened 2025-10-24
report one dast lfi/open-redirect in nuclei

kadewu:
As dast nuclei templates contains multiple templates for those vulnerabilities, we want to make reports cleaner.

### DIFF
--- a/artemis/reporting/modules/nuclei/reporter.py
+++ b/artemis/reporting/modules/nuclei/reporter.py
@@ -31,16 +31,6 @@ SEVERITY_OVERRIDES = {
 }
 
 
-def is_open_redirect_template(template_id: str) -> bool:
-    keys = ("open redirect", "open-redirection")
-    return any(key in template_id.lower() for key in keys)
-
-
-def is_lfi_template(template_id: str) -> bool:
-    keys = ("lfi", "local file inclusion")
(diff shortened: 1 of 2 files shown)


artemis/reporting/modules/nuclei/reporter.py
```diff
@@ -110,6 +122,15 @@ def _is_url_without_query_fragment(url: str) -> bool:
                 + Config.Modules.Nuclei.NUCLEI_TEMPLATES_TO_SKIP_WHEN_REPORTING
             ):
                 continue
+            if is_lfi_template(template):
```
kadewu: I went with that change, but I'm still thinking that maybe it's better to do change to file "nuclei_template_groups.json"
with
{
"dast/vulnerabilities/lfi/lfi-keyed.yaml" : "lfi"
"dast/vulnerabilities/lfi/linux-lfi-fuzz.yaml": "lfi"
"dast/vulnerabilities/lfi/windows-lfi-fuzz.yaml": "lfi"
}
and manually add more lfi if their appears.

We may also then add e.g. http/cves/2015/CVE-2015-4074.yaml - which is also lfi, but not dast and not currently catched by that check in code.

kazet: you're right to use grouping

PR #2061, opened 2025-10-30
Better approach normal formas

kazet:


artemis/api.py
```diff
@@ -237,6 +244,16 @@ async def post_export_delete(id: int) -> Dict[str, Any]:
     }
 
 
+@router.post("/build-html-message", dependencies=[Depends(verify_api_token)])
+async def post_build_html_message(language: str = Body(), data: Dict[str, Any] = Body()) -> str:
```
bulkowy, line 248: where is it used?

artemis/reporting/export/main.py
```diff
@@ -154,15 +135,8 @@ def _build_messages_and_print_path(
...
-        filtered_message = _filter_message(
-            export_data_dict["messages"][top_level_target], exclude_normal_forms_from_html or []
-        )
-
-        if not filtered_message["reports"]:
-            continue
-
         with open(output_messages_directory_name / (top_level_target_shortened + ".html"), "w") as f:
-            f.write(message_template.render({"data": filtered_message}))
+            f.write(message_template.render({"data": export_data_dict["messages"][top_level_target]}))
```
bulkowy, line 139: isn't this essentially rollback of previous MR with filtering messages based on normal forms?

kazet: precisely

artemis/api.py
kazet: in moje

PR #2075, opened 2025-11-06
Add retry mechanism to dangling dns module

kadewu:
To avoid false-positives in case of service being not available, instead of retrying N-times in small time frame, we will schedule retry tasks in extended time frame.
The pr also tends to remove delay on generating reports.

### DIFF
--- a/test/modules/test_dangling_dns_detector.py
+++ b/test/modules/test_dangling_dns_detector.py
@@ -114,7 +114,7 @@ class TestDanglingDnsDetector(ArtemisModuleTestCase):
 
         # when
         result: list[dict[str, Any]] = []
-        self.karton.check_dns_ip_records_are_alive("dangling.example.com", result, True)
+        self.karton.check_dns_ip_records_are_alive("valid.example.com", result, True)
 
         # then
         self.assertFalse(result)
@@ -158,7 +158,7 @@ class TestDanglingDnsDetector(ArtemisModuleTestCase):
 
         # when
         result: list[dict[str, Any]] = []
-        self.karton.check_dns_ip_records_are_alive("dangling.example.com", result, True)
+        self.karton.check_dns_ip_records_are_alive("norecords.example.com", result, True)
 
         # then
         self.assertFalse(result)


artemis/modules/dangling_dns_detector.py
```diff
@@ -205,9 +274,18 @@ def check_dns_ip_records_are_alive(self, domain: str, result: list[dict[str, Any
...
                 pass
+        return ip_records_alive
 
     def run(self, current_task: Task) -> None:
         domain = get_target_host(current_task)
+        task_type = current_task.headers["type"]
+
+        if self.handle_retry_timeout(current_task, task_type):
+            # small delay in case of no active tasks to not spam que
+            time.sleep(5)
```
kadewu, line 285: Probably should be smaller?

artemis/modules/dangling_dns_detector.py
```diff
@@ -220,13 +298,24 @@ def run(self, current_task: Task) -> None:
...
+        ip_records_alive = self.check_dns_ip_records_are_alive(domain, result, save_results=last_ip_scan)
+
+        if task_type == TaskType.DOMAIN_THAT_MAY_NOT_EXIST:
+            # we only check that once, no need to rerun that when suspecting dangling ip
+            self.check_cname(domain, result)
+            self.check_ns(domain, result)
+
+        retry_message = None
+        if self.handle_scheduling_retry(domain, current_task, ip_records_alive, last_ip_scan):
+            retry_message = f"Defined domain has dangling IP record, scheduling {retry_count + 1} retry."
```
kazet: shouldn't we save the message only after the **last** check?

kadewu: Yeah, I wasnt sure if we should save that. The reasoning was, to be able to quickly check in the dashboard how many retries there were. But we can save only after last one, I'm fine with both.

test/modules/test_dangling_dns_detector.py
```diff
@@ -114,7 +114,7 @@ def test_check_dns_ip_records_valid_a(self, mock_resolve, mock_ip_exists) -> Non
 
         # when
         result: list[dict[str, Any]] = []
-        self.karton.check_dns_ip_records_are_alive("valid.example.com", result)
+        self.karton.check_dns_ip_records_are_alive("dangling.example.com", result, True)
```
kazet: hm, should these domain really have been  changed?

kadewu: Yeah, unnecessary change, reverted.

PR #2223, opened 2026-01-09
Reduce number of false positives in dangling-dns-module

kadewu:
PR also reduce the possible number of retries in case of multiple subdomains pointing to the same IP.

### DIFF
--- a/artemis/modules/dangling_dns_detector.py
+++ b/artemis/modules/dangling_dns_detector.py
@@ -302,12 +302,13 @@ class DanglingDnsDetector(ArtemisBase):
 
         return None
 
-    def check_for_public_institutions(self, domain: str) -> bool:
-        # Dangling ip records can generate a high number of FP when dns ip records points to not used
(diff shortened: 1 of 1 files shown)


artemis/modules/dangling_dns_detector.py
```diff
@@ -276,8 +289,59 @@ def check_dns_ip_records_are_alive(self, domain: str, result: list[dict[str, Any
...
+                    if hasattr(record, "rdtype") and record.rdtype == rdatatype.A:
+                        return record.address  # type: ignore[attr-defined, no-any-return]
+
+        except (dns.resolver.NXDOMAIN, dns.resolver.NoNameservers, dns.resolver.Timeout):
+            return None
+
+        return None
+
+    def check_for_public_institutions(self, domain: str) -> bool:
+        # in case of dangling ip records, we run into a number of FP cause of not properly configured
```
kazet: this comment is too laconic, I would expand it (and correct institution -> entity) as it's important to explain what's happening here. BTW, isn't `technology` too broad?

PR #2237, opened 2026-01-13
Detecting and reporting open telnet port

kazet:


artemis/modules/data/ports-artemis-short.txt
```diff
@@ -1,2 +1,2 @@
 # Top 100 ports detected by Artemis
```
bulkowy, line 1: now, Top 101 ports :hehe:

PR #2291, opened 2026-02-05
QOL for dev mode

R3VANEK:
- added pdbpp debugger to each container
- refactored starting scripts

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -34,7 +34,7 @@ for arg in "$@"; do
   esac
 done
 
-echo "Starting Artemis in $MODE mode..."
+echo "Starting Artemis in $MODE mode (single worker=$SINGLE_WORKER_MODE)..."
 
 DOCKER_CONFIGS=$(scripts/find_docker_compose_files --mode='production')
 
@@ -66,7 +66,7 @@ fi
 
 SCALE_ARGS=""
 
-if [ "$SINGLE_WORKER_MODE" != true ]; then
+if [ "$SINGLE_WORKER_MODE" != "true" ]; then
     SCALE_ARGS+="
         --scale=karton-identifier=3
         --scale=karton-port_scanner=5
--- a/scripts/start_dev
+++ b/scripts/start_dev
@@ -2,4 +2,4 @@
 
 # depreciated way to start artemis meant to be a backward compatibility way for open source users to start project
 # delete it after some time
-. scripts/start --mode=development
+. scripts/start --mode=development $@


scripts/create_development_docker_compose.py
```diff
@@ -81,21 +87,45 @@ def process_file(self, strategy: YamlProcessor) -> None:
...
+        required=False,
+        default=["docker-compose.yaml"],
+        help="Input docker-compose YAML files",
+    )
+
+    args = parser.parse_args()
+
+    for input_file in args.input:
+
+        index = input_file.rfind(".")
```
kazet: we can just do os.path.splitext

scripts/create_development_docker_compose.py
```diff
@@ -81,21 +87,45 @@ def process_file(self, strategy: YamlProcessor) -> None:
...
+        processor.process_file(LocalBuildStrategy())
 
-    processor.process_file(LocalBuildStrategy())
+        processor.process_file(WebCommandStrategy())
 
-    processor.process_file(WebCommandStrategy())
+        if input_file == "docker-compose.yaml":
 
-    processor.process_file(VolumeDevelopStrategy())
+            processor.process_file(VolumeDevelopStrategy())
```
kazet, line 123: why we don't add it in submodules?

scripts/find_docker_compose_files
```diff
@@ -0,0 +1,78 @@
...
+BASE_COMPOSE=""
+
+if [ "$DOCKER_COMPOSE_STRING" = "true" ]; then
+    BASE_COMPOSE="-f"
+fi
+
+if [ "$MODE" = "development" ]; then
+    BASE_COMPOSE="$BASE_COMPOSE docker-compose.dev.yaml"
+    YML_BASE_PATH="docker-compose.dev.yml"
+    YML_ADDITIONAL_PATH="docker-compose.additional*.dev.yaml"
```
kazet: I'd ignore additional, the wpscan one in -extra is disabled on purpose :)

scripts/find_docker_compose_files
```diff
@@ -0,0 +1,78 @@
...
+    --docker-compose-string)
+      DOCKER_COMPOSE_STRING=true
+      ;;
+    *)
+      POSITIONAL_ARGS+=("$arg")
+      ;;
+  esac
+done
+
+set -- "${POSITIONAL_ARGS[@]}"
```
kazet: what this does?

scripts/find_docker_compose_files
R3VANEK: ok, deleted all logic connected to yml additional paths

scripts/create_development_docker_compose.py
R3VANEK: Its becouse, submodules have their different context and dockerfiles, and by adding :/opt I override their `COPY` instructions and break path to find the module. This bricks everything.

I could add to `VolumeDevelopStrategy` some ifs, e.g if the service has different context than . then dont do this strategy, but I thought my way is slightly more defensive since this script didn't touched anything besides main docker-compose.yaml previously

scripts/find_docker_compose_files
R3VANEK: it set the positional arguments of $1 and $2 to positional arguments. Forgot to delete it after writing this script, its not necessary

scripts/run_docker_compose
```diff
@@ -2,27 +2,7 @@
...
-fi
-
-if exists */docker-compose.additional*.yml; then
-    for DOCKER_COMPOSE_FILE in */docker-compose.additional*.yml; do
-        FILE_OPTIONS="$FILE_OPTIONS -f $DOCKER_COMPOSE_FILE"
-    done
-fi
-
-docker compose $FILE_OPTIONS $ADDITIONAL_DOCKER_COMPOSE_OPTIONS "$@"
+docker compose $DOCKER_CONFIGS_STRING $ADDITIONAL_DOCKER_COMPOSE_OPTIONS "$@"
```
kadewu: It will not work if we start in 'production' mode If I'm seeing things correctly (cause we basically ovveride compose files for production with development ones)

scripts/start
```diff
@@ -22,43 +63,61 @@ else
...
+SCALE_ARGS=()
+
+if [ "$SINGLE_WORKER_MODE" = true ]; then
+    SCALE_ARGS+=(
+        --scale=karton-identifier=1
+        --scale=karton-port_scanner=1
+        --scale=karton-wordpress_plugins=1
+        --scale=karton-nuclei=1
+        --scale=karton-bruter=1
+        --scale=karton-sqlmap=1
```
kadewu: karton-sqlmap is not by default in Artemis (it's from Artemis-modules-extra)

scripts/start
```diff
@@ -22,43 +63,61 @@ else
...
+        --scale=karton-port_scanner=5
+        --scale=karton-wordpress_plugins=5
+        --scale=karton-nuclei=1
+        --scale=karton-bruter=10
+    )
+
+    if exists Artemis-modules-extra; then
+        SCALE_ARGS+=(--scale=karton-sqlmap=10)
+    else
+        SCALE_ARGS+=(--scale=karton-sqlmap=1)
```
kadewu: same, I believe this should not be included right?

scripts/start
```diff
@@ -14,6 +16,45 @@ else
...
+      POSITIONAL_ARGS+=("$arg")
+      ;;
+  esac
+done
+
+echo "Starting Artemis in $MODE mode..."
+
+DOCKER_CONFIGS=$(scripts/find_docker_compose_files --mode='production')
+
+if [ "$MODE" = "development" ]; then
```
kadewu, line 41: I'm fine with moving this to this file, but maybe small suggestion, that may save some time for our open source users: let's have the 'start_dev' script left in repo for now (and it will start properly this script)

requirements.txt
```diff
@@ -48,11 +48,12 @@ shodan==1.31.0
...
-tqdm==4.67.2
+tqdm==4.67.1
 typer==0.21.1
 typing-extensions==4.15.0
 unittest-parallel==1.7.6
 urllib3==2.6.3
 uvicorn==0.40.0
 validators==0.35.0
 whoisdomain==1.20260106.1
+pdbpp==0.11.7
```
kadewu: Fine be my, but if I'm seeing correctly we don't necessary need to include that library in every module right?
I believe leaving the choice for the develepers if they want to install it may be better.

scripts/run_docker_compose
```diff
@@ -2,27 +2,7 @@
 
 cd $(dirname $0)/..
 
-. scripts/check_docker_compose
+DOCKER_CONFIGS_STRING="$(scripts/find_docker_compose_files --mode=production --docker-compose-string) "
+DOCKER_CONFIGS_STRING+="$(scripts/find_docker_compose_files --mode=development --docker-compose-string)"
```
kadewu: It does seems to work, but is it correct to run it this way?
The flow now seems to always override the 'production' compose files with the 'dev' one.
It may leed to some weird problems if someone will use this script.

requirements.txt
R3VANEK: hm ok if you say so, but I dont see why someone would want plain pdb over pdbpp and these are only 2 choices afaik when dev containers have tty and stdin_open to use for debuggers, not counting some weird IDE integrations

PR #2309, opened 2026-02-12
skip FP in missconfigurations

kadewu:


artemis/modules/dangling_dns_detector.py
```diff
@@ -182,6 +200,16 @@ def _is_cname_dangling(self, record: Rdata, parent_domain: str) -> bool | None:
...
                     break
 
+        if dangling:
+            # check against the main domain
+            main_domain = get_main_domain(cname_target)
+            if main_domain:
+                response = dns_query(main_domain, rdatatype.NS)
+                ns_records = [r.to_text() for r in response] if response else []
+                if not self._is_saas_namespace(ns_records):
+                    # It's more likely a misconfiguration rather than dangling cname record
```
kazet, line 210: are you sure? if it's a nonexistent TLD it might be dangling

PR #2321, opened 2026-02-17
Update quickhits.txt in bruter module, copy from original repository

marcel2012:
I want to update quickhits.txt file with new changes from original repo https://github.com/danielmiessler/SecLists/blob/master/Discovery/Web-Content/quickhits.txt

### DIFF
--- a/artemis/modules/data/bruter/lists/full/quickhits.txt
+++ b/artemis/modules/data/bruter/lists/full/quickhits.txt
@@ -588,8 +588,6 @@
 /accounts.sql
 /accounts.txt
 /activity
-/actuator
-/actuator/heapdump
 /add.php
 /adduser
 /adm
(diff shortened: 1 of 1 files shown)


kazet:
Would you mind prefixing the URLs with /, as the originals were prefixed, so that the diff would be more manageable?

artemis/modules/data/bruter/lists/full/quickhits.txt
```diff
@@ -588,8 +588,6 @@
 /accounts.sql
 /accounts.txt
 /activity
-/actuator
-/actuator/heapdump
```
kazet, line 592: Is there any particular reason you left out these two? I think they are worth keeping

marcel2012: This paths were duplicated. Actuator paths are still present there https://github.com/marcel2012/Artemis/blob/Update-quickhits.txt-in-bruter-module-copy-from-original-repository/artemis/modules/data/bruter/lists/full/quickhits.txt#L2548-L2564

kazet:
Thanks for the contribution!

PR #2323, opened 2026-02-17
Correctly discover tls against domains

kadewu:
Due to possible SNI, TLS can be hostname dependent.
Double check TLS against domains, to avoid false negatives with IP only checking.

### DIFF
--- a/test/images/nginx-with-sni-tls/Dockerfile
+++ b/test/images/nginx-with-sni-tls/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:latest
+FROM nginx:1.29.0
 
 COPY nginx.conf /etc/nginx/nginx.conf
 COPY certs /etc/nginx/certs


artemis/modules/port_scanner.py
```diff
@@ -208,6 +217,20 @@ def _scan(self, target_ips: List[str]) -> Dict[str, Dict[str, Dict[str, Any]]]:
...
+    def tls_handshake(
+        self,
+        ip: str,
+        port: int,
+        domain: str,
+        timeout: float = 3.0,
+    ) -> bool:
+        try:
+            with socket.create_connection((ip, port), timeout=timeout) as sock:
+                with SSL_CONTEXT.wrap_socket(sock, server_hostname=domain):
```
kazet, line 226: test bym poprosił (do modułu jako całości IMO, nie do samego tls_handshake)

artemis/modules/port_scanner.py
```diff
@@ -236,6 +259,10 @@ def run_multiple(self, tasks: List[Task]) -> None:
                 all_results[host] = scan_results.get(host, {})
 
                 for port, result in all_results[host].items():
+                    if task.headers["type"] == TaskType.DOMAIN and int(port) in TLS_PORTS_TO_CHECK:
```
kazet: why dont we make this check for all ports?

kadewu: I was thinking that this additional tls_handshake on all discovered ports may lead to unnecessary overhead, as port_scanner is already quite heavy, therefore adding those port list, but yes, maybe false negatives would be a bigger issue.
Anyway I'm fine with removing that list and checking all ports.

PR #2325, opened 2026-02-18
feat: Add task ID returns to create_tasks and API

mozzar:
- Modified create_tasks() to return List[str] of task IDs
- Updated /add endpoint to return {"ok": true, "ids": [...]} response
- Task UIDs are now accessible to clients after creation


kazet:
Thanks for the contribution!

PR #2327, opened 2026-02-18
Detect wildcard DNS during subdomain enumeration

### LINKED ISSUE #1036: Detect wildcards during reporting

Parvm1102:
Closes #1036

Adds wildcard DNS detection to the subdomain enumeration module. Before dispatching subdomain tasks to downstream scanners, the module now:

Probes 10 random nonexistent subdomains to detect wildcard DNS (threshold: ≥80% resolving)
For each discovered subdomain on a wildcard domain, applies a two-layer filter:

- **DNS layer**: Checks if the subdomain's IPs are a subset of the wildcard IPs
- **HTTP layer**: Compares the subdomain's HTTP response (status + body hash) against a wildcard baseline

Skips subdomains that match on both layers; keeps subdomains with unique IPs or distinct content
Task results are enriched with `wildcard_dns_detected` and `wildcard_subdomains_skipped` fields. Non-wildcard domains are unaffected. Includes 10 unit tests.


Parvm1102:
@kazet Please review this PR.

kadewu (commented):
Hello, thank you for contributing and sorry for late response.
Regarding your changes I currently don't see it being integrated into Artemis reporting - which is what issue is about. Did you look into that?

PR #2329, opened 2026-02-19
Fix event loop starvation in post_build_html_message by replacing blocking requests.post() with async aiohttp call

WHOIM1205:
# Fix: Prevent Event Loop Blocking in `post_build_html_message` Endpoint

## Description

This PR fixes a critical event loop blocking issue in the `post_build_html_message` FastAPI endpoint.

Previously, the endpoint used a synchronous `requests.post()` call inside an `async def` function. Because FastAPI executes `async def` endpoints directly on the event loop thread, this blocking call would freeze the entire Uvicorn event loop whenever the external `autoreporter` service was slow or unavailable.

Since the deployment runs a single Uvicorn worker, this meant **all API endpoints became unresponsive** until the blocking call returned.

This PR replaces the blocking call with a fully asynchronous `aiohttp` implementation and introduces a timeout and proper error handling.

---

### Problem

```python
requests.post(...)

### DIFF
--- a/artemis/api.py
+++ b/artemis/api.py
@@ -1,9 +1,7 @@
 import datetime
-import json
 from typing import Annotated, Any, Dict, List, Optional
 
 import aiohttp
-import requests
 from fastapi import APIRouter, Body, Depends, Header, HTTPException, Query, Request
 from fastapi.responses import RedirectResponse
 from karton.core.backend import KartonBackend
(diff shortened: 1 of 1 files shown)


WHOIM1205:
@kazet is there anything i can change

kazet:
Nope, thank you for the contribution!

PR #2332, opened 2026-02-21
Add Retire.js-based vulnerable JS library detection module

### LINKED ISSUE #2308: Feature Request: Add module for detecting vulnerable JavaScript libraries using Retire.js
Hi @kazet,

After reviewing the repository, documentation (especially the features list at https://artemis-scanner.readthedocs.io/en/latest/features.html), and existing modules, I noticed that Artemis provides excellent coverage for server-side vulnerabilities, configuration issues, CMS checks, exposed resources (e.g., .git via vcs, directory indexing), brute-forcing, Nuclei templates, port scanning, DNS/mail misconfigs, and more. However, there appears to be no dedicated support for scanning client-side JavaScript libraries for known vulnerabilities.

Many modern websites load third-party or outdated JavaScript libraries (e.g., old versions of jQuery, Lodash, Bootstrap, Moment.js) that have public CVEs, often leading to exploitable issues like XSS, prototype pollution, or DoS. These are a common real-world finding (part of OWASP A06:2021 – Vulnerable and Outdated Components) but aren't systematically checked in the current module set (e.g., scripts_unregistered_domains checks domain registration for script sources, but not version/vuln status).

### Proposal
Add a new module (e.g., `js_vuln_detector.py`) that integrates **Retire.js**[](https://github.com/RetireJS/retire.js), a mature, open-source scanner specifically designed to detect vulnerable JavaScript libraries in web pages or Node environments.

**Key benefits this would bring to Artemis:**
- Detects known CVEs in front-end libraries loaded via `<script>` tags (local files, CDNs, etc.).
- Complements existing modules without overlap (Nuclei might catch some via custom templates, but Retire.js is purpose-built for this with a comprehensive, up-to-date vulnerability database).
- Enables more complete vulnerability reporting, e.g.:  
  "Vulnerable JavaScript library detected: jQuery 1.12.4 (CVE-2019-11358 – potential XSS) loaded from /static/jquery.min.js. Recommendation: Update to a patched version."
- Fits perfectly with Artemis's modular, tool-integration philosophy (similar to how `nuclei.py` calls Nuclei via subprocess, or `humble` integrates header analysis).

### High-Level Implementation Ideas
Following the pattern of existing modules (subclass `ModuleBase`, implement `run()`):

1. Trigger on HTTP/URL tasks (or after `webapp_identifier` / crawling modules).
2. Fetch the target page and extract JavaScript resources (using existing HTTP utils or a simple crawler to collect `<script src="...">` URLs and inline scripts if needed).
3. Run Retire.js:
   - Preferred: Use the Python port (`retirejs` on PyPI – `pip install retirejs`) for native integration: `retirejs.scan_endpoint(url)` or scan downloaded files.
   - Alternative: Use the original Node.js CLI via subprocess (install in Docker, similar to other external tools).
4. Parse the output (JSON format) → map findings to Artemis `ReportingMessage` with appropriate severity/risk level.
5. Add dependency handling in `requirements.txt` / Docker setup (Retire.js is Apache-2.0 licensed, compatible with BSD-3-Clause).
6. Tests: Follow the `example.py` template; add unit tests for sample pages with known vulnerable libs.

This seems like a valuable, self-contained addition that could significantly improve the scanner's coverage for client-side risks without major architectural changes.

I'd be happy to work on a prototype PR if this direction looks good. Any thoughts, preferences (e.g., Python port vs. CLI), or adjustments to scope/priority?

Looking forward to your feedback.

nishantxscooby:
## Summary

Implements #2308.

Adds a new Karton module **`js_vuln_detector`** that detects vulnerable client-side JavaScript libraries loaded via `<script src="...">` tags, using a bundled subset of the [Retire.js](https://retirejs.github.io/retire.js/) vulnerability database.

### New module: `artemis/modules/js_vuln_detector.py`
- Fetches HTTP/HTML pages and parses `<script src="...">` tags with BeautifulSoup.
- Matches library names and versions against a bundled vulnerability database (jQuery, Bootstrap, Lodash, AngularJS, Moment.js, Underscore.js).
- Extracts versions from **URL/filename first** (no extra HTTP request), falling back to fetching and scanning script file content (lazy-fetched, capped at 256 KB).
- Reports findings via `TaskStatus.INTERESTING` including library name, detected version, CVE identifiers, script URL, and upgrade guidance.
- Constants: `MAX_SCRIPTS_PER_PAGE = 25`, `MAX_SCRIPT_CONTENT_BYTES = 256 KB`.
- Inline scripts are intentionally excluded (documented in module docstring).

### Bundled DB: `artemis/modules/data/jsrepository.json`
- Retire.js-format vulnerability database covering the 6 most widely-deployed libraries.
- No network access required at runtime.

### Reporting: `artemis/reporting/modules/js_vuln_detector/`
- `reporter.py` — auto-discovered reporter with `VULNERABLE_JS_LIBRARY` ReportType.
- Jinja2 email template with per-finding details.
- English and Polish (`pl_PL`) translation files.

### Tests: `test/modules/test_js_vuln_detector.py`
- Pure unit tests for `_version_is_vulnerable()` and `check_library()` (no Karton/Redis required).
- Integration-style tests using `requests_mock` for the full module flow:
  - Vulnerable jQuery detected → `INTERESTING`
  - Safe jQuery → `OK`
  - Page with no scripts → `OK`
  - Non-HTML response → `OK`
  - Multiple vulnerable libraries on same page

### Fixture: `test/data/vulnerable_js.html`
- HTML page loading jQuery 1.12.0 and Bootstrap 3.3.7 (both vulnerable).

### docker-compose.yaml
- New `karton-js_vuln_detector` service entry.

### docs/features.rst
- Added bullet point describing the new module.

## Dependencies
No new dependencies — `beautifulsoup4` and `packaging` are already in `requirements.txt`.

Closes #2308

### DIFF
--- a/artemis/modules/data/jsrepository.json
+++ b/artemis/modules/data/jsrepository.json
@@ -1,0 +1,337 @@
+{
+  "jquery": {
+    "bowername": [
+      "jquery"
+    ],
+    "extractors": {
+      "filename": [
+        "jquery[._-]([0-9]+(?:\\.[0-9]+)+)"
+      ],
+      "filecontent": [
+        "\\*!? jQuery JavaScript Library v?([0-9][0-9.a-zA-Z_\\-]*)",
+        "\\*!? jQuery v?([0-9][0-9.a-zA-Z_\\-]*)",
+        "jQuery\\s+v([0-9][0-9.a-zA-Z_\\-]*)"
+      ],
+      "uri": [
+        "jquery[._-]([0-9]+(?:\\.[0-9]+)+)"
+      ],
+      "hashes": {}
+    },
+    "vulnerabilities": [
+      {
+        "below": "1.9.0",
+        "severity": "medium",
+        "identifiers": {
+          "CVE": [
+            "CVE-2011-4969"
+          ]
+        },
+        "info": [
+          "https://nvd.nist.gov/vuln/detail/CVE-2011-4969"
+        ]
+      },
+      {
+        "below": "1.12.0",
+        "atOrAbove": "1.11.1",
+        "severity": "medium",
+        "identifiers": {
+          "CVE": [
+            "CVE-2015-9251"
+          ]
+        },
+        "info": [
+          "https://nvd.nist.gov/vuln/detail/CVE-2015-9251"
+        ]
+      },
+      {
+        "below": "3.4.0",
+        "atOrAbove": "1.0.3",
+        "severity": "medium",
+        "identifiers": {
+          "CVE": [
+            "CVE-2019-11358"
+          ]
+        },
+        "info": [
+          "https://nvd.nist.gov/vuln/detail/CVE-2019-11358"
+        ]
+      },
+      {
+        "below": "3.5.0",
+        "severity": "medium",
+        "identifiers": {
+          "CVE": [
+            "CVE-2020-11022",
+            "CVE-2020-11023"
+          ]
+        },
+        "info": [
+          "https://blog.jquery.com/2020/04/10/jquery-3-5-0-released/"
+        ]
+      }
+    ]
+  },
+  "bootstrap": {
+    "bowername": [
+      "bootstrap"
+    ],
+    "extractors": {
+      "filename": [
+        "bootstrap[._-]([0-9]+(?:\\.[0-9]+)+)"
+      ],
+      "filecontent": [
+        "Bootstrap v([0-9][0-9.a-zA-Z_\\-]*)",
+        "\\* Bootstrap v([0-9][0-9.a-zA-Z_\\-]*)"
+      ],
+      "uri": [
+        "bootstrap[._-]([0-9]+(?:\\.[0-9]+)+)"
+      ],
+      "hashes": {}
+    },
+    "vulnerabilities": [
+      {
+        "below": "3.4.0",
+        "atOrAbove": "3.0.0",
+        "severity": "medium",
+        "identifiers": {
+          "CVE": [
+            "CVE-2018-14041",
+            "CVE-2018-14042"
+          ]
+        },
+        "info": [
+          "https://nvd.nist.gov/vuln/detail/CVE-2018-14041"
+        ]
+      },
+      {
+        "below": "4.3.1",
+        "atOrAbove": "4.0.0",
+        "severity": "medium",
+        "identifiers": {
+          "CVE": [
+            "CVE-2019-8331"
+          ]
+        },
+        "info": [
+          "https://nvd.nist.gov/vuln/detail/CVE-2019-8331"
+        ]
+      },
+      {
+        "below": "3.4.1",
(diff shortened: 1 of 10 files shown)


kazet:
Hello,

how frequently are such vulnerabilities exploitable? Do you have a way to automatically check the exploitability?

nishantxscooby:
Hi @kazet, Vulnerable JavaScript libraries are very common and often exploited, especially for XSS or prototype pollution. Retire.js and similar tools flag libraries with known CVE, but can’t always verify if the vulnerability is exploitable without analyzing how the page uses the library. Most industry tools, including the new Artemis module, surface these findings as potential risks so security teams can triage them according to site context and usage.
Thankyou :D

kazet:
> Vulnerable JavaScript libraries are very common and often exploited

I don't want scanned entities to be spammed with false positives - could you share these statistics? If you don't want to share them publicly, you could share them at `artemis@cert.pl`.

nishantxscooby:
Hi @kazet, sorry for the delayed response firstly, I was hospitalised due to health issues. Also, my earlier response was too generic, i'll be answering it in detail below.

Tbh most outdated JS library detections are not directly exploitable without additional context. A site loading jQuery 1.12.0 isn't automatically vulnerable — it depends on whether the application passes user-controlled input to .html(), .append(), or similar DOM manipulation functions. The same applies to Bootstrap XSS (requires user data in tooltip/popover attributes) and AngularJS sandbox escapes (requires user-controlled expressions in ng-app scope).

So,I don't have first-hand scan statistics to share :(

To address this in a way that's actually useful rather than spammy, I've reworked the module significantly:

1. Curated the DB to XSS-only CVEs - Removed Lodash (prototype pollution), Moment.js (ReDoS), Underscore.js (template injection), and jQuery prototype pollution. These vulnerability classes almost never lead to exploitable issues from library presence alone. Only XSS vulnerabilities in jQuery, Bootstrap, and AngularJS remain, as these have the highest real-world exploitation rate.

2. Added a severity threshold (JS_VULN_DETECTOR_MIN_SEVERITY, default: high) - following the same pattern as NUCLEI_SEVERITY_THRESHOLD. By default, only high-severity XSS with well-documented attack vectors is reported.

3. Module is disabled by default - Added to MODULES_DISABLED_BY_DEFAULT so operators explicitly opt in.

4. Documented exploitability per CVE - Each entry in jsrepository.json now has an exploitability_note explaining the specific conditions under which it's exploitable.

I think this is a good starting point - a small, high-confidence set of XSS CVEs, disabled by default, that operators can opt into. Would this approach work for you, or would you prefer the module to also attempt some form of runtime exploitability verification (e.g., checking if the page actually calls vulnerable jQuery DOM methods)?

kazet:
Hello,

unfortunately, the percentage of FPs would be too high to get this one merged.

PR #2335, opened 2026-02-22
Remove -recursive flag from subfinder and add integration test

### LINKED ISSUE #2334: Impact of -recursive on enumeration coverage
While exploring the `SubdomainEnumeration` module, I noticed that subfinder is executed with:
`-all -recursive`

According to subfinder’s documentation, `-recursive` restricts the enumeration to sources that support recursive discovery, which effectively limits the set of providers being used.

I ran a small comparison across multiple domains to understand the impact of this flag.

### Test setup

- Compared `-all -recursive` vs `-all`
- Deduplicated results (`sort -u`)
- Resolved active subdomains using dnsx
- Counted only resolving subdomains

### Results

Image

**airbnb.com**
- `-all -recursive`: 296 unique → 215 resolved
- `-all`: 1702 unique → 349 resolved (~1.6× increase in active subdomains)

**spotify.com**
- `-all -recursive`: 139 unique → 77 resolved
- `-all`: 1346 unique → 224 resolved (~2.9× increase in active subdomains)

So while raw enumeration increases significantly without `-recursive`, there is also a noticeable increase in resolving/active subdomains. But of course, the impact varies depending on the target.

I’m wondering whether the use of `-recursive` is an intentional trade-off to reduce noise/performance impact, or if it would make sense to reconsider this.

imraflip:
This PR closes #2334.

Remove the `-recursive` flag from subfinder invocation and add an integration test verifying that `-all` results include everything returned by `-all -recursive` for several example domains.

### DIFF
--- a/artemis/modules/subdomain_enumeration.py
+++ b/artemis/modules/subdomain_enumeration.py
@@ -142,22 +142,28 @@ class SubdomainEnumeration(ArtemisBase):
 
     def get_subdomains_from_subfinder(self, domain: str) -> Optional[Set[str]]:
         """
-        In practice, the -all and -all -recursive flags can return slightly different outputs depending on timing and provider rate limits. 
+        In practice, the -all and -all -recursive flags can return slightly different outputs depending on timing and provider rate limits.
         Running subfinder twice and merging the results helps reduce the chance of missing subdomains.
         """
         # Run with -all
-        all_result = self.get_subdomains_from_tool(
-            "subfinder",
-            ["-d", domain, "-silent", "-all"],
-            domain,
-        ) or set()
+        all_result = (
+            self.get_subdomains_from_tool(
+                "subfinder",
+                ["-d", domain, "-silent", "-all"],
+                domain,
+            )
+            or set()
+        )
 
         # Run with -all -recursive
-        recursive_result = self.get_subdomains_from_tool(
-            "subfinder",
-            ["-d", domain, "-silent", "-all", "-recursive"],
-            domain,
-        ) or set()
+        recursive_result = (
+            self.get_subdomains_from_tool(
+                "subfinder",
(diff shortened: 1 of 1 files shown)


kazet:
How do you interpret the unit test results?
image

imraflip:
Hello,

After running the tests multiple times locally, I’m seeing inconsistent behavior.

Sometimes `cert.pl` passes, sometimes it fails. I’ve also seen `badssl.com` both fail and pass depending on the run.

This suggests that subfinder’s output isn’t fully deterministic. Things like concurrency or rate limiting of the providers likely affect which results end up in the final set.

Given that, I guess it's fair to say that `-all` is not always a strict superset of `-all -recursive`.

What do you think would be the best direction here?
Should we accept some variability and keep the simpler `-all` approach, or move toward something more deterministic (for example running both and merging the results)?

kazet:
I think running both and merging the results is the safest (although maybe not the fastest) option. Let's go in this direction.

imraflip:
Sounds good :)
I've updated the implementation to run both and merge the results, and removed the superset test.

artemis/modules/subdomain_enumeration.py
```diff
@@ -141,7 +141,22 @@ def get_subdomains_from_tool(
             return None
 
     def get_subdomains_from_subfinder(self, domain: str) -> Optional[Set[str]]:
-        return self.get_subdomains_from_tool("subfinder", ["-d", domain, "-silent", "-all", "-recursive"], domain)
+        # Run with -all
```
kazet, line 148: could you add a small comment describing the rationale for running subfinder twice? This will help future developers not repeat our discussions ;)

imraflip: Yes, sure absolutely

imraflip:
Thanks for the merge and guidance, really appreciate it :)

PR #2336, opened 2026-02-23
fix: replace blocking requests with aiohttp in async proxy endpoints

WHOIM1205:
### Issue

`karton_dashboard()` and `prometheus()` were defined as `async def` but used the synchronous `requests` library.  
This blocks the FastAPI event loop and can freeze the entire application if the upstream service is slow or unavailable.

Additionally, no timeout was configured, which could cause indefinite hangs.

### Fix

- Replaced `requests` with `aiohttp` for non-blocking async HTTP calls.
- Added explicit timeouts:
  - 30s for `/karton-dashboard/*`
  - 10s for `/metrics`
- Updated related type hints and imports.

### Impact

- Prevents event loop starvation.
- Ensures other requests remain responsive even if upstream services are slow or down.
- Improves overall availability and resilience of the frontend service.

### DIFF
--- a/test/unit/test_frontend_async_proxy.py
+++ b/test/unit/test_frontend_async_proxy.py
@@ -11,6 +11,9 @@ FRONTEND_PATH = Path(__file__).resolve().parents[2] / "artemis" / "frontend.py"
 class TestFrontendNoBlockingIO(unittest.TestCase):
     """Static analysis tests ensuring frontend.py doesn't use blocking requests in async functions."""
 
+    source: str
+    tree: ast.Module
+
     @classmethod
     def setUpClass(cls) -> None:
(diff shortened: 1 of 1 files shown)


WHOIM1205:
@kazet 
Replaced blocking `requests` calls inside async endpoints with `aiohttp` to prevent event loop starvation. Also added explicit timeouts to avoid indefinite hangs if upstream services are slow or down. This should eliminate the full-app freeze scenario under Uvicorn.

kazet:
Could you fix the linter issues?

WHOIM1205:
> Could you fix the linter issues?

Thanks! I’ve pushed a commit fixing the linter issues — please take another look when you get a chance.

kazet:
No more comments, thank you for the contribution!

PR #2338, opened 2026-02-23
add deduplication to wordpress plugins reporting module

kadewu:
With many redirects the report may become messy.

### DIFF
--- a/artemis/reporting/modules/wordpress_plugins/reporter.py
+++ b/artemis/reporting/modules/wordpress_plugins/reporter.py
@@ -9,7 +9,7 @@ from artemis.fallback_api_cache import FallbackAPICache
 from artemis.reporting.base.asset import Asset
 from artemis.reporting.base.asset_type import AssetType
 from artemis.reporting.base.language import Language
-from artemis.reporting.base.normal_form import NormalForm, get_url_normal_form
+from artemis.reporting.base.normal_form import NormalForm
 from artemis.reporting.base.report import Report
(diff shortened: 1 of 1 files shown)


artemis/reporting/modules/wordpress_plugins/reporter.py
```diff
@@ -129,18 +129,26 @@ def get_scoring_rules() -> Dict[ReportType, Callable[[Report], List[int]]]:
     @staticmethod
     def get_normal_form_rules() -> Dict[ReportType, Callable[[Report], NormalForm]]:
         """See the docstring in the Reporter class."""
+
+        def actual_target(report: Report) -> str:
+            if "redirect_url" in report.additional_data and report.additional_data["redirect_url"]:
```
kadewu, line 134: There is still one thing I'm not sure if should be addressed.

we will not deduplicate:
- ://url/
- ://url

artemis/reporting/modules/wordpress_plugins/reporter.py
```diff
@@ -129,18 +129,26 @@ def get_scoring_rules() -> Dict[ReportType, Callable[[Report], List[int]]]:
...
+                if not isinstance(report.additional_data["redirect_url"], str):
+                    raise TypeError("redirect_url must be a string")
+                return report.additional_data["redirect_url"]
```
kazet: I would skip the path (directory etc), let's group by domains

PR #2349, opened 2026-02-23
Fix variable shadowing in handle_single_task causing wrong report params

WHOIM1205:
# Fix: Variable Shadowing in `handle_single_task()` Causing Incorrect Report Parameters

## Summary

This PR fixes a variable shadowing bug in `artemis/reporting/task_handler.py` inside the `handle_single_task()` function.

The loop variable inside the `skip_previously_exported` block reused the name `report_generation_task`, which shadows the function parameter of the same name. Because Python loop variables leak into the outer scope, this caused the original task reference to be overwritten.

As a result, the final `export()` call could use parameters from the **last task returned by `db.list_report_generation_tasks()`** instead of the task currently being processed.

---

## Root Cause

```python
def handle_single_task(report_generation_task: ReportGenerationTask) -> Path:
    if report_generation_task.skip_previously_exported:
        for report_generation_task in db.list_report_generation_tasks():

### CODE REVIEW

**Comment by WHOIM1205** [comment]
HI @kazet 
There’s a variable shadowing issue in `handle_single_task()` — the loop reuses `report_generation_task`, which overwrites the original parameter. This can cause `export()` to run with the wrong task data when `skip_previously_exported=True`. Renaming the loop variable fixes it.

File: test/unit/test_task_handler_variable_shadowing.py [hunk 1]
```diff
@@ -0,0 +1,157 @@
...
+        self.language = language
+        self.custom_template_arguments = custom_template_arguments or {}
+        self.skip_previously_exported = skip_previously_exported
+        self.skip_hooks = skip_hooks
+        self.skip_suspicious_reports = skip_suspicious_reports
+        self.include_only_results_since = include_only_results_since
+        self.output_location = output_location
+
+
+class TestHandleSingleTaskNoVariableShadowing(unittest.TestCase):
```
kadewu: Please do not write such names in test and do not write test having 'specifix code changes' in mind when doing so.
After some time passes developer will look at this and would need to go through git history to really understand what was intention of this.

Instead of having test that 'TestThat<NameSpecificFix>' just named it accordingly of what logic is expected.
e.g. Unittest class covering `task_handler.py` would be reasonable. This way in the future tests can be added to other methods from this file.

test/unit/test_task_handler_variable_shadowing.py
```diff
@@ -0,0 +1,157 @@
...
+            tag="org-A",
+            language="pl_PL",
+            custom_template_arguments={"key": "value"},
+            skip_previously_exported=True,
+            skip_hooks=True,
+            skip_suspicious_reports=True,
+            include_only_results_since=None,
+        )
+
+        # A different task that exists in the DB — this is what the old buggy
```
kadewu: Tests are fine, but a similar comment to as above.
Will you be able to tell what mean 'old buggy code' in a year forward?

test/unit/test_task_handler_variable_shadowing.py
```diff
@@ -0,0 +1,157 @@
...
+        call_kwargs = mock_export.call_args[1]
+        self.assertEqual(call_kwargs["tag"], "org-A")
+        self.assertEqual(call_kwargs["language"].value, "pl_PL")
+        self.assertEqual(call_kwargs["custom_template_arguments"], {"key": "value"})
+        self.assertTrue(call_kwargs["skip_hooks"])
+        self.assertTrue(call_kwargs["skip_suspicious_reports"])
+
+    @patch("artemis.reporting.task_handler.export")
+    @patch("artemis.reporting.task_handler.db")
+    def test_export_called_with_original_task_params_multiple_existing_tasks(self, mock_db, mock_export):  # type: ignore
```
kadewu: Actually this test is literally testing the same behavior as the above one.
In both you have an array of 'other existing tasks' that is `len(array)>1`
So no need for that duplication.

But if you are willing to, you can add one test instead when there is no other tasks in db and ensure this also works.

test/unit/test_task_handler_variable_shadowing.py
kadewu: To elaborate on that to clarify: your tests presented here do not test `is there a variable shadowing`, your tests are testing: `task_handler.handle_single_task` is correctly exporting specific task.

WHOIM1205:
@kadewu

WHOIM1205:
hey @kadewu 
Thanks for the feedback — I’ve updated the tests accordingly.

Renamed the file to test_handle_single_task.py
Renamed the test class to TestHandleSingleTask
Removed bug-specific wording and historical references
Cleaned up comments to describe behavior rather than the previous issue
Removed the duplicate test
Added a test case for when the DB returns no existing tasks

The tests now focus purely on validating the expected behavior of handle_single_task() and should be clearer and more maintainable going forward.

kazet:
Could you fix the linter comments?

WHOIM1205:
> Could you fix the linter comments?

done

PR #2350, opened 2026-02-23
Fix release_all_locks not actually releasing locks

WHOIM1205:
### Summary

`ResourceLock.release_all_locks()` was intended to act as a safety net at the start of each worker iteration. However, it only logged held locks without actually releasing them.

Because the heartbeat thread continuously refreshes entries in `LOCKS_TO_SUSTAIN`, any leaked lock would remain sustained in Redis for the lifetime of the worker process.

This could lead to permanent target lock-out after transient failures.

---

### Root Cause

The method:
- Logged lock names
- Did **not** call `REDIS.delete(lock)`
- Did **not** clear `LOCKS_TO_SUSTAIN`

As a result, leaked locks were indefinitely refreshed by the heartbeat thread.

---

### Fix

- Iterate over a safe copy of `LOCKS_TO_SUSTAIN.keys()`
- Call `REDIS.delete(lock)` for each lock
- Clear `LOCKS_TO_SUSTAIN` after deletion

This preserves the existing locking model while ensuring proper cleanup.

---

### Tests Added

- Verifies Redis keys are deleted
- Verifies sustain dict is cleared
- Ensures no-op behavior when empty
- Simulates acquire → forced cleanup scenario
- Thread-safety validation

---

### Impact

- Prevents permanent lock leaks
- Restores expected self-healing behavior
- Eliminates silent scan starvation in long-running workers

### DIFF
--- a/artemis/resource_lock.py
+++ b/artemis/resource_lock.py
@@ -50,8 +50,10 @@ class ResourceLock:
     @staticmethod
     def release_all_locks(logger: Logger) -> None:
         with LOCKS_TO_SUSTAIN_LOCK:
-            for lock in LOCKS_TO_SUSTAIN:
+            for lock in list(LOCKS_TO_SUSTAIN.keys()):
                 logger.info(f"Releasing lock: {lock} -> {LOCKS_TO_SUSTAIN[lock]}")
+                REDIS.delete(lock)
+            LOCKS_TO_SUSTAIN.clear()
 
     def acquire(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


WHOIM1205:
Hi @kazet 
 I noticed that `release_all_locks()` was only logging locks without actually deleting them from Redis or clearing `LOCKS_TO_SUSTAIN`, which could cause leaked locks to be sustained indefinitely by the heartbeat thread. This change ensures the locks are properly deleted and the sustain dict is cleared so the safety net works as intended.

Issue #1977, opened 2025-09-25
Somehow minimize the PoC URLs generated by the Nuclei DAST templates as they're big


kazet:
[...]

PR #2351, opened 2026-02-23
fix issue: #1977 (shorten PoC URLs length)

codedByJana:
## Changes
- Added `NUCLEI_POC_MAX_LENGTH` config (default: 200)
- Truncation helper in `reporter.py` for `curl-command`/`path_query_fragment`
- Set to `0` for full PoCs

## Before/After
**Before**: ```GET /vulnerabilities/sqli.php?dest=http%3A%2F%2F127.0.0.1%2Fabc.html'&redirect=...[continues 800+ chars with 50+ parameters]```
**After**: `https://target...?param...[truncated]` (200 chars max)

## Test
```bash
$ pytest -q test/test_nuclei_poc_truncation.py
.                                                                    [100%]
1 passed in 0.67s
```

### DIFF
--- a/artemis/config.py
+++ b/artemis/config.py
@@ -899,6 +899,11 @@ class Config:
                 "NUCLEI_CHUNK_SIZE is 200, three calls will be made with 200 templates each.",
             ] = get_config("NUCLEI_CHUNK_SIZE", default=200, cast=int)
 
+            NUCLEI_POC_MAX_LENGTH: Annotated[
+                int,
+                "Maximum length (in characters) of PoC URL/command stored in reports. Set to 0 to keep full content.",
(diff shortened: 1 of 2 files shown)


kazet:
Unfortunately, it's not that simple. To shorten a PoC we need not just to trim it at 200 characters - we want a PoC that still works. For example, if we want a sql_injection_detector PoC with 50 parameters, maybe we can trigger an error with just one? That might be the first or any other parameter. Therefore some trial-and-error is needed.

Therefore I  close this one but definitely encourage you to build a generic PoC minimizer for sql_injection_detector and lfi_detector.

codedByJana:
understood, on it

PR #2352, opened 2026-02-24
Pre commit chores

kadewu:
- updated pre-commit
- lint fixes after update
- inspired by the idea of running static code analysis for potential merge conflicts:
  - other static analysis tests don't really test anything useful (after the change such code doesn't exist, and it's not common that someone would really try to add it), therefore removed whole file
  
Additionaly added a note to README/Contributing part to install `pre-commit` before contributing to Artemis

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,13 @@ Contributions are welcome! We will appreciate both ideas for new Artemis modules
 
 However obvious it may seem we kindly remind you that by contributing to Artemis you agree that the BSD 3-Clause License shall apply to your input automatically, without the need for any additional declarations to be made.
 
+### Development
+
+Before opening PR please make sure to install `pre-commit`:
+- `pre-commit install`
+
+And run it against your changes.
+
 ## Contributors
 Huge thanks to the following people that contributed to Artemis development, especially the [KN Cyber](https://kncyber.pl/) science club of [Warsaw University of Technology](https://pw.edu.pl) that initiated the project!
 


artemis/modules/nuclei.py
```diff
@@ -251,7 +251,6 @@ def _get_requests_per_second_statistics(sef, stderr_lines: List[str]) -> str:
         requests_per_second_per_host: List[int] = []
 
         def _finish_current_second() -> None:
-            nonlocal requests_per_second_per_host
```
kazet, line 254: hmm, why was this removed?

artemis/json_utils.py
```diff
@@ -9,7 +9,7 @@
 
 class JSONEncoderAdditionalTypes(json.JSONEncoder):
     def default(self, o: Any) -> Any:
-        if dataclasses.is_dataclass(o):
+        if dataclasses.is_dataclass(o) and not isinstance(o, type):
```
kazet, line 12: I don't understand this one

kadewu: with the first check the expected type of o is `"DataclassInstance | type[DataclassInstance]"`
that's why I added the second check - this way eliminating the possible 'type'

artemis/modules/nuclei.py
kadewu: It's actually not needed, nonlocal would be needed if we assigned anything to varialble.
We only do `requests_per_second_per_host.extend()`, and variable is visible inside this scope.

PR #2355, opened 2026-02-24
Fix zombie process and fd leak in OutputRedirector

WHOIM1205:
### Summary
This PR fixes a resource leak in `OutputRedirector.__exit__` where the spawned `tee` subprocess was killed but never reaped, and duplicated file descriptors were not properly closed.

### What Was Happening
- `self._tee.kill()` was called without `self._tee.wait()`, leaving zombie processes.
- Duplicated stream file descriptors were never closed.
- `tee.stdin` pipe remained open.
- Repeated task retries/timeouts caused FD accumulation and zombie growth.

### Fix
- Close duplicated stream copies after restoring original FDs.
- Close `self._tee.stdin` before termination.
- Call `self._tee.wait()` after `kill()` to properly reap the process.

### Impact
- Prevents zombie process accumulation.
- Eliminates file descriptor leaks.
- Improves stability of long-running Artemis workers under high task volume and retries.


WHOIM1205:
hey @kadewu @kazet 
The `tee` process was killed but never waited on, and duplicated FDs weren’t closed, causing zombie processes and file descriptor leaks over time.

This fix properly closes the FDs and calls `wait()` to ensure clean resource cleanup.

PR #2360, opened 2026-02-25
Add more logs in scheduled tasks cleanup job

kadewu:


artemis/cleanup.py
```diff
@@ -80,12 +80,14 @@ def _cleanup_scheduled_tasks() -> None:
     for analysis in db.list_analysis():
         finished_analyses_ids_set.add(analysis["id"])
 
-    kept_rows = 0
+    logger.info("Found %d analyses", len(finished_analyses_ids_set))
```
kazet, line 83: I'd add `finished` analyses

kadewu: Actually that line is logging all analyses at the start of cleanup.
But added the 'finished' keyword below.

artemis/cleanup.py
```diff
@@ -101,7 +106,13 @@ def _cleanup_scheduled_tasks() -> None:
             removed_rows += rows
 
             logger.debug("Cleaned up ScheduledTask table for analyses: %s", ",".join(analysis_ids))
-        logger.info("Removed %d rows in ScheduleTask table, kept %d.", removed_rows, kept_rows)
+        logger.info(
+            "Removed %d rows in ScheduleTask table for %d finished analyses. "
+            "Number of remaining unfinished analyses: %d.",
```
kazet, line 111: unfinished or finished?

kadewu: The second row is indeed about unfinished (method always should clear all scheduled tasks for finished ones)

PR #2366, opened 2026-02-26
Fix premature raise skipping error saving for batch tasks

WHOIM1205:
### Problem
In `process_multiple()` (`artemis/module_base.py:716`), the `raise` statement was placed **inside** the `for task in task_group` loop.

When a batch failed after exhausting retries:
- Only the **first task** in the batch had its error details saved.
- Remaining tasks were marked as `CRASHED` by the outer handler,
- But without traceback information — causing silent debugging data loss.

This affects modules using `batch_tasks = True` (e.g., `nuclei`, `port_scanner`) with batch sizes > 1.

---

### Fix
Moved the `raise` statement **outside** the loop so that:
1. All tasks in the batch persist their error details.
2. The exception is re-raised only after error persistence completes.

---

### Result
- All failed batch tasks now correctly store traceback information.
- No silent loss of debugging data.
- Exception propagation behavior remains unchanged.

### DIFF
--- a/artemis/module_base.py
+++ b/artemis/module_base.py
@@ -713,7 +713,7 @@ class ArtemisBase(Karton):
                                 self.db.save_task_result(
                                     task=task, status=TaskStatus.ERROR, data=traceback.format_exc()
                                 )
-                                raise
+                            raise
             finally:
                 for task in task_group:
                     if Config.Data.SAVE_LOGS_IN_DATABASE:


WHOIM1205:
Heyy @kazet 
The raise was inside the for task in task_group loop which caused only the first task in a failed batch to persist its error details.  
I’ve moved the `raise` outside the loop so all tasks now save their traceback before the exception propagates. This prevents silent loss of debugging information for batch modules like nuclei and port_scanner.

PR #2368, opened 2026-02-27
Refactor templates to consistently use request.url_for instead of hardcoded paths

### LINKED ISSUE #2361: Replace hardcoded URL paths in templates with request.url_for
I noticed that many templates use root-relative hardcoded paths (e.g. `/export/delete/{id}`, `/add`, `/results`, etc.) instead of using `request.url_for()`. This might cause issues if router prefixes change or if the app is mounted under a subpath. Would it make sense to replace them with dynamic URL generation? I can prepare a PR if that aligns with the project direction.

imraflip:
Closes #2361 

This PR replaces hardcoded root-relative paths across templates and related modules with `request.url_for` to ensure routing consistency and better alignment with FastAPI’s routing model

Using `request.url_for` avoids potential issues if router prefixes or paths change in the future and ensures all navigation links are generated dynamically, and the change has been applied consistently across all affected templates

### DIFF
--- a/artemis/api.py
+++ b/artemis/api.py
@@ -196,7 +196,7 @@ def archive_tag(tag: str) -> Dict[str, bool]:
 
 
 @router.get("/exports", dependencies=[Depends(verify_api_token)])
-def get_exports(request: Request, tag_prefix: Optional[str] = None) -> List[ReportGenerationTaskModel]:
+def get_exports(tag_prefix: Optional[str] = None) -> List[ReportGenerationTaskModel]:
     """List all exports. An export is a request to create human-readable messages that may be sent to scanned entities."""
     return [
         ReportGenerationTaskModel(
@@ -208,7 +208,7 @@ def get_exports(request: Request, tag_prefix: Optional[str] = None) -> List[Repo
             language=task.language,
             include_only_results_since=task.include_only_results_since,
             skip_previously_exported=task.skip_previously_exported,
-            zip_url=str(request.url_for("download_zip", id=task.id)) if task.output_location else None,
+            zip_url=f"/api/export/download-zip/{task.id}" if task.output_location else None,
             error=task.error,
             alerts=task.alerts,
         )
@@ -224,9 +224,9 @@ def is_blocklisted(domain: str) -> bool:
 
 # This is a redirect so that we have an entry in api docs
 @router.get("/export/download-zip/{id}", dependencies=[Depends(verify_api_token)])
-def download_zip(request: Request, id: int) -> RedirectResponse:
+def download_zip(id: int) -> RedirectResponse:
     """Download a zip file containing an export - all messages that can be sent to scanned entities + additional data such as statistics."""
-    return RedirectResponse(str(request.url_for("export_download_zip", id=id)))
+    return RedirectResponse(f"/export/download-zip/{id}")
 
 
 @router.post("/export/delete/{id}", dependencies=[Depends(verify_api_token)])
--- a/templates/exports.jinja2
+++ b/templates/exports.jinja2
@@ -87,8 +87,8 @@
                         </td>
                         <td>
                             {%  if report_generation_task.status == "done" %}
-                                <a href="{{ request.url_for('export_download_zip', id=report_generation_task.id) }}">Download ZIP</a> |
-                                <a href="{{ request.url_for('view_export', id=report_generation_task.id) }}">Browse</a> |
+                                <a href="{{ request.url_for('export_download_zip', id=report_generation_task.id).path }}">Download ZIP</a> |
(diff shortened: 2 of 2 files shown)


kadewu:
Hello @imraflip As we do not have any gui tests in artemis please do a double check with all your changes.
I did not debug anything, just the fast glance; after pulling your changes the table of analysed targets at root is not showing for me anymore.

imraflip:
Hey, I had accidentally removed the block scripts and closing tag in the base template during refactor, which prevented the child template scripts from rendering

I’ve pushed a fix. Please let me know if everything looks fine now

templates/components/navbar.jinja2
```diff
@@ -1,38 +1,38 @@
...
-                        <a class="nav-link {{ 'active' if request.path == '/queue' }}" href="/queue">Task queue</a>
+                        <a class="nav-link {{ 'active' if request.url.path == request.url_for('get_queue').path }}" href="{{ request.url_for('get_queue') }}">Task queue</a>
                     </li>
                     <li class="nav-item">
-                        <a class="nav-link {{ 'active' if request.path == '/restart-crashed-tasks' }}" href="/restart-crashed-tasks">Restart crashed tasks</a>
+                        <a class="nav-link {{ 'active' if request.url.path == request.url_for('get_restart_crashed_tasks').path }}" href="{{ request.url_for('get_restart_crashed_tasks') }}">Restart crashed tasks</a>
                     </li>
                     <li class="nav-item">
-                        <a class="nav-link {{ 'active' if request.path == '/docs' }}" href="/docs">API</a>
+                        <a class="nav-link {{ 'active' if request.url.path == request.url_for('api_docs_information').path }}" href="{{ request.url_for('api_docs_information') }}">API</a>
```
kadewu: We cannot do this one, cause of the following code:
```
if not Config.Miscellaneous.API_TOKEN:

    @router.get("/docs", include_in_schema=False)
    def api_docs_information(request: Request) -> Response:
```

Let's revert this one, and I believe we would be good to go.

imraflip:
Hello, I already fixed that. I'm sorry I didn't double check it before

kadewu:
The e2e tests are failing and I'm suspecting cause of the changes.
Can you verify that?

imraflip:
Yes, you were right, the e2e failures were caused by my changes

After running the tests on my own, the failures were all coming from `test_exporting.py`. Looking more closely,the issue wasn’t really with `request.url_for()` itself, but with how the tests assert specific URL formats. In `exports.jinja`, I just added `.path` there so the rendered links stay relative, because the GUI test matches them with a strict regex

In some endpoints in `api.py`, replacing the hardcoded redirect with `request.url_for()` changed the redirect target to an absolute URL, which affected the API test flow. Since those endpoints are basically simple redirects, I reverted them back to relative paths to preserve the original behavior

And also I re ran the e2e tests and all 12 tests pass on my end now ;)

kadewu:
Thank you!
I've triggered test run, and we will be able to merge it once they pass.

imraflip:
Youre welcome!
Finally the tests passed 😅 

By the way, I noticed that some confirmation actions like removing finished analysis and deleting export redirect to a separate page instead of using a modal popup

Do you think it would be better if we handled them with modals instead or is there a specific reason theyre handled as separate pages?

kazet:

Regarding handling removing finished analyses/deleting export as separate pages - you can potentially make them modals but I don't think it will be the best use of your time, improving vulnerability detection capabilities of Artemis would be :)

imraflip:
You're welcome and thanks for the merge!
And no worries, I actually have some free time, so I think (I hope) I can handle both. I’ll keep working on detection improvements as well

For the modal change, should I open an issue first, or just submit a small PR?

imraflip:
Also, I had an idea regarding the `admin_panel_login_bruter` module. I saw someone is already working on improvements there

Their proposal focuses more on form parsing and JSON handling and I was thinking about improving the success detection logic by handling HTTP redirects, since sometimes a successful login results in a redirect to another page

Would it be okay if I explored that as an additional improvement? I just want to make sure not to overlap with their work

kazet:
Regarding modal, you can submit a PR, but let me reiterate that it's not the top priority ;)

Regarding admin_panel_login_bruter - you need to coordinate that with Sumit

imraflip:
Hello, I’ve submitted the PR for the modal changes.

Regarding the `admin_panel_login_bruter` module, I’ll wait for Sumit to open their PR first or work on something else in the meantime, I think it's better not to interfere with their work

kazet:
> Their proposal focuses more on form parsing and JSON handling and I was thinking about improving the success detection logic by handling HTTP redirects, since sometimes a successful login results in a redirect to another page

that makes sense!

imraflip:
Is that a greenlight for me to start? 😅

kazet:
yes

PR #2369, opened 2026-02-27
Add leak_scanner module to detect PDFs with bad redaction

### LINKED ISSUE #327: Detect PDFs with bad censorship attempts (e.g. black squares that can be removed to obtain the "hidden" information)

Sahityaaryan:
Closes #327

 Adds a leak_scanner module that crawls websites for PDFs and uses the [x-ray](https://github.com/nicholasgasior/x-ray) library to detect text hidden beneath black rectangles or other overlays (bad censorship attempts).

### What's new

- `artemis/modules/leak_scanner.py` — the module itself. Receives HTTP service tasks, discovers PDF links via existing crawling utility, downloads and inspects each one.

- `artemis/reporting/modules/leak_scanner/` — reporter, email template, and translation files for the leaked_sensitive_data report type (severity:HIGH).

- `artemis/config.py` — two config options: max PDFs to check (default 10) and max PDF size (default 5MB).

- `requirements.txt` — added x-ray==0.3.6. No Dockerfile changes needed (PyMuPDF ships pre-built wheels).

- `docker-compose.yaml / docker-compose.test.yaml` — added the karton service and a test nginx container.

-` test/` — module test with a real PDF containing hidden text, plus a reporter integration test.

Addressed feedback from the previous attempt: module is named `LeakScanner` for broader scope, crawls for PDFs instead of expecting PDF URLs directly, uses `leaked_sensitive_data report type`, and does not override `get_normal_form_rules()`.

### DIFF
--- a/artemis/modules/leak_scanner.py
+++ b/artemis/modules/leak_scanner.py
@@ -5,7 +5,7 @@ from collections import deque
 from typing import Any, Callable, Dict, List, Set, Tuple
 
 import requests
-import xray
+import xray  # type: ignore
 from bs4 import BeautifulSoup
 from karton.core import Task
 
@@ -84,7 +84,7 @@ class LeakScanner(ArtemisBase):
 
         visited: Set[str] = set()
         document_urls: List[str] = []
-        queue: deque = deque([(base_url, 0)])
+        queue: deque[Tuple[str, int]] = deque([(base_url, 0)])
         visited.add(base_url)
         pages_crawled = 0
 
@@ -205,9 +205,7 @@ class LeakScanner(ArtemisBase):
         if documents_with_findings:
             status = TaskStatus.INTERESTING
             num_total_findings = sum(
-                len(items)
-                for doc in documents_with_findings
-                for items in doc["findings"].values()
+                len(items) for doc in documents_with_findings for items in doc["findings"].values()
             )
             status_reason = (
                 f"Found {num_total_findings} leaked sensitive data item(s) "


artemis/modules/leak_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    identity = "leak_scanner"
+    filters = [
+        {"type": TaskType.SERVICE.value, "service": Service.HTTP.value},
+    ]
+
+    def _discover_pdf_urls(self, base_url: str) -> List[str]:
+        """Crawl the website and return URLs that point to PDF files."""
+        all_links = get_links_and_resources_on_same_domain(base_url)
```
kazet: this is too shallow crawling to actually detet something (the depth is 1). What is the crawling depth to actually detect something interesting?

artemis/modules/leak_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+        pdf_urls = []
+        for link in all_links:
+            parsed = urllib.parse.urlparse(link)
+            if parsed.path.lower().endswith(".pdf"):
```
kazet: I would factor the list of extensions to a variable, there will be more to come, not only PDF

artemis/modules/leak_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+                pdf_urls.append(link)
+        return pdf_urls[: Config.Modules.LeakScanner.LEAK_SCANNER_MAX_PDFS_TO_CHECK]
+
+    def _scan_pdf(self, pdf_url: str) -> Dict[str, Any]:
```
kazet: I think we need to be able to run multiple checks, not only x-ray, against a file

artemis/config.py
```diff
@@ -391,6 +391,17 @@ class Humble:
                 cast=decouple.Csv(str, delimiter=","),
             )
 
+        class LeakScanner:
+            LEAK_SCANNER_MAX_PDFS_TO_CHECK: Annotated[
+                int,
+                "Maximum number of PDFs to download and scan per website.",
+            ] = get_config("LEAK_SCANNER_MAX_PDFS_TO_CHECK", default=10, cast=int)
```
kazet: We would need to be lucky to find something across 10 PDFs on a website, I think the scanner needs to go deeper.

docker-compose.yaml
```diff
@@ -1,15 +1,6 @@
...
   image: certpl/artemis:latest
-  # Prevent NDOTS-based DNS search domain expansion for short hostnames such as
-  # "s3mock", "redis", "postgres".  Without this, a host whose /etc/resolv.conf
-  # contains e.g. "search corp.example.com" causes every DNS lookup for "s3mock"
-  # to first try "s3mock.corp.example.com" against the *external* nameserver,
-  # generating hundreds-of-thousands of spurious DNS queries per day (see #2310).
-  # ndots:0 means every name is treated as absolute, so search domains are never
-  # appended; Docker's embedded DNS (127.0.0.11) still resolves service names.
-  dns_opt:
-    - ndots:0
```
kazet, line 12: you accidentally reverted somebody else's change

kazet (changes requested):

Sahityaaryan:
hi @kazet, thanks for the review! I've addressed all 5 comments, but before pushing the changes, I want to ask  you about few things:

1. Crawling defaults — I've implemented a BFS recursive crawler with configurable depth. I've set the defaults to CRAWL_DEPTH=3, MAX_PAGES_TO_CRAWL=100, and MAX_DOCUMENTS_TO_CHECK=100. Do these defaults seem reasonable, or would you prefer different values?
  
2. External URL sources — You mentioned using sources like archive.org/Common Crawl via https://github.com/lc/gau for discovering documents in the earlier PR review. For now I've only implemented BFS crawling on the site itself. Should I also integrate gau for document URL discovery, or is BFS sufficient for the initial version?
  
3. Check registry - I'm planning to create a DOCUMENT_CHECKS registry where each check declares its name, applicable file extensions, and a check function. Currently it would only have the x-ray bad redaction check. Do you have any specific additional checks in mind that I should add, or is the extensible pattern enough for this PR?

4. Naming — I've renamed everything from PDF-specific to document-generic naming (DOCUMENT_EXTENSIONS, documents_with_findings, _scan_document, etc.) to support future file types. Does this align with what you had in mind?

docker-compose.yaml
Sahityaaryan: I am really sorry for this, I have fixed it now

kazet:
> 1. Crawling defaults — I've implemented a BFS recursive crawler with configurable depth. I've set the defaults to CRAWL_DEPTH=3, MAX_PAGES_TO_CRAWL=100, and MAX_DOCUMENTS_TO_CHECK=100. Do these defaults seem reasonable, or would you prefer different values?

They seem reasonable for the start, I would maybe change max_pages_to_crawl to 1000 - we will see how they perform in practice ;)

> 2. External URL sources — You mentioned using sources like archive.org/Common Crawl via https://github.com/lc/gau for discovering documents in the earlier PR review. For now I've only implemented BFS crawling on the site itself. Should I also integrate gau for document URL discovery, or is BFS sufficient for the initial version?

Let's integrate both, I don't think it'd be much work.

> 3. Check registry - I'm planning to create a DOCUMENT_CHECKS registry where each check declares its name, applicable file extensions, and a check function. Currently it would only have the x-ray bad redaction check. Do you have any specific additional checks in mind that I should add, or is the extensible pattern enough for this PR?
 extensible pattern is enough for this PR
 
> 4. Naming — I've renamed everything from PDF-specific to document-generic naming (DOCUMENT_EXTENSIONS, documents_with_findings, _scan_document, etc.) to support future file types. Does this align with what you had in mind?

yup, it's precisely what I meant :)

Sahityaaryan:
Hi @kazet, thanks for the feedback on the clarification questions! I've pushed the updated changes
  addressing everything:

  Changes in this update:

  1. BFS recursive crawler — Replaced the shallow single-page crawl with a BFS crawler that follows
  same-domain links up to a configurable depth. Defaults: CRAWL_DEPTH=3, MAX_PAGES_TO_CRAWL=1000,
  MAX_DOCUMENTS_TO_CHECK=100.
  2. gau integration — Added _discover_documents_via_gau() that queries Wayback Machine and Common Crawl for
   archived document URLs. Results are merged with BFS-discovered URLs, deduplicated, and capped at
  MAX_DOCUMENTS_TO_CHECK. If gau fails, the module falls back to BFS results only.
  3. Extensible document support — Factored file extensions into a DOCUMENT_EXTENSIONS variable and created
  a DOCUMENT_CHECKS registry where each check declares its name, applicable extensions, and check function.
  Adding a new check is a single tuple append.
  4. Renamed to document-generic naming — documents_with_findings, _scan_document, DOCUMENT_EXTENSIONS, etc.
  5. Rebased on latest main — Includes the dns_opt: ndots:0 fix.

Sahityaaryan:
@kazet, If you got some time then please review this :)

artemis/modules/leak_scanner.py
```diff
@@ -0,0 +1,235 @@
...
+        url = get_target_url(current_task)
+        self.log.info(f"LeakScanner scanning {url}")
+
+        document_urls = self._discover_document_urls(url)
+        self.log.info(f"Found {len(document_urls)} document URL(s) on {url}")
+
+        documents_with_findings: List[Dict[str, Any]] = []
+        for document_url in document_urls:
+            self.log.info(f"Scanning document: {document_url}")
+            result = self._scan_document(document_url)
```
kadewu, line 201: One question, how expensive is currently performing such scanning?
Does it make sense to implement caching?

I could imagine we may scan same document multiple times per host?

Sahityaaryan: within a sigle run, the seen set in _discover_document_urls already handles this - the same URL won't be downloaded twice even if the BFS crawler and gau both surface it.

For cross-run caching (i.e the same PDF being re-checked on future rescan of the same host), the module doesn't do anything special - but from what I can see, other modules in Artemis don't cache across runs either. Happy to add it if you think it's worth it here, but I didn't want to add complexity that isn't established pattern elsewhere.

on the expense question - it's heavier than most modules.
**Worst Case**: 100 pdfs at up to 5 mb each, plus PyMuPDF parsing each one. The LEAK_SCANNER_MAX_DOCUMENT_SIZE_BYTES and  LEAK_SCANNER_MAX_DOCUMENTS_TO_CHECK configs let operators tune it dow if needed.

kadewu: I wouldn't worry about memory usage, it doesn't seem that you try to load everything into a memory at once.
I should have been more specific on my question, my question is how much time each scan takes (per one document). Were you able to check that? (I know it's document based, but any benchmark would be sufficient).

Sahityaaryan: Ran a benchmark on real government PDFs — the kind Artemis would actually encounter:

I have run each of them 5 times to get the avg scan time

# Documents

NIST FIPS 140-3 --> 303 KB --> (avg scan time) 111.3 ms --> (min) 92 ms --> (max) 140 ms
NIST SP 800-61r2 --> 1704 KB --> (avg scan time) 961.9 ms --> (min) 840 ms --> (max) 1028 ms
CISA Ransomeware Guide --> 2484 KB --> (avg scan time) 812 ms --> (min) 798 ms --> (max) 846 ms


Roughly 100ms–1s per document depending on size. Interestingly the 2.5MB PDF was faster than the 1.7MB one — scan time depends on page complexity, not just file size.

At the default cap of 100 documents, worst case is ~1-2 minutes per host, but real government sites typically have far fewer PDFs so in practice it's closer to 10-20 seconds.

PDFs used: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.140-3.pdf, https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-61r2.pdf, https://www.cisa.gov/sites/default/files/publications/CISA_MS-ISAC_Ransomware%20Guide_S508C.pdf

Sahityaaryan: Note: This measures x-ray.inspect() only. Download time depneds on network conditions and is separate.

Sahityaaryan: Like if we are talking about "how much time each scan takes per one document" -- it's just `_scan_document()` call, not the full crawl.

and it's essentially:

1. download the pdf
2. xray.inspect(data) -- scan the doc

kadewu: Yes I'm aware it's measurement of crawling.
And that answer initial question, there is no point in adding caching.

test/modules/test_leak_scanner.py
```diff
@@ -0,0 +1,42 @@
...
+                "ssl": False,
+            },
+        )
+        self.run_task(task)
+        (call,) = self.mock_db.save_task_result.call_args_list
+        self.assertEqual(call.kwargs["status"], TaskStatus.INTERESTING)
+        self.assertIn("leaked sensitive data", call.kwargs["status_reason"].lower())
+
+        data = call.kwargs["data"]
+        self.assertGreater(data["documents_checked"], 0)
```
kadewu: Let's be specific here, the tests should describe the tested behavior as precisely as possible.
(and it applies to whole test)

test/data/leak_scanner/index.html
```diff
@@ -0,0 +1,6 @@
+<html>
+<body>
+    <h1>Test Page</h1>
```
kadewu, line 3: Can we also have an additional pdf (correct one)
This way we should cover most cases with this setup.

artemis/modules/leak_scanner.py
```diff
@@ -0,0 +1,235 @@
...
+        visited: Set[str] = set()
+        document_urls: List[str] = []
+        queue: deque = deque([(base_url, 0)])
+        visited.add(base_url)
+        pages_crawled = 0
+
+        while queue and pages_crawled < max_pages and len(document_urls) < max_documents:
+            current_url, depth = queue.popleft()
+
+            if _is_document_url(current_url):
```
kadewu: Can that actually happen?
If it's only for base_url, lets move that from the loop, for better readability.

artemis/modules/leak_scanner.py
```diff
@@ -0,0 +1,235 @@
...
+                        continue
+
+                    new_url = urllib.parse.urljoin(current_url, tag[attribute])
+                    new_url = new_url.split("#")[0]
+                    new_url_parsed = urllib.parse.urlparse(new_url)
+
+                    if new_url_parsed.hostname != base_hostname:
+                        continue
+
+                    if new_url in visited:
```
kadewu: We may as well merge those two ifs

artemis/modules/leak_scanner.py
```diff
@@ -0,0 +1,235 @@
...
+        """Query gau for archived document URLs from Wayback Machine and Common Crawl."""
+        try:
+            result = check_output_log_on_error(["gau"], self.log, input=domain.encode("idna"))
+            urls: Set[str] = set()
+            for line in result.decode().splitlines():
+                line = line.strip()
+                if line and _is_document_url(line):
+                    urls.add(line)
+            return urls
+        except Exception:
```
kadewu: Can we narrow that exception?
and also log it, so we do not drop silentely.

test/modules/test_leak_scanner.py
kadewu: So I mean by that, assertEqual is always better if possible.

artemis/modules/leak_scanner.py
Sahityaaryan: it can only happen for base_url since document URLs are never added to the queue (only non-document URLs reach queue.append). Moved it before the loop.

artemis/modules/leak_scanner.py
Sahityaaryan: done

artemis/modules/leak_scanner.py
Sahityaaryan: done

test/data/leak_scanner/index.html
Sahityaaryan: done

test/modules/test_leak_scanner.py
Sahityaaryan: done

kadewu:
Sorry for the delay.
Please make sure to have pre-commit installed locally. Some things need to be adjusted there manually (no need to worry about translations, I will update them if needed)

I've updated the branch in the meantime as the test would have failed otherwise.

Sahityaaryan:
Done! Installed pre-commit locally and fixed the issues it flagged (black formatting + mypy errors for xray import and deque type parameter). Pushed the fixes. :)

Sahityaaryan:
The license check is failing because PyMuPDF (dependency of x-ray) is AGPL-3.0, which isn't in the authorized licenses. Should I add it to authorized_packages, or would you prefer to handle it?

kadewu:
We will need to move this PR to https://github.com/CERT-Polska/Artemis-modules-extra
Even if x-ray claim to be BSD, it's not gonna be BSD-compatible as what we claim.

Will you be able to move those files up there? Feel free to ask any question.

Sahityaaryan:
Ahhh......,

  Sure! I'll move it to Artemis-modules-extra. A few questions to make sure I get the structure right:

  1. Should I follow the same pattern as karton_ssl_checks (module file + Dockerfile + requirements.txt + test)?
  2. Should the reporter and translation files stay in Artemis-modules-extra as well, or is reporting handled differently there?
  3. Should I close this PR once the Artemis-modules-extra PR is opened?

kadewu:
1. Yes, exactly the same pattern and please remember to update compose-file as well.
2. autoreporter_addons/<module_name> the reporting should go there
3. Yes this PR will be closed. You can even link the new PR while closing.

You can verify if everthing works with cloning `Artemis-modules-extra` inside our main repository. Either by using provided scripts in this repository or by using docker compose with `-f` flags to provide both compose files.

Sahityaaryan:
@kadewu 

I have created the Pull request in the Artemis-modules-extra: https://github.com/CERT-Polska/Artemis-modules-extra/pull/264

kadewu:
Thank you.
Closing this PR then, and we will move to the new one.

PR #2371, opened 2026-02-28
Fix lock heartbeat thread dying on transient Redis errors

WHOIM1205:
### Problem
`sustain_locks()` runs in a daemon thread and had no exception handling.  
If `REDIS.set()` raised (e.g., ConnectionError, TimeoutError), the thread
terminated silently. This caused:

- Lock heartbeats to stop
- Locks to expire after `LOCK_HEARTBEAT_TIMEOUT`
- Other workers to acquire the same locks
- Concurrent execution on the same resource
- Potential data corruption and race conditions

### Root Cause
Uncaught exceptions inside a critical background infrastructure thread.

### Solution
Wrap the heartbeat loop body in `try/except` and log failures.
This ensures transient Redis errors do not permanently kill the thread
and heartbeats resume automatically on recovery.

### Impact
- Prevents silent lock expiry during Redis outages
- Eliminates race conditions caused by expired locks
- Improves resilience of distributed locking mechanism
- Makes lock heartbeat fault-tolerant to transient network failures

### DIFF
--- a/artemis/resource_lock.py
+++ b/artemis/resource_lock.py
@@ -1,3 +1,4 @@
+import logging
 import sys
 import threading
 import time
@@ -9,6 +10,8 @@ from uuid import uuid4
 from redis import Redis
 
 from artemis.config import Config
+
+logger = logging.getLogger(__name__)
 
 
 class FailedToAcquireLockException(Exception):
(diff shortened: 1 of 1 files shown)


WHOIM1205:
@kazet , @kadewu 
This fixes a critical failure mode in `sustain_locks()` where any Redis exception (e.g., connection blip) would silently terminate the daemon heartbeat thread once the thread died active locks stopped receiving heartbeats and expired allowing other workers to acquire the same locks and causing race conditions
The loop is now wrapped in try/except so transient Redis errors are logged and retried instead of killing the thread this keeps the distributed locking mechanism resilient to temporary Redis/network failures

PR #2374, opened 2026-02-28
fix: add .gitattributes to prevent CRLF issues on Windows

StefanCostache10:


.gitattributes
```diff
@@ -0,0 +1 @@
+*.sh text eol=lf
```
kazet: Is .sh sufficient? Are there no other file types (e.g. .patch) you needed to fix in order to successfully run Artemis on Windows?

StefanCostache10: On my machine this was the only required command in order to fix it. I'm thinking that it should be a global fix though as I did not install/do anything special beforehand, I just cloned the repo and tried to build with docker. If you think that the issue needs more than just this fix, I would happily look into any edge-cases.

kazet: If you ran the tests on your Windows machine and they succeeded, I'm ok with the fix ;)

Just remove this troubleshooting item: https://artemis-scanner.readthedocs.io/en/latest/user-guide/troubleshooting.html#windows-build-issues as it's not needed anymore

kazet:
Hello,

did you ran tests on a Windows machine? did they pass?

StefanCostache10:
Hello. Sorry I did not get around to testing it further. On my machine I set git autocrlf to true when I was first trying to see if it works, and it built successfully then, but I want to try it again tomorrow on a new machine to check if it works. I will start on the main branch and try to build the project, then I will try the fixed branch I have on a fork to see if it works, and then I will report back with the results.

StefanCostache10:
Greetings!
So I ran the mentioned above test and also removed every piece of cached data on the machine by git. On the main branch it crashed with the no dependencies error, on the fixed branch it built successfully

kazet:
Cool! Did you also run the tests via ./scripts/test?

StefanCostache10:
So I just finished running the script which needed by the way another line added to the attributes file because the scripts were converted to CRLF so, "scripts/* text eol=lf" was needed. the results of the run were 51 passed tests and 7 fails/errors:
test_reverse_dns_lookup and test_dangling_dns_detector
test_measure_request_time in the SQLi detector
A few mismatch assertions in test_bruter and test_api_scanner

kazet:
Did you look into the failed tests what they are caused by?

StefanCostache10:
I just looked into it and the failures were IndexError out of range for the reverse_dns and test_webapp_identifier tests, some case sensitive mismatches in test_bruter and an AssertionError: False is not true on sql_injection_detector so I think that they were not caused by my fix. But I also looked a bit further into some of the files and now I understand why you asked if *.sh text eol=lf . In the current scenario of my machines it was enough but I think that some other files like .patch, .txt, .yaml, and .yml might also lead to the same error, but I am not 100% sure. Should I check and add more rules to cover every type of file that has CRLF line endings?

kazet:
Before we can get this pull request merged, make sure the tests pass on Windows and there's nothing more to fix.

StefanCostache10:
After 2 days of trial and error with coming up with solutions I have reached a few conclusions. The tests associated with the bruter fail because windows has the case insensitive feature for files. I think that if I can make the module check if the same path has the same content it is likely that they are the same file. I saw that the test expects the output to start with config.dist, but because of the ASCII sorting /TEST takes precedence and fails the test. Regarding the reverse/dangling dns and sql injection tests I am starting to think that docker on windows might be at fault with how inconsistent/slow it is vs how nimble it's on linux. I can try to fix the module and if you think that tinkering with the tests to maybe find how much time  docker needs on windows to finish the tasks is acceptable, I could investigate increasing specific timeouts or adding a small buffer to the timing-based assertions to account for the overhead and jitter of Docker Desktop. In case you think that the tests should not be modified, I have no idea how the reverse/dangling dns and sql injection tests can pass

kazet:
Let's merge this one, it's still and improvement and making Artemis fully work on Windows is not top priority

kazet:
Let's merge this one, it's still and improvement and making Artemis fully work on Windows is not top priority

PR #2379, opened 2026-03-01
Minimize Nuclei DAST PoC URLs

### LINKED ISSUE #1977: Somehow minimize the PoC URLs generated by the Nuclei DAST templates as they're big

Sumit-ai-dev:
Fixes #1977 by minimizing the excessively long PoC URLs generated by Nuclei DAST templates.

Instead of including 100+ filler parameters in the vulnerability report, this uses Nuclei's `fuzzing_parameter` JSON field to isolate exactly which parameter was vulnerable and drops the rest. 

As requested in the issue, the logic is written defensively:
- **If `fuzzing_parameter` is found:** We keep only the vulnerable parameter (e.g. `?id=testing'`).
- **If not (fallback):** We return the full original URL unchanged to guarantee the PoC remains functional.

### Changes
- **`poc_url_utils.py`:** Added the core `minimize_nuclei_matched_at_url` function with the defensive fallback.
- **`reporter.py`:** Wired up the new function and cleaned up obsolete filler parameters.
- **Tests:** Added 8 unit tests covering all edge cases, the defensive fallback, and the exact SQLi payload from the issue.

### DIFF
--- a/artemis/modules/nuclei.py
+++ b/artemis/modules/nuclei.py
@@ -49,7 +49,7 @@ NUCLEI_TEMPLATES_LOCATION = "/root/nuclei-templates/"
 logger = logging.getLogger(__name__)
 
 
-def _verify_with_nuclei(url: str, template_id: str) -> bool:
+def _verify_with_nuclei(url: str, template_path: str) -> bool:
     try:
         command = [
             "nuclei",
@@ -57,7 +57,7 @@ def _verify_with_nuclei(url: str, template_id: str) -> bool:
             "-u",
             url,
             "-t",
-            os.path.join(NUCLEI_TEMPLATES_LOCATION, template_id),
+            template_path,
             "-dast",
             "-fuzzing-mode",
             "multiple",
@@ -700,10 +700,14 @@ class Nuclei(ArtemisBase):
 
             for finding in findings_per_task[task.uid]:
                 if "matched-at" in finding:
+                    template_path = finding.get(
+                        "template-path",
+                        os.path.join(NUCLEI_TEMPLATES_LOCATION, finding["template-id"]),
+                    )
                     finding["matched-at"] = minimize_nuclei_matched_at_url(
                         finding["matched-at"],
                         fuzzing_parameter=finding.get("fuzzing_parameter"),
-                        verify_url_fn=lambda url: _verify_with_nuclei(url, finding["template-id"]),
+                        verify_url_fn=lambda url: _verify_with_nuclei(url, template_path),
                     )
                 result.append(finding)
                 messages.append(


artemis/reporting/modules/nuclei/poc_url_utils.py
```diff
@@ -0,0 +1,34 @@
+"""Utilities for minimizing Nuclei DAST PoC URLs."""
+
+import urllib.parse
+from typing import Optional
+
+
+def minimize_nuclei_matched_at_url(
```
kazet, line 18: I see one thing missing - checking, that the resulting URL contains the vulnerability. We can have rare cases where the single-parameter URL doesn't work as multiple needs to be set (as the app checks them via isset()). In these cases we need to fallback to the original URL.

If this one gets fixed, we can get this PR merged.

Sumit-ai-dev: Added a check to make a GET request to the minimized URL. If it doesn't return 200, it falls back to the original. Just pushed!

kazet: What about re-running this particular Nuclei template? I know it's harder but will decrease the number of FPs

Sumit-ai-dev: I updated the logic to spawn the [nuclei](cci:1://file:///Users/sumitdas/Desktop/Artemis/Artemis/artemis/reporting/modules/nuclei/reporter.py:45:4-70:24) CLI in the background and actually re-run the template against the minimized URL.

I used the exact same flags and command structure from [artemis/modules/nuclei.py](cci:7://file:///Users/sumitdas/Desktop/Artemis/Artemis/artemis/modules/nuclei.py:0:0-0:0) to keep it consistent. If it doesn't trigger anymore (like if an `isset()` param was dropped), it safely falls back to the original URL.

Just pushed the update!

artemis/reporting/modules/nuclei/reporter.py
```diff
@@ -38,6 +43,33 @@ class NucleiReporter(Reporter):
...
+                "-t",
+                os.path.join(NUCLEI_TEMPLATES_LOCATION, template_id),
+                "-dast",
+                "-fuzzing-mode",
+                "multiple",
+                "-jsonl",
+                "-silent",
+            ]
+            result = subprocess.check_output(command, stderr=subprocess.DEVNULL)
+            return bool(result.strip())
```
kazet: I'm not sure checking that the result is nonempty is enough, could you add a little more checks?

artemis/reporting/modules/nuclei/reporter.py
```diff
@@ -38,6 +43,33 @@ class NucleiReporter(Reporter):
...
+        except FileNotFoundError:
+            # If nuclei isn't available (e.g. in some local dev envs), fail open
+            # to preserve the minimized URL so reporting doesn't break.
+            logging.getLogger(__name__).warning("Nuclei binary not found, skipping URL verification")
```
kazet: we have self.log ;)

artemis/reporting/modules/nuclei/reporter.py
```diff
@@ -38,6 +43,33 @@ class NucleiReporter(Reporter):
...
+            return True
```
kazet: IMO we should return False to be consistent with the comment

artemis/reporting/modules/nuclei/reporter.py
```diff
@@ -145,7 +177,13 @@ def _is_url_without_query_fragment(url: str) -> bool:
                 #
                 # We want to restrict this behavior only to services on empty path (e.g. redis://some-domain:6379) because
                 # if the path is nonempty, it may contain an exploit that may get filtered by e-mail filters.
-                matched_at = add_protocol_if_needed(vulnerability.get("matched-at", target))
+                matched_at = minimize_nuclei_matched_at_url(
```
kazet: Could you add an end-to-end for that? Along the lines of the tests in tests/modules/nuclei.py

artemis/reporting/modules/nuclei/reporter.py
Sumit-ai-dev: Fixed now parsing each JSONL line with json.loads() and only returning True if matched-at is present in the parsed output. Avoids false positives from verbose stdout noise.

artemis/reporting/modules/nuclei/reporter.py
Sumit-ai-dev: _verify_with_nuclei
 is a @staticmethod so self isn't available here. Added a module-level logger = logging.getLogger(__name__) instead — same effect, no more inline getLogger() calls.

artemis/reporting/modules/nuclei/reporter.py
Sumit-ai-dev: Done, changed to return False.

artemis/reporting/modules/nuclei/reporter.py
Sumit-ai-dev: Added  test_dast_matched_at_url_is_minimized in 
test_nuclei_autoreporter_integration.py— uses obtain_http_task_result against test-dast-vuln-app and asserts matched_ at URLs have ≤ 1 query parameter after minimization.

Sumit-ai-dev:
@kazet Addressed all your concerns, JSONL parsing with matched-at check, module-level logger, return False on missing binary, and the end-to-end test. Just pushed. Happy to adjust anything further. :)

kazet:
The test didn't pass, could you fix it?

Sumit-ai-dev:
@kazet Sure, just pushed the fix. Should pass now!

kazet:
Would it require much changes to also make it work on linux-lfi-fuzz?

Sumit-ai-dev:
Not much when fuzzing_parameter is absent, we could iterate each query param individually and verify each one to find which one triggers the vulnerability. Happy to implement once this PR merges. I can open an issue and raise a separate PR for it, or just raise a PR directly. Whatever you prefer, @kazet.

Sumit-ai-dev:
Turns out lfi-keyed doesn't trigger on the test app  only linux-lfi-fuzz fires. So I implemented the per-param iteration instead iterates each param individually, verifying each one, returns whichever triggers. Just pushed. :)

artemis/reporting/modules/nuclei/reporter.py
```diff
@@ -38,6 +48,38 @@ class NucleiReporter(Reporter):
     with open(Config.Modules.Nuclei.NUCLEI_TEMPLATE_GROUPS_FILE) as f:
         GROUPS = json.load(f)
 
+    @staticmethod
+    def _verify_with_nuclei(url: str, template_id: str) -> bool:
```
kazet: One minor comment - I think it would be optimal to do it during `scanning`, not `report generation`, so that e.g. generating multiple reports doesn't run multiple minimizations. After that we are good to go!

Sumit-ai-dev: Done moved the minimization to nuclei.py at scan time. 
_verify_with_nuclei
 is now a module-level function there, and matched-at gets minimized before saving to DB. Reporter just reads the already-minimized value. Pushed.

kazet:
if the tests pass, we can get it merged!

Issue #2397, opened 2026-03-03
Missing Type Hints in Data Modules


ajinkyap9:
## Description

Several data modules contain lists, dictionaries, and tuples without proper type annotations. Adding type hints improves code maintainability and enables better IDE support and static analysis with mypy.

## Affected Files

1. artemis/sql_injection_data.py
2. artemis/modules/data/common_sql_credentials.py
3. artemis/modules/data/api_scanner_data.py
4. artemis/modules/admin_panel_login_bruter.py

## Actual Output

```python
# artemis/sql_injection_data.py
INTERESTING_PATHS = [
    "/",
    "/api",
    "/admin",
]

ERROR_PATTERNS = {
    "mysql": ["SQL syntax", "mysql_fetch"],
    "postgresql": ["PostgreSQL", "pg_query"],
}
```

```python
# artemis/modules/data/common_sql_credentials.py
COMMON_SQL_CREDENTIALS = [
    ("root", "root"),
    ("admin", "admin"),
    ("sa", ""),
]
```

## Expected Output

```python
# artemis/sql_injection_data.py
from typing import List, Dict

INTERESTING_PATHS: List[str] = [
    "/",
    "/api",
    "/admin",
]

ERROR_PATTERNS: Dict[str, List[str]] = {
    "mysql": ["SQL syntax", "mysql_fetch"],
    "postgresql": ["PostgreSQL", "pg_query"],
}
```

```python
# artemis/modules/data/common_sql_credentials.py
from typing import List, Tuple

COMMON_SQL_CREDENTIALS: List[Tuple[str, str]] = [
    ("root", "root"),
    ("admin", "admin"),
    ("sa", ""),
]
```

## Steps to Fix

1. Open artemis/sql_injection_data.py
2. Add import: from typing import List, Dict
3. Add type hints to all module-level variables
4. Repeat for other affected files
5. Run mypy to verify: mypy artemis/sql_injection_data.py
6. Run tests: ./scripts/test

PR #2399, opened 2026-03-03
feat: add type hints to data modules for better maintainability

ajinkyap9:
## Description

Added type hints to data modules to improve code maintainability and enable better IDE support and static analysis with mypy.
Fixed- #2397 

## Changes

- artemis/sql_injection_data.py: Added List[str] and Dict[str, str] type hints
- artemis/modules/data/common_sql_credentials.py: Added List[Tuple[str, str]] type hint
- artemis/modules/data/api_scanner_data.py: Added List[str] and Dict[str, str] type hints

## Type of Change

- Code quality improvement

### DIFF
--- a/artemis/modules/data/api_scanner_data.py
+++ b/artemis/modules/data/api_scanner_data.py
@@ -1,10 +1,11 @@
 from os.path import dirname
+from typing import Dict, List
 
 with open(f"{dirname(__file__)}/swagger.txt", "r") as f:
-    COMMON_SPEC_PATHS = f.read().splitlines()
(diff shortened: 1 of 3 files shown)


artemis/modules/data/api_scanner_data.py
```diff
@@ -1,10 +1,11 @@
 from os.path import dirname
+from typing import Dict, List
```
kadewu, line 2: No reason to import typing anymore.
Just use build-in `list` , `dict` etc.

kadewu: And it applies of course to whole PR.

PR #2401, opened 2026-03-03
Reduce FP in dangling cname module

kadewu:


artemis/config.py
```diff
@@ -367,6 +367,20 @@ class DanglingDnsDetector:
             DANGLING_DNS_DELAY_STEP: Annotated[int, "Number of seconds for incremental step for retries."] = get_config(
                 "DANGLING_DNS_DELAY_STEP", default=600, cast=int
             )
+            DANGLING_DNS_KNOWN_DNS_ZONE_RECORDS_TO_SKIP: Annotated[
```
kazet, line 370: why do we have two settings about dangling skipping, isn't one sufficient?

artemis/modules/dangling_dns_detector.py
```diff
@@ -186,6 +188,10 @@ def _is_cname_dangling(self, record: Rdata, parent_domain: str) -> bool | None:
 
         cname_target_types = [rdatatype.A, rdatatype.AAAA, rdatatype.TXT]
         cname_target = record.target.to_text()  # type: ignore[attr-defined]
+        cname_target_zone = get_main_domain(cname_target)
+
+        if cname_target_zone in Config.Modules.DanglingDnsDetector.DANGLING_DNS_KNOWN_DNS_ZONE_RECORDS_TO_SKIP:
```
kazet, line 193: isn't it better to check whether it's registered? if cname_target_zone == 'wp.pl' it's still not claimable1

kadewu: Just checking if zone is registered is not sufficient.
E.g. something.github.io; github.io is registered, but provided example is claimable.

The list is purely for the zones that there is not way of claiming subdomains, but are someway used as cname targets (e.g: sipdir.online.lync.com -> I will add additional comment with that example)

kazet (approved):

Issue #2195, opened 2025-12-23
Documentation: example on how to add a scan using the REST API


fuzzy76:
The documentation for the REST API is very sparse, and I am not able to figure out how to do scans using it. Can the documentation be expanded in this area? I would like to automate scanning and sending reports using the REST API, so that teams can get a monthly report from their staging environment.

Likhith-Root:
Hi! I’m preparing for GSoC and would like to work on this issue. Can I take this?

kazet:
Sure! Remember to also write a test that extracts the commands from the docs, runs them and checks whether they work.

HarshitPal25:
assign it to me @kazet

kazet:
This issue is free to take for potential contributors

MiracleInvoker:
Hello, I would like to take this on.

I will write the RST docs covering the full API workflow (add targets -> monitor -> view results -> export reports) using only real endpoints from `api.py`, and add integration tests under `test/e2e/` that actually call the API and verify responses, following the same pattern as `test_automated_interaction.py`. No infrastructure changes. Could you assign this to me?

PR #2414, opened 2026-03-04
docs: add REST API scan example with tests

HarshitPal25:
## Description

Added REST API documentation with curl examples, a unit test that validates
the documented endpoints match the source code, and an e2e test that executes the documented
commands against the live API.

Closes: #2195

### Documentation (`docs/api/rest-api.rst`)
- **Authentication**: `API_TOKEN` in `.env`, `X-API-Token` header
- **Scan workflow**: `POST /api/add` → `GET /api/analyses` → `GET /api/num-queued-tasks` → `GET /api/task-results`
- **Export workflow**: `POST /api/export` → `GET /api/exports` → `GET /api/export/download-zip/{id}`
- **Other endpoints**: `stop-and-delete-analysis`, `is-blocklisted/{domain}`, `export/delete/{id}`
- All endpoints verified from `artemis/api.py` (mounted at `/api` in `main.py`)

### Unit Test (`test/unit/test_rest_api_docs.py`)
10 tests (Docker not  needed) that parse the docs and validate every curl command references a real route in `api.py`.

### E2E Test (`test/e2e/test_rest_api_docs.py`)
Extracts curl commands from the docs and executes them against the live API in the existing e2e pipeline.

### Docs Index (`docs/api/index.rst`)
Added `rest-api` to the toctree.

### DIFF
--- a/docker-compose.test.yaml
+++ b/docker-compose.test.yaml
@@ -26,7 +26,7 @@ services:
       RUNNING_TESTS: "true"
 
       # Use only one provider to speed up the tests
-      SUBDOMAIN_ENUMERATION_GAU_ADDITIONAL_OPTIONS: "--providers wayback"
+      SUBDOMAIN_ENUMERATION_GAU_ADDITIONAL_OPTIONS: "--providers wayback --timeout 60"
 
       SQL_INJECTION_STOP_ON_FIRST_MATCH: False
 
@@ -45,6 +45,9 @@ services:
       - ./test/:/opt/test/
       - ./docs/:/opt/docs/
       - ./docker/karton.ini:/etc/karton/karton.ini
+    depends_on:
+      test-grafana:
+        condition: service_healthy
 
   test-redis:
     image: redis:7.0.5
@@ -107,6 +110,11 @@ services:
     volumes:
       - ./test/data/php_easy_admin_password/:/var/www/html/
 
+  test-php-redirect-login:
+    image: php:8.0-apache
+    volumes:
+      - ./test/data/php_redirect_login/:/var/www/html
+
   test-php-mock-CVE-2020-28976:
     image: php:8.0-apache
     volumes:
@@ -125,12 +133,12 @@ services:
   test-ftp-server-with-easy-password:
     image: stilliard/pure-ftpd:latest
     environment:
-        - FTP_USER_NAME=admin
-        - FTP_USER_PASS=12345
-        - FTP_USER_HOME=/dev/shm
+      - FTP_USER_NAME=admin
+      - FTP_USER_PASS=12345
+      - FTP_USER_HOME=/dev/shm
 
   test-old-joomla:
-    image: joomla:4.0.5-php8.0-apache  # this is an old image on purpose
+    image: joomla:4.0.5-php8.0-apache # this is an old image on purpose
     environment:
       JOOMLA_DB_HOST: test-old-joomla-mysql
       JOOMLA_DB_PASSWORD: example
@@ -188,6 +196,11 @@ services:
     image: mysql:5.6
     environment:
       MYSQL_ROOT_PASSWORD: root
+
+  test-ssh-with-bad-key:
+    build:
+      context: test/data/ssh_bad_keys
+      dockerfile: Dockerfile
 
   test-ssh-with-easy-password:
     image: lscr.io/linuxserver/openssh-server:latest
@@ -256,6 +269,20 @@ services:
     build:
       context: test/images/nginx-with-sni-tls
       dockerfile: Dockerfile
+
+  test-grafana:
+    image: grafana/grafana:latest
+    environment:
+      - GF_SECURITY_ADMIN_USER=admin
+      - GF_SECURITY_ADMIN_PASSWORD=admin
+      - GF_AUTH_DISABLE_LOGIN_FORM=false
+      - GF_SECURITY_DISABLE_BRUTE_FORCE_LOGIN_PROTECTION=true
+    healthcheck:
+      test: [ "CMD-SHELL", "wget -qO /dev/null http://localhost:3000/api/health" ]
+      interval: 5s
+      timeout: 10s
+      retries: 12
+      start_period: 60s
 
 volumes:
   data-test:
--- a/docs/api/rest-api.rst
+++ b/docs/api/rest-api.rst
@@ -250,10 +250,8 @@ Stop and delete an analysis
 .. code-block:: bash
 
    # rest-api-stop-analysis
-   curl -s -X POST http://localhost:5000/api/stop-and-delete-analysis \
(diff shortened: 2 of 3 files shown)


HarshitPal25:
hi @kazet check them out i have analyse what the last person pr lagging and added those in it too

kazet:
Would you fix the linter issues?

HarshitPal25:
yes ofcourse now please have a look at it

test/unit/test_rest_api_docs.py
```diff
@@ -0,0 +1,261 @@
...
+    path = re.sub(r"/is-blocklisted/[a-zA-Z0-9._-]+", "/is-blocklisted/{domain}", path)
+    return path
+
+
+class TestRestApiDocumentation(unittest.TestCase):
+    """Verify that the REST API documentation references real, existing endpoints."""
+
+    def setUp(self) -> None:
+        if not os.path.exists(DOCS_PATH):
+            self.skipTest(f"REST API documentation not found at {DOCS_PATH}")
```
kazet: Instead of skipping the test, I'd prefer making sure the docs_path is working

HarshitPal25: i have made the changes now instead of skiptest i revert back to assertTrue and the dock path will work as i mounted it in docker-compose.test.yaml

HarshitPal25:
@kazet i revert the changes to assertTrue frp, skiptest as you said and added the path for the doc in docker-compose.test.yaml

test/e2e/test_rest_api_docs.py
```diff
@@ -0,0 +1,245 @@
...
+        allow_redirects=allow_redirects,
+    )
+
+
+class RestApiDocExamplesTestCase(BaseE2ETestCase):
+    """Test that the curl commands documented in docs/api/rest-api.rst actually work."""
+
+    def test_documented_curl_commands_work(self) -> None:
+        """Extract curl commands from the REST API docs and verify each one works against the live API."""
+        commands = extract_curl_commands(DOCS_PATH)
```
kadewu: In e2e test do not rely on documentation.
Just run the tests as you suggested to in the documentation.

The e2e test should break when code changes, not when documentation change.

test/unit/test_rest_api_docs.py
```diff
@@ -0,0 +1,265 @@
+"""Unit test to verify that REST API documentation references real API endpoints.
```
kadewu: I'm fine with adding tests to check if the documentation is up to date,
but let's move it from units, as this is not unit testing. 

So under `test/documentation` would probably suit best.

test/e2e/test_rest_api_docs.py
HarshitPal25: E2E test rewritten — test/e2e/test_rest_api.py now calls the API directly instead of parsing docs.

test/unit/test_rest_api_docs.py
HarshitPal25: test/documentation/test_rest_api_docs.py from test/unit/ since it validates documentation

HarshitPal25:
@kadewu @kazet i have make the changes as you demand please check it out

kadewu:
The tests you did write seems to fail.
Please make sure they actually pass.

HarshitPal25:
@kadewu run the tests now they should have been cleared

kadewu:
Please also make sure to have pre-commit installed as described in README.

HarshitPal25:
hello @kadewu i got a bit late because i got into an accident but now i have fixed the error and check for them too
Screenshot 2026-03-13 at 16 56 58

HarshitPal25:
let me check the error and fix it

HarshitPal25:
@kadewu  i have made the changes and 
Added graceful Docker volume creation before unit tests (2>/dev/null || true)
Added volume recreation before e2e tests with error handling
Added 120-second web service readiness check (polls http://localhost:5000/ for HTTP 200/302/404)

if it won't work fine i have got to know the error explained to me by on log tab i would add another changes to the (rest_api test ) and make the deletion test robust to async operations:

HarshitPal25:
@kadewu please check it out

kadewu:
I don't see how the changes introduced in last commit may help with the failure of test.
Please do not introduce changes not related to the PR. Not to mention that waiting for the services is already implemented. Please revert last commit.

Did you run the tests you have written? The test fails cause is incorrect.
Also the commit (#ea65a3915ca5c0aa4c01f8107cdd80a13b0d3f99) should also not be needed.
Please try to dig and understand why the test fail.

HarshitPal25:
@kadewu i have analyse the test this time i use help of claude to understand the problem and fix the problem i have revert the changes for the docker volume fix documentation test and simplify e2e test

kadewu:
This PR is about writing docs and simple test for already existing endpoints. There is no valid reason to adjust current working code in favor of test.

And this is last warning; All I'm asking is to run test locally and fix it. If there is problem with documentation on how to do that with Artemis then raise a proper issue or ask here, I can help.

HarshitPal25:
Okey I am working on it now I am using Azure cloud to run tests on it i would bring results to you for sure and sorry for the inconveniences sir

HarshitPal25:
i am closing this one and opening a new pr with all the test cases passes..

HarshitPal25:
@kazet check and review the pr

kadewu:
There is no reason to ping anyone, not to mention you've got the review from me.
So back to the thing, you already got an answer in previously opened PR - this is not the way to test it.

Why did you remove old commits and tests?

Please bring back the old commits for history and write the tests properly. The direction was right, but you just needed to fix one test.

HarshitPal25:
okey i would do it i wanna ask you said the commit (#ea65a3915ca5c0aa4c01f8107cdd80a13b0d3f99) is not needed it contails the integration test cases that kazet asks to include in the pr...

kadewu:
The mentioned commit did not add those tests.

edit: and also changes from this one were correctly reverted, as you even said so.

HarshitPal25:
@kadewu hi please check out and review this time i have run the CI test and analyse it and fix the failure happen because of the documentation

by the way i run into some failed test which include files named test_dangling_dns and test_wordpress_plugin i thought it maybe because i am running them locally but i make changes in them too but in this commit i didn;t add those files if the CI test fails and those test files are the reason i would add them and make sure every single test cases passes

thanks...

kadewu:
How could you possibly fix the failure if the failing test is missing alongside other e2e tests that you were supposed to write?

HarshitPal25:
i write the integration test in the test_api instead of different e2e test i can implement that and work on it
and those test i suppose they are failing locally as they require live integration

kadewu:
You did not write e2e tests neither reverted back the old ones.
This PR is closed and that's final decision.

Issue #2378, opened 2026-02-28
Check whether we can make admin_panel_login_bruter better, e.g. by supporting more types of admin panels


kazet:

Sumit-ai-dev:
Hey @kazet, went through the module and had a couple of ideas.

The form parser currently checks for username/password field names before handling hidden fields  this means if a hidden field's name happens to contain a common keyword, its value gets overwritten instead of preserved. Putting the hidden field check first would make it more robust for any panel where the CSRF token field name overlaps with the username heuristics.

Also noticed some panels like Portainer skip forms entirely and just expect a JSON body. A small fallback when no form is found  using the same session that's already open would cover them without adding anything new.

Plus a few missing paths and failure messages for these panels.

Happy to discuss and happy to be assigned!

kazet:
These are good ideas to implement, assigned you! What I think would be awesome would be to implement them in separate PRs, starting from the easiest one - that will make it easier to get them merged.

Sumit-ai-dev:
Hi @kazet 
Opened both as separate PRs as you suggested:

- #2415 — hidden field fix (CSRF tokens preserved)
- #2416 — JSON login fallback for Grafana/Portainer + missing paths/messages

Starting with #2415 as it's the simpler one. :)

PR #2416, opened 2026-03-04
Feat: fallback to JSON login for panels without HTML forms

Sumit-ai-dev:
Panels like Grafana and Portainer don't use HTML forms — they expect a JSON POST. When no form is found, the bruter now tries a JSON login using common payload formats and detects success via response tokens.

Also added missing paths (`/api/auth`, `/grafana/login`) and failure messages for Portainer and Netdata.

Part of #2378

### DIFF
--- a/artemis/modules/admin_panel_login_bruter.py
+++ b/artemis/modules/admin_panel_login_bruter.py
@@ -38,8 +38,8 @@ with open(os.path.join(os.path.dirname(__file__), "data", "admin_panel_login_bru
 # JSON-only API login endpoints that do not respond to GET with 200 (POST-only).
 # These are always tried regardless of discovery, because check_url(GET) would fail.
 JSON_API_PATHS = [
-    "/api/login",   # Grafana
-    "/api/auth",    # Portainer and other tools
+    "/api/login",  # Grafana
+    "/api/auth",  # Portainer and other tools
 ]
 
 


artemis/modules/admin_panel_login_bruter.py
```diff
@@ -69,6 +67,71 @@ def check_url(self, url: str) -> bool:
...
+                )
+                has_error = "error" in data or data.get("status", "") == "error"
+
+                if has_token and not has_error:
+                    self.log.info(
+                        "successful JSON API brute force on %s username=%s password=%s", login_url, username, password
+                    )
+                    return (
+                        True,
+                        AdminPanelLoginBruterResult(
```
kazet, line 129: Before merging, let's add a test against a real application (e.g. Grafana, whatever is simplest) - starting it via docker-compose and checking whether the module is able to bruteforce the login

Sumit-ai-dev: Added Grafana to [docker-compose.test.yaml](cci:7://file:///Users/sumitdas/Desktop/Artemis/Artemis/docker-compose.test.yaml:0:0-0:0) and a [test_grafana_json_login](cci:1://file:///Users/sumitdas/Desktop/Artemis/Artemis/test/modules/test_admin_panel_login_bruter.py:29:4-44:88) test in [test_admin_panel_login_bruter.py](cci:7://file:///Users/sumitdas/Desktop/Artemis/Artemis/test/modules/test_admin_panel_login_bruter.py:0:0-0:0).

Grafana was the cleanest fit no HTML form on `/login`, pure JSON API, exactly the case [_try_json_login](cci:1://file:///Users/sumitdas/Desktop/Artemis/Artemis/artemis/modules/admin_panel_login_bruter.py:69:4-132:34) handles. Tested locally against a real container, works as expected.

kazet:
If the pipeline passes, we can get it merged

kazet:
The test didn't pass, could you fix it?

Sumit-ai-dev:
> The test didn't pass, could you fix it?

Fixed  Grafana just needed a healthcheck so the tests wait for it to be ready. Pushed.

kazet:
I think fixing Grafana is a better way than mocking it - it failed for a reason.

Sumit-ai-dev:
It’s fixed. You can check it now.

kazet:
Could you random.shuffle() the credentiasl before checking? This way even if the panel has some brute force protection we won't check the same pairs of credentails every time

kazet (approved):

PR #2417, opened 2026-03-04
Detect domain controller services exposed to the internet

### LINKED ISSUE #2400: Detect domain controllers exposed to the internet
Or, maybe, instead of domain controllers, check what risky services they tend to expose and detect them.

Remember that we don't want reports about every port, only the risky ones.

Sumit-ai-dev:
New module that detects known domain controller service ports exposed to the internet.

Filters on `SERVICE / UNKNOWN` tasks and checks the port against known DC ports. Flags as `INTERESTING` if matched.

Ports covered:
- 88 — Kerberos
- 389 — LDAP
- 445 — SMB
- 636 — LDAPS
- 3268 — Global Catalog LDAP
- 3269 — Global Catalog LDAPS

Also adds ports 636, 3268, 3269 to the short port scan list.

Closes #2400


Sumit-ai-dev:
@kazet could you review it as well? I’m happy to discuss and update it.

artemis/modules/domain_controller_detector.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if port not in DC_PORTS:
+            self.db.save_task_result(task=current_task, status=TaskStatus.OK, status_reason=None)
+            return
+
+        service_name = DC_PORTS[port]
+        self.db.save_task_result(
+            task=current_task,
+            status=TaskStatus.INTERESTING,
+            status_reason=f"{service_name} (port {port}) exposed on {host} — likely a domain controller",
```
kazet: I don't think this is a correct inference - you might have hosts that have the ports exposed but aren't a domain controller.

What about reporting these ports separately, as they pose a risk even if it's not a domain controller?

E.g. currently the port_scanner module reports SMB and Nuclei reports LDAP. Could you make sure all others are covered as well?

BTW, are you sure it's the correct/full port list?

Sumit-ai-dev: Hi @kazet! Thanks for the review. I've addressed all the points:

1. **Removed the inference:** The module no longer states `— likely a domain controller`. 
2. **Reported separately:** Each port is now reported as its own distinct finding based on the specific service (e.g., `Kerberos (port 88) exposed on <host>`).
3. **Full port list:** I reviewed the MS Active Directory firewall docs and added the missing ports to complete the list so all others are covered:
   - `135` (RPC Endpoint Mapper)
   - `464` (Kerberos Password Change)
   - `593` (RPC over HTTP)
   - `5722` (DFSR)
   - `9389` (Active Directory Web Services)

**One design question :**
To ensure these missing AD ports are actually detected (and not skipped by the scanner), I added them to [ports-artemis-short.txt](cci:7://file:///Users/sumitdas/Desktop/Artemis/Artemis/artemis/modules/data/ports-artemis-short.txt:0:0-0:0). Since that file says _"Top 100 ports"_, is that the right place for them, or would you prefer a separate DC-specific port list? 

Let me know what you think!

Sumit-ai-dev:
@kazet  u can review it :)

Sumit-ai-dev:
@kazet and @kadewu sir, can you review this

artemis/modules/domain_controller_detector.py
```diff
@@ -0,0 +1,62 @@
...
+    636: "LDAPS",
+    3268: "Global Catalog LDAP",
+    3269: "Global Catalog LDAPS",
+    5722: "DFSR (Distributed File System Replication)",
+    9389: "Active Directory Web Services",
+}
+
+
+@load_risk_class.load_risk_class(load_risk_class.LoadRiskClass.MEDIUM)
+class DomainControllerDetector(ArtemisBase):
```
kadewu, line 34: I don't think this is worth having separate module.
Let's have a proper fallback in port_scanner.

artemis/modules/domain_controller_detector.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if port not in DC_PORTS:
+            self.db.save_task_result(task=current_task, status=TaskStatus.OK, status_reason=None)
+            return
+
+        service_name = DC_PORTS[port]
+        self.db.save_task_result(
+            task=current_task,
+            status=TaskStatus.INTERESTING,
+            status_reason=f"{service_name} (port {port}) exposed on {host}",
```
kadewu, line 57: Simple check like that, that would assign service based on port is not acceptable.
Please introduce proper probing.

And if not service detected - let's have the list of those ports that we will inform that they are open - but for that we need a proper list. Please do proper research on that.

artemis/modules/data/ports-artemis-short.txt
```diff
@@ -1,2 +1,2 @@
 # Top 100 ports detected by Artemis
-80,443,21,993,143,995,110,465,587,25,3306,53,22,2222,5432,222,8080,990,8443,111,8888,10000,144,81,8081,8008,82,27017,444,84,83,3000,20000,8000,1022,1500,4443,9100,5000,8010,10443,3389,2121,9080,1433,9000,9001,389,7070,9999,88,2000,6001,26,554,8009,1029,2001,1025,2022,5060,6000,5900,631,5051,5009,119,1028,543,1720,37,2049,5800,445,9090,8001,8082,8090,666,23,9443,50001,4000,2522,8083,8181,3128,1935,8088,873,2200,4433,7443,8085,3001,139,30000,3003,85,1984,90
+80,443,21,993,143,995,110,465,587,25,3306,53,22,2222,5432,222,8080,990,8443,111,8888,10000,144,81,8081,8008,82,27017,444,84,83,3000,20000,8000,1022,1500,4443,9100,5000,8010,10443,3389,2121,9080,1433,9000,9001,389,7070,9999,88,2000,6001,26,554,8009,1029,2001,1025,2022,5060,6000,5900,631,5051,5009,119,1028,543,1720,37,2049,5800,445,9090,8001,8082,8090,666,23,9443,50001,4000,2522,8083,8181,3128,1935,8088,873,2200,4433,7443,8085,3001,139,30000,3003,85,1984,90,636,3268,3269,135,464,593,5722,9389
```
kadewu, line 2: I don't think this is a place for that, not to mention that comment now is inaccurate.
We can e.g: introduce separate config for risky ports that we will scan against.

kazet:
Closed due to inactivity

PR #2420, opened 2026-03-04
Add support for configuring Subfinder providers via environment variable

imraflip:
This PR adds support for configuring Subfinder provider API keys directly through the Artemis configuration system

Previously, Subfinder expected the `provider-config.yaml` file to exist inside `~/.config/subfinder/`, which usually required users to either run Subfinder once to generate the file or manually create and populate it inside the container

With this change, Artemis can now generate `provider-config.yaml` automatically if API keys are provided via the `.env` file using the `SUBFINDER_API_KEYS` variable. The value is expected to be in JSON format and will be converted into the YAML file that Subfinder expects

The file is generated during module initialization, so it exists before Subfinder runs. This means provider configuration works immediately without requiring a first manual run of Subfinder

If the variable is not set, Subfinder behaves as before and runs without provider API keys

### DIFF
--- a/artemis/modules/subdomain_enumeration.py
+++ b/artemis/modules/subdomain_enumeration.py
@@ -1,13 +1,13 @@
 #!/usr/bin/env python3
 import binascii
+import json
 import os
 import time
-import json
-import yaml
 import urllib.parse
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Set
 
+import yaml
 from karton.core import Consumer, Task
 from karton.core.config import Config as KartonConfig
 from publicsuffixlist import PublicSuffixList


artemis/config.py
```diff
@@ -1062,6 +1062,11 @@ class SubdomainEnumeration:
                 "Time to sleep between retries for subdomain enumeration in seconds.",
             ] = get_config("SUBDOMAIN_ENUMERATION_SLEEP_TIME_SECONDS", default=60, cast=int)
 
+            SUBFINDER_PROVIDER_CONFIG: Annotated[
```
kazet, line 1065: I think a good idea would be to provide an example

artemis/modules/subdomain_enumeration.py
```diff
@@ -45,6 +48,8 @@ def __init__(self, db: Optional[DB] = None, *args, **kwargs) -> None:  # type: i
         # before we migrate the tasks, let's create binds to make sure the new tasks will hit the queue of this module
         self.backend.register_bind(self._bind)
 
+        self._ensure_subfinder_provider_config()
```
kazet, line 51: what would happen if the keys don't work? Does subfinder skip them quietly or something else happens?

artemis/config.py
imraflip: Done, I just added the example

artemis/modules/subdomain_enumeration.py
imraflip: image

Yes, I tested it directly inside the container with invalid API keys and subfinder still runs normally. It just skips and won't use the providers that fail, so it doesn’t break the module nor the subfinder

PR #2422, opened 2026-03-05
fix: handle git clone failure in BaseNewerVersionComparerModule

WHOIM1205:
### Summary
BaseNewerVersionComparerModule was cloning the endoflife.date repository
using subprocess.call without checking the return code. If the clone failed
(for example due to network issues or GitHub being unreachable), the code
would still continue and try to read a file from the repository that was
never created. This resulted in a FileNotFoundError during module
initialization.

Because this happens in the constructor, the worker process would crash
during startup and containers such as joomla_scanner and drupal_scanner
could enter restart loops.

### Changes
- Replace subprocess.call with subprocess.check_call to ensure failures are detected
- Add a timeout to prevent git clone from hanging indefinitely
- Clean up the temporary directory if the clone fails
- Raise a clearer error message when the repository cannot be fetched

### Impact
This prevents startup crash loops when GitHub is temporarily unreachable and
provides a clearer failure path during initialization. It also avoids leaving
temporary directories behind when the clone operation fails.

### DIFF
--- a/artemis/modules/base/base_newer_version_comparer.py
+++ b/artemis/modules/base/base_newer_version_comparer.py
@@ -25,8 +25,7 @@ class BaseNewerVersionComparerModule(ArtemisBase):
         except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
             shutil.rmtree(self.endoflife_data_folder, ignore_errors=True)
             raise RuntimeError(
-                f"Failed to clone endoflife data repository: {e}. "
-                "Check network connectivity to github.com."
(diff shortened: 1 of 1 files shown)


WHOIM1205:
heyy @kazet 
This change adds proper error handling around the git clone used to fetch endoflife data.

Previously the code used subprocess.call and ignored the return code. If the clone failed the code would still try to read the product file which resulted in a FileNotFoundError during initialization and could cause scanner containers to crash and restart repeatedly.

The fix switches to subprocess.check_call adds a timeout and cleans up the temporary directory if the clone fails so initialization fails in a clearer and safer way.

Issue #2410, opened 2026-03-04
Add parameter minimization to sql_injection_detector/lfi_detector


kazet:
So that the PoCs aren't very long and contain only the parameter that is needed to trigger SQLi/LFI

imraflip:
Hi, is anyone already working on this? If not, I'd love to try implementing it

Divyateja2709:
Hi @kazet can you assign this to me

kazet:
@Divyateja2709 the issue is already taken and Rafli has already submitted a PR.

PR #2423, opened 2026-03-05
Add parameter minimization to LFIDetector and fix batching bug

imraflip:
The previous implementation of batching logic compared the loop index against `len(URL_PARAMS) - 1` while iterating over a combined parameter list. This could cause the last batch condition to behave incorrectly when the original URL already had parameters. The condition is now based on the actual combined parameter list, ensuring the final batch is always processed properly.

This PR also implements parameter minimization for generated PoCs. Previously, when LFI (or RCE) was detected, the PoC URL contained the full injected parameter batch, which often made it unnecessarily long. After detection, the module now re-tests each parameter individually and keeps only those that independently reproduce the issue. If none of them work alone, it falls back to the original batch to avoid changing detection behavior.

The tests were updated accordingly to validate behavior.

#2410

### DIFF
--- a/test/modules/test_lfi_detector.py
+++ b/test/modules/test_lfi_detector.py
@@ -24,8 +24,7 @@ class LFIDetectorTestCase(ArtemisModuleTestCase):
 
             self.assertEqual(call.kwargs["status"], TaskStatus.INTERESTING)
             self.assertIn(
-                "It appears that this URL is vulnerable to LFI: "
-                "http://test-apache-with-lfi-and-rce:80/page.php?id=",
+                "It appears that this URL is vulnerable to LFI: " "http://test-apache-with-lfi-and-rce:80/page.php?id=",
                 call.kwargs["status_reason"],
             )
 
@@ -34,8 +33,7 @@ class LFIDetectorTestCase(ArtemisModuleTestCase):
                 call.kwargs["status_reason"],
             )
             self.assertIn(
-                "It appears that this URL is vulnerable to RCE: "
-                "http://test-apache-with-lfi-and-rce:80/page.php?id=",
+                "It appears that this URL is vulnerable to RCE: " "http://test-apache-with-lfi-and-rce:80/page.php?id=",
                 call.kwargs["status_reason"],
             )
 


test/modules/test_lfi_detector.py
```diff
@@ -24,12 +24,25 @@ def test_lfi_detector_with_rce(self) -> None:
...
+                "%0a/bin/cat%20/etc/passwd",
```
kazet, line 41: why the full URL match was removed? The tests would be more robust if you matched the full URL, not a single parameter value

test/modules/test_lfi_detector.py
```diff
@@ -24,12 +24,25 @@ def test_lfi_detector_with_rce(self) -> None:
...
                 call.kwargs["status_reason"],
             )
+
+            self.assertIn(
+                "page.php?id=",
+                call.kwargs["status_reason"],
+            )
+
+            self.assertIn(
+                "etc/passwd",
```
kazet, line 32: why the full URL match was removed? The tests would be more robust if you matched the full URL, not a single parameter value

artemis/modules/lfi_detector.py
```diff
@@ -71,6 +71,33 @@ def contains_lfi_indicator(self, original_response: HTTPResponse, response: HTTP
...
+
+        for param in params:
+            test_url = self.create_url_with_batch_payload(url, [param], payload)
+            response = self.http_get(test_url)
+
+            if self.contains_lfi_indicator(original_response, response):
+                minimal_params.append(param)
+
+        if minimal_params:
+            self.log.debug(
```
kazet: i'd make it an INFO

test/modules/test_lfi_detector.py
imraflip: I removed the full URL match because previously the PoC contained a long batch of injected parameters, so the test was tightly coupled to that implementation and became quite noisy. My goal was to avoid depending on the exact batch composition and instead just verify that the vulnerability was detected and that the payload appeared

Since the detector now minimizes the PoC to the actual vulnerable parameter (`id`), the resulting URL is deterministic again. I’ve pushed an update to the tests to match the full minimized URL instead

PR #2426, opened 2026-03-06
refactor: factor out common parts of lfi_detector and sql_injection_detector

### LINKED ISSUE #2409: Factor out common parts of lfi_detector and sql_injection_detector - the modules have common structure and a lot can be factored out.

HarshitPal25:
Closes #2409

## Summary

Both `lfi_detector` and `sql_injection_detector` had a lot of duplicated structure and logic. This PR introduces `InjectionDetectorBase`, a shared base class that centralizes the common parts, while each module retains its specific scanning logic.

## What was factored out into `InjectionDetectorBase`

| Element | Description |
|---|---|
| `num_retries` | Class attribute (both used the same `SLOW_MODULE_NUM_RETRIES`) |
| `filters` | HTTP service filter (identical in both) |
| `_strip_query_string()` | URL utility method (duplicated) |
| `is_url_with_parameters()` | Static method (duplicated) |
| `create_url_with_batch_payload()` | Payload URL builder (duplicated) |
| `_collect_urls()` | URL collection, deduplication, filtering & shuffling (was ~10 identical lines in each `run()`) |
| `run()` | Template method pattern — calls hooks for scanning and result formatting |

## How the subclasses work

Each module implements three abstract methods:
- **`scan(urls, task)`** — module-specific vulnerability scanning logic
- **`create_status_reason(messages)`** — formats findings into a status string
- **`create_data(messages)`** — builds the data payload for task results

`LFIDetector` also overrides `_pre_run_check()` to call `check_connection_to_base_url_and_save_error()` before scanning.

## No test changes needed

All existing imports (`LFIDetector`, `SqlInjectionDetector`, `LFIFindings`, `Statements`) are still exported from the same modules. The public API, output format, and behavior are fully preserved.

## Files changed

- **New:** `artemis/modules/injection_detector_base.py` — shared base class
- **Modified:** `artemis/modules/lfi_detector.py` — now extends `InjectionDetectorBase`
- **Modified:** `artemis/modules/sql_injection_detector.py` — now extends `InjectionDetectorBase`


HarshitPal25:
@kazet here i have centralized the common part of Both lfi_detector and sql_injection_detector, while retaining their specific scanning logic. added a base class for both the file called `InjectionDetectorBase` in file  `artemis/modules/injection_detector_base.py `

artemis/modules/injection_detector_base.py
```diff
@@ -0,0 +1,104 @@
...
+
+from artemis.binds import Service, TaskStatus, TaskType
+from artemis.config import Config
+from artemis.crawling import get_links_and_resources_on_same_domain
+from artemis.module_base import ArtemisBase
+from artemis.modules.data.static_extensions import STATIC_EXTENSIONS
+from artemis.task_utils import get_target_url
+
+
+class InjectionDetectorBase(ArtemisBase):
```
kadewu: I would strongly advise against this approach.
Right now it may work, but then in the future we will most likely add another scanner that has similar approach to lfi etc. and we will need to adjust code just to this.

In artemis you can see that we have a lot of utils-like functions.
Could you please do it this way? While doing so, you can have in mind that reusability not only in the currently extracted two modules.

HarshitPal25: Thanks for the feedback! Reworked it to use standalone utility functions in a new `injection_utils.py`, following the existing **utils.py** pattern. Extracted

- `strip_query_string`— strips query string and fragment from a URL

- `is_url_with_parameters`— checks if a URL contains query parameters
- `create_url_with_batch_payload`— builds a URL with multiple params set to a given payload
- `collect_urls_to_scan`— handles the full pipeline of gathering, filtering, deduplicating, and shuffling URLs before scanning

all plain functions, so any future scanner can reuse them directly without any structural changes.

HarshitPal25:
@kazet hey bro checkout my work as the review is pending

kazet:
I think there is more than can be extracted to the common module - LFI detector and SQL injection detector have very similar structure.

HarshitPal25:
@kazet i have made the changes and change_url_params, create_status_reason and create_scan_result_data is it great if any changes require please let me know

kadewu:
I've missed update on that one, sorry for that.
In the meantime conflicts were introduced, can you fix them.

Also as for review; it seems that the logic how scanning actually works can be extracted?

HarshitPal25:
Umm like conflicts were introduced? Can you elaborate it and for scanning logic okey I can extract it too and work on it

kadewu:
> Umm like conflicts were introduced? Can you elaborate it

I'm not sure what to elaborate on. Can you ask more specific question?

HarshitPal25:
oh sorry i didn't see the conflicts that were introduced in sql_injection_detector.py i would resolve them and i would extract the scanning.

HarshitPal25:
Hi @kadewu @kazet,

1.  Merged upstream/main and resolved conflicts in sql_injection_detector.py
2. Extracted common result processing logic into `process_and_save_scan_results()` helper in injection_utils.py
3. Updated both sql_injection_detector and lfi_detector to use the new helper function
4. Fixed test_lfi_detector.py to patch the module-level import after the refactoring

The helper function now handles status determination, status_reason creation, and database saving for both injection detectors, reducing code duplication.

PR #2428, opened 2026-03-06
fix: remaining blocking I/O in async endpoints

WHOIM1205:
## Fix blocking I/O in async endpoints

Some async endpoints were still calling synchronous DB/Redis operations directly, which can block the asyncio event loop under load.

This wraps the blocking calls with `asyncio.to_thread(...)` so they run in a background thread instead of freezing the event loop.

### Changes
- `frontend.py`: run `create_tasks()` via `await asyncio.to_thread(...)` in `post_add`
- `api.py`: wrap `db.delete_report_generation_task()` in `post_export_delete`
- `api.py`: wrap `db.create_report_generation_task()` in `post_export`

### Impact
Keeps the event loop responsive during large submissions or export operations and prevents request hangs for other users.

### DIFF
--- a/artemis/api.py
+++ b/artemis/api.py
@@ -1,3 +1,4 @@
+import asyncio
 import datetime
 from typing import Annotated, Any, Dict, List, Optional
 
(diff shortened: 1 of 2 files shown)


WHOIM1205:
hey @kazet
This endpoint is declared as async but it directly calls synchronous DB and Redis operations (via create_tasks and report task functions). Because of this large submissions or export operations can block the event loop and temporarily freeze other request
Wrapping these calls with asyncio.to_thread allows the blocking work to run in a background thread and keeps the event loop responsive

PR #2429, opened 2026-03-06
fix: prevent DetachedInstanceError in report generation tasks

WHOIM1205:
### Problem
Some database methods were returning SQLAlchemy ORM objects from inside a session context (`with self.session() as session:`). Once the session closed, these objects became detached. Accessing their attributes later could raise `DetachedInstanceError`, especially with SQLAlchemy 2.0 where expired attributes on detached instances are strictly enforced.

This affected report generation tasks where the worker accessed task attributes after the session had already been closed, potentially causing repeated failures while processing pending tasks.

### Fix
- Expunge ORM objects from the session before returning them so their loaded attributes remain accessible after the session closes.
- Use `session.merge()` when saving task results to properly reattach detached instances to the active session.
- Apply the same expunge pattern to methods returning task lists or single tasks.

### Impact
- Prevents `DetachedInstanceError` during report generation.
- Ensures tasks can be processed and updated reliably.
- Improves compatibility with SQLAlchemy 2.0+ session behavior.

### DIFF
--- a/artemis/api.py
+++ b/artemis/api.py
@@ -1,3 +1,4 @@
+import asyncio
 import datetime
 from typing import Annotated, Any, Dict, List, Optional
 
@@ -232,7 +233,7 @@ def download_zip(id: int) -> RedirectResponse:
 @router.post("/export/delete/{id}", dependencies=[Depends(verify_api_token)])
 async def post_export_delete(id: int) -> Dict[str, Any]:
     """Delete an export."""
-    db.delete_report_generation_task(id)
+    await asyncio.to_thread(db.delete_report_generation_task, id)
     return {
         "ok": True,
     }
@@ -265,7 +266,8 @@ async def post_export(
     skip_suspicious_reports: bool = Body(False),
 ) -> Dict[str, Any]:
     """Create a new export. An export is a request to create human-readable messages that may be sent to scanned entities."""
-    db.create_report_generation_task(
+    await asyncio.to_thread(
+        db.create_report_generation_task,
         skip_previously_exported=skip_previously_exported,
         tag=tag,
         comment=comment,
--- a/artemis/frontend.py
+++ b/artemis/frontend.py
@@ -182,7 +182,7 @@ async def post_add(
             csrf_protect,
         )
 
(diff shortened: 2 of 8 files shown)


WHOIM1205:
heyy @kazet 
Hi maintainer, this fixes an issue where ReportGenerationTask objects were returned after the session closed and later accessed in the worker. That can cause detached instance errors. The change expunges the object before returning it and uses session.merge when saving results so the task is properly attached to the new session.

artemis/db.py
```diff
@@ -556,11 +556,14 @@ def task_to_dict(self, task: Task) -> Dict[str, Any]:
...
     def take_single_report_generation_task(self) -> Optional[ReportGenerationTask]:
         with self.session() as session:
-            return (  # type: ignore
+            task = (
                 session.query(ReportGenerationTask)
                 .filter(ReportGenerationTask.status == ReportGenerationTaskStatus.PENDING.value)
                 .first()
             )
+            if task:
+                session.expunge(task)
```
kadewu, line 565: AFAIK It will not prevent `DetachedInstanceError` , how did you come up to that conclusion?
And in what artemis flow did you encounter this error anyway?

WHOIM1205:
@kadewu 
thanks for pointing that out you are right that expunge alone would not fully prevent DetachedInstanceError i updated the change so we no longer pass orm objects across session boundaries the save_report_generation_task_results method now takes task_id and re queries the task inside its own session before updating it the expunge calls remain only as a defensive measure for the read methods

kazet:
Did you encounter this error in practice? If yes, in what usage?

WHOIM1205:
> Did you encounter this error in practice? If yes, in what usage?

i did not encounter it during a normal run i noticed the pattern while reviewing the reporting task flow where orm objects were returned from a closed session and later accessed in the worker loop in sqlachemy this can lead to detached instance issues depending on session lifecycle changes so i proposed the change to avoid passing orm objects across session boundaries

kazet:
Let's not solve theoretical issues

WHOIM1205:
> Let's not solve theoretical issues

sure

PR #2430, opened 2026-03-07
feat: Traceback task path from its origin  #448

### LINKED ISSUE #448: FR: Traceback task origin
It would be great if we could traceback the whole path for a given task (as opposed to just its origin). This could be useful to better debug crashes or even to map all task "routes" into a graph better visualise what's happening.

codedByJana:
# feat: Traceback task path from its origin  #448
“Closes #448” + API response example
#### the tracking process 

the issue was in ``` task.jinja2(31-36) ```: displaying only immediate parent, dispite the complete task JSON is stored in the database ``` db.py:(148-149) ``` , enabling full path reconstruction

#### To trace the complete path from root to current task:
- Added ``` get_task_path(self, task_uid: str) -> List[TaskResult]: ``` method to DB class to recursively fetch parent tasks
- Add ``` @router.get("/api/task/{task_uid}/path", dependencies=[Depends(verify_api_token)]) ``` endpoint to retrieve the full path with preserving the same pattern across other endpoints
- Modified the task template to display the full path
- Modify the task route handler to fetch and pass the path in ``` frontend.py ``` to render the task details page

#### Testing:
I observed the results of 3 sequential tasks, which two of them were run in parallel and noticed how the root_uid was the same, and the parent_uid was null in the first; now this was already there before the implementation,  the implementation successfully adds visibility to this existing tree structure without changing the underlying task creation logic.

Root Task
   ├── Task 2
   └── Task 3

``` 
-Result { "tag": null, "analysis_id": "1a590968-008e-4798-9a6d-8bd4e4ea4bf0", "created_at": "2026-03-07T01:04:30.842901", "target_string": "vulnweb.com", "task": { "uid": "4ac4ce0b-0edf-4c0f-8126-06d90900d820", "root_uid": "1a590968-008e-4798-9a6d-8bd4e4ea4bf0", "parent_uid": null, "orig_uid": "{1a590968-008e-4798-9a6d-8bd4e4ea4bf0}:1a590968-008e-4798-9a6d-8bd4e4ea4bf0", "priority": "normal", "last_update": 1772845470.528782, "payload": { "data": "vulnweb.com" }, 

-Result { "tag": null, "analysis_id": "1a590968-008e-4798-9a6d-8bd4e4ea4bf0", "created_at": "2026-03-07T01:04:35.167789", "target_string": "vulnweb.com", "task": { "uid": "ed6e092a-9819-4d82-816a-bc9add942c5a", "root_uid": "1a590968-008e-4798-9a6d-8bd4e4ea4bf0", "parent_uid": "4ac4ce0b-0edf-4c0f-8126-06d90900d820", "orig_uid": "{1c59f78a-7f96-49ff-8ae5-ff44499e88f9}:1c59f78a-7f96-49ff-8ae5-ff44499e88f9", "priority": "normal", "last_update": 1772845474.331498, "payload": { "created_at": "2026-03-07T01:04:30.829505", "domain": "vulnweb.com", "last_domain": "vulnweb.com" }, 

-Result { "tag": null, "analysis_id": "1a590968-008e-4798-9a6d-8bd4e4ea4bf0", "created_at": "2026-03-07T01:04:35.708719", "target_string": "vulnweb.com", "task": { "uid": "66356b10-05cb-4a4b-9cb0-1e4d0543dd5c", "root_uid": "1a590968-008e-4798-9a6d-8bd4e4ea4bf0", "parent_uid": "4ac4ce0b-0edf-4c0f-8126-06d90900d820", "orig_uid": "{1c59f78a-7f96-49ff-8ae5-ff44499e88f9}:1c59f78a-7f96-49ff-8ae5-ff44499e88f9", "priority": "normal", "last_update": 1772845473.0723968, "payload": { "created_at": "2026-03-07T01:04:30.829505", "domain": "vulnweb.com", "last_domain": "vulnweb.com" }, ```

### CODE REVIEW

File: artemis/api.py [hunk 1]
```diff
@@ -173,7 +174,13 @@ def get_task_results(
         analysis_id=analysis_id,
         task_filter=TaskFilter.INTERESTING if only_interesting else None,
     ).data
-
+#Add an endpoint to retrieve the full path
```
kazet: I don't think these comments are needed

templates/task.jinja2
```diff
@@ -31,23 +31,35 @@
...
         </dd>
     {% endif %}
+    
+  {#display the full path:#}
+  <h4>Full Task Path</h4>  
+  <ol class="breadcrumb">  
+  {% for ancestor in task_path %}  
+      <li class="breadcrumb-item">  
+          <a href="/task/{{ancestor.id}}">  
+              {{ancestor.task.headers.type}} - {{ancestor.target_string}}
```
kazet: could you also add a test for that?

artemis/api.py
kazet: I don't think these comments are needed

kazet:
could you fix the linter issues?

artemis/api.py
codedByJana: Tried to explain the steps of implementing along the codebase :) but i'll remove it

test/e2e/test_task_path.py
```diff
@@ -0,0 +1,26 @@
...
+        tasks = response.json()
+
+        if tasks:
+            task_uid = tasks[0]["task"]["uid"]
+
+            response = requests.get(f"{BACKEND_URL}task/{task_uid}")
+            self.assertEqual(response.status_code, 200)
+
+            self.assertIn("Full Task Path", response.text)
+            self.assertIn("breadcrumb", response.text)
```
kazet: I think the e2e test can test the correctness of displayed results more  ;)

templates/task.jinja2
```diff
@@ -31,23 +31,35 @@
...
         <dd class="col-9">
-            <a href="{{ request.url_for('get_task', task_id=task.parent_uid) }}">{{task.task.parent_uid}}</a>
+            <a href="/task/{{task.parent_uid}}">{{task.task.parent_uid}}</a>
         </dd>
     {% endif %}
 
+  {#display the full path:#}
+  <h4>Full Task Path</h4>
+  <ol class="breadcrumb">
+  {% for ancestor in task_path %}
```
kazet: I think the UX of this can be improved ;) can you display e.g. the modules that were executed for this task?

Image

BTW, for subdomain enumeration, the chain is not always full - could you make sure it always is?

test/e2e/test_task_path.py
codedByJana: ok I'll work on it

templates/task.jinja2
codedByJana: alright, on it
edit: so looking at the subdomain issue the cause of it lies in early returns; the module returns early for public suffixes or IP ranges, or already processed domains

so a proposed fix would be adding placeholder "skipped" tasks for early return scenarios.

I'll implement it and see how it would run

kazet: let me know if you need a second review

codedByJana: I pushed the changes for review

kazet: I still observe both issuses - the breadcrumbs can be expanded with e.g. module names and the chain is broken in case of subdomain enumeration

codedByJana: oh I now know why it was broken even when I added the "SKIPPED" status, By design the module never runs on any subdomain it discovered itself; the enumeration chain stops after one level.
The configuration comment for [SUBDOMAIN_ENUMERATION_TTL_DAYS] even says:

"if we requested crtsh enumeration on example.com and received www.example.com, crtsh enumeration on www.example.com won't happen "

In other words: the behaviour was intentional, not a mysterious bug, and that’s why the chain was broken

How I Fixed it:
```subdomain_enumeration.py```
Removed per‑subdomain Redis marks.
Added a single marker outside the tool loop.
Added explanatory comments.

Tests
Added ```test_recursive_enumeration``` proving recursion works.
Updated imports accordingly.

Comment (codedByJana) [reply templates/task.jinja2]: For the UX enhancements:
- added the module names (i don't know how did i forgot this i'm really sorry) so it now shows the receiver and origin module
- converted the Result, Metadata, and Logs sections into a Bootstrap accordion to improve space management and allow users to expand/collapse sections as needed, reducing clutter on long pages

File: artemis/modules/subdomain_enumeration.py
Comment (kazet) [inline artemis/modules/subdomain_enumeration.py:303]: If it gets removed, subdomain enumeration will re-run on all subdomains on the scanned domain, thus taking a long time. Do you have other solution in mind?

Comment (codedByJana) [reply artemis/modules/subdomain_enumeration.py:303]: Well it's actually a trade off, we can't cache discovered subdomains to avoid redundancy without breaking recursion, because we'd skip their enumeration. 

So we can have two cache keys, ```subdomain-enumeration-discovered-{domain}``` to prevent re-discovery from multiple sources, and ```subdomain-enumeration-enumerated-{domain}``` to prevent re-processing the same task

That way we can have optimal performance by default; Full recursion: each discovered subdomain becomes a task, Performance: no redundant enumeration from multiple sources

File: artemis/modules/subdomain_enumeration.py [hunk 5]
```diff
@@ -220,12 +194,19 @@ def get_subdomains_by_dns_brute_force(self, domain: str) -> Optional[Set[str]]:
...
 
     def run(self, current_task: Task) -> None:
-        if current_task.headers.get("origin", "") == self.identity:
-            return
+        # NOTE: We intentionally dont have an origin==self.identity check here anymore.
+        # That check blocked all recursive enumeration. Instead, we rely on the cache check
+        # below to prevent immediate re-processing while still allowing discovered subdomains
+        # to be enumerated later and spawn their own children.
 
         if has_ip_range(current_task):
```
kazet, line 267: feel free to save more task results but don't change the logic of subdomain_enumeration in this PR

codedByJana: OK, So in this PR I should just revert the commits regarding ```subdomain_enumeration.py``` and implement the solution in another PR?

kadewu:
Hello please make sure to properely resolve conflicts.
The tests will fail cause of current conflicts and as well old not properly handled.
Please remember to install pre-commit locally.

test/modules/test_subdomain_enumeration.py
```diff
@@ -41,3 +41,48 @@ def test_get_subdomains_from_subfinder(self) -> None:
...
+        # skipped and we expect to see the next level
+        subtask = Task(
+            {"type": TaskType.DOMAIN},
+            payload={TaskType.DOMAIN: "foo.example.com"},
+        )
+        self.karton.redis.flushall()  # clear the cache to isolate the test
+        subresults = self.run_task(subtask)
+        self.assertTrue(any(r.payload["domain"] == "bar.foo.example.com" for r in subresults))
+=======
+>>>>>>> parent of 68fd802 (fix broken subdomain enum chain)
```
kadewu: Just to make sure -> this should be cleared.

kadewu:
Please let me know once you think it's ready for re-review.

codedByJana:
I reverted the changes and ran pre-commit on all files. I think we just need to run the CI tests now.

kadewu:
Was there a need to revert all of the changes?
As @.kazet said, adding more task result was fine.

Same with test - I only did point to merge conflicts marks.

Wouldn't we go back with the behavior you were trying to fix this way?

kadewu:
Hello, do you still plan to work on that?

codedByJana:
i could still work on it but i really got mixed up about the many changes in this pr which to keep and which to make ahead to make this work, like it was an issue about the root tasks and making sure the tasks precendence are accurate and hierarchical, then got into subdomain module edits, if you can clear that up for me I'd get working on it as soon as possible

kadewu:
Please revert to some specific point in time, as currently there isn't much change in this PR.
I don't clearly remember what were the issues, so it would help me remember.

Edit: by specific I mean whichever you choose, then we will work from that point in time

PR #2431, opened 2026-03-07
fix: prevent autoreporter crash loop on poison-pill export tasks

WHOIM1205:
### Problem
`take_single_report_generation_task()` selects tasks with status `pending` but does not update their status before returning them.

If the autoreporter process crashes while processing a task for example due to OOM kill container restart or SIGKILL the task remains `pending` in the database. When the container restarts it repeatedly picks the same task again.

In cases where the task itself causes the crash such as a very large export this results in an infinite crash loop preventing any other report generation tasks from being processed.

### How the issue was encountered
While testing exports I generated a large export for a tag with many results. During report generation the autoreporter process was killed due to memory usage. After the container restarted the same task was still marked as pending so it was picked again. This caused the worker to repeatedly restart while trying to process the same task.

### Steps to reproduce
1 Start Artemis using docker compose  
2 Run scans that generate a large number of results under one tag  
3 Request an export for that tag  
4 The autoreporter begins building the export data  
5 Memory usage grows and the worker process gets killed  
6 After restart the same task is picked again because it remained pending  
7 The autoreporter enters a restart loop

### Root Cause
The task lifecycle had no intermediate state between `pending` and terminal states `done` and `failed`.  
Tasks were read but never claimed which made the operation non atomic.

### Solution
This PR introduces an `in_progress` state to represent tasks currently being processed.

Changes
- Added `IN_PROGRESS` to `ReportGenerationTaskStatus`
- Updated `take_single_report_generation_task()` to mark tasks as `in_progress` before returning them
- Added `recover_stuck_in_progress_tasks()` to mark tasks as `failed` if the worker previously crashed
- Called the recovery step on autoreporter startup
- Updated the exports UI to display the new `in_progress` status

### Result
- Prevents infinite crash loops caused by poison pill export tasks
- Ensures crashed tasks are automatically marked as failed
- Allows the autoreporter to continue processing other pending tasks

### DIFF
--- a/artemis/reporting/task_handler.py
+++ b/artemis/reporting/task_handler.py
@@ -10,7 +10,6 @@ import traceback
 from pathlib import Path
 
 import psutil
-from sqlalchemy.orm.exc import ObjectDeletedError
 
 from artemis import utils
 from artemis.config import Config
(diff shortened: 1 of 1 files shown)


WHOIM1205:
heyy @kazet 
This PR fixes an issue where report generation tasks remained pending if the worker crashed during processing for example due to an OOM kill or container restart. Because the task was never claimed the autoreporter could repeatedly pick the same task after restart which may lead to an infinite crash loop.

This change adds an in progress state so tasks are marked when they are picked up and also adds a startup recovery step that marks previously in progress tasks as failed if the worker crashed.

kazet (changes requested):
I see one thing missing - if we legitimately restart autoreporter the task will also be marked as crashed.

WHOIM1205:
hey @kazet 
Thanks for pointing that out
I updated the recovery logic so that in progress tasks are only marked as failed if they exceed a timeout. Each task now records when processing starts and recovery only applies to tasks that have been in progress for longer than the configured timeout. This avoids marking tasks as failed during normal restarts while still recovering tasks left behind by worker crashes.

WHOIM1205:
hey @kazet is there anything i can change

kadewu:
https://github.com/CERT-Polska/Artemis/pull/2381#issuecomment-4011232551

I would assume you did you encounter that error since the last PR?
Please update the PR description with the steps.

WHOIM1205:
> [#2381 (comment)](https://github.com/CERT-Polska/Artemis/pull/2381#issuecomment-4011232551)
> 
> I would assume you did you encounter that error since the last PR? Please update the PR description with the steps.

updated

WHOIM1205:
hey @kazet is there anything i can change

kadewu:
> Steps to reproduce
> 
> 1 Start Artemis using docker compose
> 2 Run scans that generate a large number of results under one tag
> 3 Request an export for that tag
> 4 The autoreporter begins building the export data
> 5 Memory usage grows and the worker process gets killed
> 6 After restart the same task is picked again because it remained pending
> 7 The autoreporter enters a restart loop

How much is `large`? I'm not able to reproduce it.

I would rather be concerned that we have memory leak, and would be good to see why. Unfortunately I'm not able to reproduce memory leak as well.
How much memory did you hit in the peak?

WHOIM1205:
> > Steps to reproduce
> > 1 Start Artemis using docker compose
> > 2 Run scans that generate a large number of results under one tag
> > 3 Request an export for that tag
> > 4 The autoreporter begins building the export data
> > 5 Memory usage grows and the worker process gets killed
> > 6 After restart the same task is picked again because it remained pending
> > 7 The autoreporter enters a restart loop
> 
> How much is `large`? I'm not able to reproduce it.
> 
> I would rather be concerned that we have memory leak, and would be good to see why. Unfortunately I'm not able to reproduce memory leak as well. How much memory did you hit in the peak?

in my case the tag had a few thousand results i do not have the exact number anymore unfortunately while testing i noticed memory usage growing during export generation and eventually the container got killed i did not record the exact peak but the container limit in my setup is around a few hundred MB so it likely crossed that looking at the export code it seems the results are collected and grouped in memory before rendering the templates so the memory usage probably scales with the number of results 
my pr was mainly addressing the situation where the worker dies and the task remains pending which then causes the restart loop

kazet:
If we have problem with exporter memory usage, let's focus on that (e.g. build results incrementally) instead of skipping tasks.

PR #2432, opened 2026-03-07
fix: prevent silent death of report generation main loop

WHOIM1205:
## Fix autoreporter main loop crash on unhandled exceptions

The `main()` loop in `artemis/reporting/task_handler.py` had no top-level exception handling.  
If `db.take_single_report_generation_task()` or `save_report_generation_task_results()` raised an exception (for example a transient DB error), the main thread could exit while the API thread kept the process alive.

This resulted in the autoreporter container appearing healthy while report generation silently stopped processing tasks.

### Changes
- Wrapped the `while True` loop body in `try/except` so transient errors are logged and the loop continues.
- Broadened the inner error handler to catch all exceptions when saving failure status, preventing the loop from crashing.

This makes the autoreporter more resilient to temporary database or infrastructure issues.

### DIFF
--- a/artemis/reporting/task_handler.py
+++ b/artemis/reporting/task_handler.py
@@ -10,6 +10,7 @@ import traceback
 from pathlib import Path
 
 import psutil
+
 from artemis import utils
 from artemis.config import Config
 from artemis.db import DB, ReportGenerationTask, ReportGenerationTaskStatus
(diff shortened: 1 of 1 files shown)


WHOIM1205:
heyy @kazet 
This PR addresses the issue where the autoreporter main loop could stop running if an unexpected exception occurred during task processing or database access. The loop is now protected so it logs the error and continues running.

PR #2438, opened 2026-03-08
fix: use asyncio.to_thread for blocking I/O in async frontend endpoints

WHOIM1205:
### Title
Fix blocking I/O in async frontend export endpoints

### Summary
Two `async def` endpoints in `artemis/frontend.py` were calling synchronous database and filesystem operations directly. Since these handlers run on the Uvicorn event loop thread, the blocking calls could stall the entire server until the operation completed.

### Problem
The following endpoints executed blocking functions without offloading them to a thread:

- `post_export_delete`
- `post_export`

Both call synchronous database helpers (`delete_report_generation_task`, `create_report_generation_task`).  
`delete_report_generation_task` additionally performs filesystem I/O (`shutil.rmtree`), which can take noticeable time for large report directories.

When deployed with a single Uvicorn worker, these operations block the event loop and temporarily prevent other requests from being processed.

### Fix
Wrap the blocking calls with `await asyncio.to_thread()` so they run in the thread pool instead of blocking the event loop.

```python
await asyncio.to_thread(db.delete_report_generation_task, id)

### DIFF
--- a/artemis/frontend.py
+++ b/artemis/frontend.py
@@ -271,7 +271,7 @@ def view_export(request: Request, id: int) -> Response:
 @router.post("/export/confirm-delete/{id}", include_in_schema=False)
 @csrf.validate_csrf
 async def post_export_delete(request: Request, id: int, csrf_protect: CsrfProtect = Depends()) -> Response:
-    db.delete_report_generation_task(id)
+    await asyncio.to_thread(db.delete_report_generation_task, id)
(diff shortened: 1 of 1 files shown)


WHOIM1205:
Hi @kazet  I noticed that two async endpoints in frontend are calling blocking database and filesystem operations directly which can block the event loop I opened this PR to move those calls to asyncio to_thread so the server remains responsive

PR #2439, opened 2026-03-08
fix: make Jinja2 environment thread-safe in autoreporter

WHOIM1205:
### Problem
The autoreporter runs two threads in the same Python process

- API server thread
- Report generation thread

Both threads were mutating a shared global Jinja2 `Environment` defined in `reporting/export/main.py`.

`install_gettext_translations()` modifies the environment by replacing the `gettext` functions. When both threads install translations concurrently the environment state can change during template rendering.

This can cause:

- mixed language reports
- `TranslationNotFoundException` crashes during render
- silent report corruption sent to recipients

### Root Cause
A module level global `Environment` object was shared across threads without synchronization.

The sequence:

was not atomic, allowing another thread to overwrite the translation state mid render.

### Fix
Remove the shared global environment and replace it with a per invocation environment.

Key changes:

- Introduced `create_environment()` factory
- Each caller now creates its own `Environment`
- Passed the environment instance through:
  - `install_translations_and_print_path`
  - `build_message_template_and_print_path`
- Updated both call sites:
  - `export()` in report generation
  - `post_build_html_message()` API endpoint

### Result
Each report generation or API request now uses an isolated Jinja2 environment instance.

This removes shared mutable state and prevents cross thread translation corruption.

### Impact
- prevents mixed language report output
- prevents translation class swap crashes
- safe concurrent API preview requests during report generation
- no architectural changes required

### DIFF
--- a/artemis/reporting/export/main.py
+++ b/artemis/reporting/export/main.py
@@ -103,7 +103,9 @@ def build_message_template_and_print_path(environment: Environment, output_dir: 
     return message_template
 
 
-def install_translations_and_print_path(language: Language, environment: Environment, output_dir: Path, silent: bool) -> None:
+def install_translations_and_print_path(
+    language: Language, environment: Environment, output_dir: Path, silent: bool
+) -> None:
     translations_file_name = output_dir / "advanced" / "translations.po"
     compiled_translations_file_name = output_dir / "advanced" / "compiled_translations.mo"
     install_translations(language, environment, translations_file_name, compiled_translations_file_name)


WHOIM1205:
Hi @kazet this PR fixes a thread safety issue caused by a shared global Jinja2 environment
The API thread and report generation thread could install translations on the same environment at the same time which could lead to mixed language output or template rendering errors
This change creates a new environment per invocation so each thread uses its own isolated instance

WHOIM1205:
Hey @kazet is there anything i can change in this pr

PR #2461, opened 2026-03-09
Fix API scanner skipping large OpenAPI specs due to response truncation

WHOIM1205:
### Problem
`APIScanner.discover_spec()` fetched OpenAPI/Swagger specs using `http_requests.get()` without overriding the default response size limit.

The underlying `http_requests.request()` truncates responses to `CONTENT_PREFIX_SIZE` (100KB). Many real-world API specifications exceed this size. When a spec larger than 100KB is downloaded, it gets truncated, resulting in invalid JSON/YAML that fails parsing. The exception is silently caught, causing the scanner to incorrectly report that no API specification was found and skip the scan.

### Root Cause
`http_requests.get()` was called without specifying `max_size`, causing large OpenAPI specifications to be truncated.

### Fix
Add a configurable maximum download size for API specifications and pass it to `http_requests.get()`.

### Changes
- Added `API_SPEC_MAX_SIZE` configuration under `Config.Modules.APIScanner` (default **5MB**).
- Updated `api_scanner.py` to pass `max_size=Config.Modules.APIScanner.API_SPEC_MAX_SIZE` when downloading specs.

### Impact
- Prevents truncation of OpenAPI/Swagger specifications.
- Ensures large real-world API specs can be parsed and scanned.
- Allows the OFFAT integration to run correctly on APIs with complex specifications.
- Eliminates false negatives where APIs were previously skipped.

### Files Modified
- `artemis/modules/api_scanner.py`
- `artemis/config.py`

### DIFF
--- a/artemis/config.py
+++ b/artemis/config.py
@@ -330,6 +330,12 @@ class Config:
             ] = get_config("ADMIN_PANEL_LOGIN_BRUTER_NUM_RECHECKS", default=10, cast=int)
 
         class APIScanner:
+            API_SPEC_MAX_SIZE: Annotated[
+                int,
+                "Maximum size in bytes for downloading OpenAPI/Swagger specification files. "
+                "The default CONTENT_PREFIX_SIZE (100KB) is too small for most real-world API specs.",
+            ] = get_config("API_SPEC_MAX_SIZE", default=5 * 1024 * 1024, cast=int)
+
             ONLY_GET_REQUESTS: Annotated[
                 bool,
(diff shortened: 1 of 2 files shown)


WHOIM1205:
hey @kazet 
This change fixes an issue where the API scanner truncates OpenAPI specifications larger than 100KB due to the default response size limit The scanner fails to parse valid specifications and incorrectly reports that no specification was found The fix increases the allowed download size for API specs so larger real world specifications can be parsed and scanned correctly

kazet (approved):

PR #2462, opened 2026-03-09
fix: handle naabu subprocess failure in port scanner

WHOIM1205:
### Summary
This PR fixes an issue where failures from the `naabu` subprocess in `PortScanner._scan()` were silently ignored, causing the scanning pipeline to receive zero discovered services without any error indication.

### Problem
Two issues caused silent failures:

1. `stderr` was not captured in `subprocess.Popen()`, making the stderr logging block ineffective.
2. The subprocess `returncode` was never checked, so non-zero exits were treated as successful scans with no open ports.

As a result, if `naabu` failed (e.g., permission issues, missing binary, resource errors), the scanner would proceed with empty results and downstream modules would never receive `SERVICE` tasks.

### Changes
- Added `stderr=subprocess.PIPE` to capture diagnostic output.
- Logged stderr output at warning level.
- Added a check for `naabu.returncode` and raise `subprocess.CalledProcessError` when it is non-zero.

### Impact
- Prevents silent scan failures.
- Ensures port scan errors are surfaced in logs.
- Avoids false clean results when the port scanner fails.
- Improves reliability of downstream vulnerability detection modules.

### Files Modified
- `artemis/modules/port_scanner.py`

### DIFF
--- a/artemis/modules/port_scanner.py
+++ b/artemis/modules/port_scanner.py
@@ -146,7 +146,9 @@ class PortScanner(ArtemisBase):
             # communicate() to avoid that.
             stdout, stderr = naabu.communicate()
             if stderr:
-                self.log.warning(f"naabu returned the following stderr content: {stderr.decode('utf-8', errors='ignore')}")
+                self.log.warning(
+                    f"naabu returned the following stderr content: {stderr.decode('utf-8', errors='ignore')}"
+                )
 
             if naabu.returncode != 0:
                 self.log.error(f"naabu exited with code {naabu.returncode} for targets {new_target_ips}")


WHOIM1205:
hey @kazet 
naabu stderr is not captured and the return code is not checked so if the process fails the scanner silently continues with empty results this change captures stderr and checks the return code so failures are visible and do not appear as successful scans with no open ports

artemis/modules/port_scanner.py
```diff
@@ -145,7 +146,11 @@ def _scan(self, target_ips: List[str]) -> Dict[str, Dict[str, Dict[str, Any]]]:
             # communicate() to avoid that.
             stdout, stderr = naabu.communicate()
             if stderr:
-                self.log.info(f"naabu returned the following stderr content: {stderr.decode('utf-8', errors='ignore')}")
+                self.log.warning(f"naabu returned the following stderr content: {stderr.decode('utf-8', errors='ignore')}")
+
+            if naabu.returncode != 0:
+                self.log.error(f"naabu exited with code {naabu.returncode} for targets {new_target_ips}")
```
kazet, line 154: could you write an integration test for that?

WHOIM1205:
hey @kazet 
I added an integration test that mocks a failing naabu process and verifies that a non zero return code raises CalledProcessError instead of silently returning empty results

tests/modules/test_port_scanner.py
```diff
@@ -0,0 +1,26 @@
+import subprocess
+from unittest.mock import MagicMock
+
+import pytest
+
+from artemis.modules.port_scanner import PortScanner
+
+
+def test_scan_raises_on_naabu_failure(monkeypatch):
```
kazet: Let's make the used test libraries consistent with other tests ;) BTW, pls fix the linter issues

WHOIM1205:
hey @kazet 
Thanks for the suggestion I updated the test to match the existing test framework and fixed the linter issues

PR #2467, opened 2026-03-10
fix: log warning instead of silently swallowing exception in wordpress_bruter.py line 35

vishnudathks:
## Problem
In `artemis/modules/wordpress_bruter.py` line 35, a bare 
`except Exception: pass` silently swallows all errors during 
WordPress user enumeration with no logging.

## Change
Before:
    except Exception:
        pass

After:
    except Exception:
        self.log.warning("Failed to enumerate WordPress users from %s", url)

## Impact
- Failed user enumeration now produces a warning log entry
- Easier to debug network errors and blocked API responses
- Same pattern as the already merged fix in bruter.py

### DIFF
--- a/artemis/modules/wordpress_bruter.py
+++ b/artemis/modules/wordpress_bruter.py
@@ -32,7 +32,7 @@ class WordPressBruter(ArtemisBase):
             for user_entry in users:
                 usernames.append(user_entry["name"])
         except Exception:
-              self.log.exception("Failed to enumerate WordPress users from %s", url)
+            self.log.exception("Failed to enumerate WordPress users from %s", url)
         usernames += ["admin", "administrator", "wordpress"]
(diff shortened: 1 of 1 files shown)


artemis/modules/wordpress_bruter.py
```diff
@@ -32,7 +32,7 @@ def run(self, current_task: Task) -> None:
             for user_entry in users:
                 usernames.append(user_entry["name"])
         except Exception:
-            pass
+             self.log.warning("Failed to enumerate WordPress users from %s", url)
```
kazet:
```suggestion
             self.log.exception("Failed to enumerate WordPress users from %s", url)
```

What do you think?

vishnudathks: @kazet Great suggestion! Changed to `log.exception()` to capture the 
full stack trace for better debugging.

PR #2470, opened 2026-03-10
add master-addons to PLUGINS_BAD_VERSION_IN_README to stabilize flaky test`

vishnudathks:
Master-addons readme version doesn't match the actual plugin version,
causing test_plugin_identification_from_readme to fail in CI.

Adding it to PLUGINS_BAD_VERSION_IN_README to fix the flaky test,
consistent with how other such plugins are handled.


kazet:

Please fix the whitespace issues reported by linter.

BTW, this issue doesn't fix the flaky test, we have some more that fail probabilistically, so the "Fixes"  word in the description needs to be changed.

PR #2471, opened 2026-03-10
fix(nuclei): prevent batch crash when finding cannot be matched to a task

WHOIM1205:
### Problem

`run_multiple()` used a Python `assert` to ensure that every nuclei finding could be matched back to a task:

```python
assert found, "Cannot match finding: %s" % finding
```

If nuclei returned a finding whose `host` did not match any task target (e.g., CDN IPs, redirects, or domain/IP mismatches), the assertion would raise `AssertionError` and crash the entire batch. Because nuclei runs with `batch_tasks=True`, this caused **all scan results in the batch to be lost**.

Additionally, accessing `finding["host"]` could raise a `KeyError` when the nuclei JSON output lacks the `host` field.

### Fix

1. Use `finding.get("host", "")` to avoid `KeyError`.
2. Replace the `assert` with a warning and skip the unmatched finding.

```python
if finding.get("host", "").split(":")[0] == get_target_host(task).split(":")[0]:
```

```python
if not found:
    self.log.warning("Cannot match finding to any task, skipping: %s", finding)
```

### Impact

* Prevents `run_multiple()` from crashing when nuclei outputs an unmatched finding.
* Preserves results for all other tasks in the batch.
* Logs unmatched findings for debugging instead of terminating the scan.

### DIFF
--- a/artemis/modules/nuclei.py
+++ b/artemis/modules/nuclei.py
@@ -693,8 +693,7 @@ class Nuclei(ArtemisBase):
                         findings_per_task[task.uid].append(finding)
                         found = True
                         break
-                if not found:
-                    self.log.warning("Cannot match finding to any task, skipping: %s", finding)
+                assert found, "Cannot match finding: %s" % finding
 
         for task in tasks:
             result = []


WHOIM1205:
hey @kazet 
i noticed that run multiple uses an assert when matching nuclei findings to tasks If a finding host does not match any task the assertion crashes the whole batch and all results are lost This change replaces the assert with safe handling and logs the unmatched finding so the scan can continue

artemis/modules/nuclei.py
```diff
@@ -689,11 +689,12 @@ def run_multiple(self, tasks: List[Task]) -> None:
             for finding in findings_unmatched:
                 found = False
                 for task in tasks:
-                    if finding["host"].split(":")[0] == get_target_host(task).split(":")[0]:
+                    if finding.get("host", "").split(":")[0] == get_target_host(task).split(":")[0]:
                         findings_per_task[task.uid].append(finding)
                         found = True
                         break
-                assert found, "Cannot match finding: %s" % finding
```
kazet, line 696: The assertion is required, a finding that we weren't able to be matched is a condition that warrants a module failure so that an operator looks into this

WHOIM1205:
hey @kazet 
Thanks for the clarification I restored the assertion and kept the safe host access using finding get host to avoid a potential KeyError

PR #2472, opened 2026-03-10
Fix TypeError in dangling_dns_detector whois error handling

WHOIM1205:
### Problem
`check_for_public_entities` incorrectly performs a containment check using `token in e`, where `e` is an Exception object.  
Since exceptions are not iterable, this raises:

TypeError: argument of type 'HTTPLookupError' is not iterable

This occurs when `IPWhois.lookup_whois()` throws `HTTPLookupError` (commonly due to WHOIS rate limiting).

Additionally, if the condition failed, execution would fall through and reference `whois_data`, which was never assigned, causing an `UnboundLocalError`.

### Fix
- Convert the exception to string before performing the containment check (`token in str(e)`).
- Add a fallback branch to log non-rate-limit errors.
- Ensure the function safely returns `False` for all `HTTPLookupError` cases to prevent fallthrough.

### Impact
- Prevents scanner crashes caused by WHOIS rate limiting or other lookup failures.
- Ensures dangling DNS detection continues running instead of terminating the module.

### DIFF
--- a/artemis/modules/dangling_dns_detector.py
+++ b/artemis/modules/dangling_dns_detector.py
@@ -360,7 +360,7 @@ class DanglingDnsDetector(ArtemisBase):
             whois_data = IPWhois(ip).lookup_whois()
         except HTTPLookupError as e:
             if any(
-                token in e
+                token in str(e)
                 for token in (
                     "429",
                     "too many requests",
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
the issue occurs because token in e checks containment on an exception object which raises a TypeError during HTTPLookupError handling converting the exception to string and returning False for all error paths prevents the crash and avoids using uninitialized whois_data

PR #2473, opened 2026-03-10
Add redirect detection for successful login attempts in admin_panel_login_bruter module

imraflip:
This PR adds redirect-based success detection to `admin_panel_login_bruter` module.

Many admin panels redirect users to another page (e.g. a dashboard) after a successful login instead of returning a different response body. Previously, the module relied mainly on response content indicators such as logout links or failure messages, which could miss successful logins when authentication was implemented using HTTP redirects.

The new logic detects redirects by comparing the path of the response URL (or the Location header) with the original login path. If the request results in a redirect to a different path, it is treated as a potential successful login indicator.

Redirect detection is also performed against the actual form action URL, which improves accuracy for applications where the login form posts to a different endpoint than the login page itself.

This improves detection reliability for admin panels that use redirect-based authentication flows.

#2378


kazet:
Thank you for the contribution! Could you add a test for that?

imraflip:
You're welcome :)
I've already added it in `test_simple`, the assertion verifies that '`redirect`' is in the `indicators`. But do you think that is not sufficient to test it?

kazet:
I meant a test against a real app :)

imraflip:
Aha I see what you mean now. I was thinking that since `test-php-easy-admin-password` already does a redirect after login, the existing test would cover it, but that app also has session cookies and a logout link, so the redirect indicator was never the only reason the login got flagged as successful

I'll add a dedicated test app that only signals success via redirect so we can properly verify the redirect detection in isolation

imraflip:
I've added `test_redirect_login` with a test app that only signals success via redirect. This ensures redirect detection is explicitly tested and not masked by other indicators

That said, I still feel this is somewhat redundant, since `php_easy_admin_password` already redirects on successful login, so `test_simple` already tests redirect detection in practice. The only difference is that `test_simple` could still pass via the `logout_link` fallback if redirect detection broke

PR #2478, opened 2026-03-11
Fix tee process killed before flushing output buffer

WHOIM1205:
## Summary
Fix an issue where OutputRedirector.__exit__ terminated the tee process immediately after closing stdin, which could truncate the final part of task output logs.

## Changes
- Removed the immediate kill of the tee process.
- Wait for tee to exit naturally after receiving EOF on stdin.
- Added a timeout based fallback to kill the process only if it does not exit.

## Why
When writing to a file, tee uses buffered I/O. Killing the process before it flushes its buffers can silently drop the final portion of logs.

## Result
Ensures that all task output is properly flushed and written to the log file while still preventing potential hangs.

### DIFF
--- a/artemis/output_redirector.py
+++ b/artemis/output_redirector.py
@@ -33,8 +33,11 @@ class OutputRedirector:
             self._stream_copy[fd].close()
         if self._tee.stdin:
             self._tee.stdin.close()
-        self._tee.kill()
-        self._tee.wait()
+        try:
+            self._tee.wait(timeout=5)
+        except subprocess.TimeoutExpired:
+            self._tee.kill()
+            self._tee.wait()
 
     def get_output(self) -> bytes:
         self._output_collector_file.seek(0)


WHOIM1205:
hey @kazet 
I found that OutputRedirector exit kills tee right after stdin close which may stop tee before it flushes buffered writes and this can truncate the end of task logs

kazet:
How did you cause this problem to occur in real Artemis usage?

WHOIM1205:
> How did you cause this problem to occur in real Artemis usage?

After reviewing the actual code path more carefully I realized the implementation already waits up to 5 seconds for tee to exit after stdin is closed. With stdbuf o0 and the wait timeout the truncation scenario I described earlier is extremely unlikely in normal Artemis usage. The only case where kill would trigger is if tee is blocked on a filesystem write for more than the timeout which would be a pathological IO situation. Thanks for pointing this out.

PR #2479, opened 2026-03-11
Fix response time precision loss in sql_injection_detector measure_request_time

WHOIM1205:
### Issue
In `sql_injection_detector.py` (line 109), the request timing logic uses `datetime.timedelta(...).seconds`, which truncates the measured response time to integer seconds.

### Impact
This removes sub-second precision required for reliable **time-based SQL injection detection**.  
With the default threshold of **5 seconds**, response delays between **4.0–4.99 seconds** are truncated to `4`, causing the `>= 5` comparison to fail and potentially missing valid injection delays under normal network jitter.

### Fix
Replace `.seconds` with `.total_seconds()` to preserve full timing precision.

### Result
Time-based SQL injection detection now uses accurate response timing and avoids silently missing vulnerabilities due to integer truncation.

### DIFF
--- a/artemis/modules/sql_injection_detector.py
+++ b/artemis/modules/sql_injection_detector.py
@@ -1,4 +1,4 @@
-import datetime
+
 import random
 import re
 import urllib
@@ -106,7 +106,7 @@ class SqlInjectionDetector(ArtemisBase):
         except requests.exceptions.Timeout:
             return Config.Modules.SqlInjectionDetector.SQL_INJECTION_TIME_THRESHOLD
 
-        return datetime.timedelta(seconds=timer() - start).seconds
+        return timer() - start
 
     def contains_error(self, url: str, response: Optional[HTTPResponse]) -> str | None:
         if response is None:


WHOIM1205:
heyy @kazet 
sql injection detector uses timedelta seconds which truncates sub second response time this can cause time based sql injection delays below the threshold to be missed using total seconds will preserve timing precision

kazet:
SQL injection times below the threshold should be treated as no vulnerability.

If you ask the db to sleep(5) and it sleeps 4.9, it's a FP not a time-based SQLi, isn't it?

WHOIM1205:
> SQL injection times below the threshold should be treated as no vulnerability.
> 
> If you ask the db to sleep(5) and it sleeps 4.9, it's a FP not a time-based SQLi, isn't it?

I understand the threshold logic. My point was about precision loss in measurement. timedelta seconds truncates the value so a response like 4.9 becomes 4. Using total seconds would keep the real measured time and avoid that truncation

PR #2480, opened 2026-03-11
Fix silent vulnerability finding loss in detection pipeline

WHOIM1205:
### Bug Summary
Certain modules could silently drop valid vulnerability findings due to improper handling of unexpected responses or missing validation during the finding attribution and processing pipeline. In some cases, results returned from scanners or helper functions were not correctly propagated, causing legitimate detections to be ignored.

### Fix
Added proper validation and handling to ensure findings are consistently propagated through the detection pipeline. This prevents valid vulnerability results from being silently discarded and ensures they reach the reporting layer.


WHOIM1205:
hey @kazet 
This fix ensures valid vulnerability findings are not silently dropped during processing

PR #2481, opened 2026-03-11
Fix fd corruption in OutputRedirector when SIGALRM interrupts cleanup

WHOIM1205:
### Summary
Fix a signal interruption race condition in `OutputRedirector.__exit__()` that can corrupt `stdout`/`stderr` file descriptors when `timeout_decorator` raises `TimeoutError` via `SIGALRM` during cleanup.

### Problem
If `SIGALRM` fires while `__exit__()` is restoring file descriptors using `os.dup2`, the cleanup sequence can be interrupted between restoring `stdout` and `stderr`. This leaves `stderr` pointing to the dead `tee` pipe and prevents the remaining cleanup steps from running.

This leads to:
- zombie `tee` processes
- broken `stderr` causing lost logs or `BrokenPipeError`
- file descriptor leaks across retries
- eventual worker crashes with `Too many open files`

### Fix
Temporarily block `SIGALRM` during the critical cleanup section of `OutputRedirector.__exit__()` so the file descriptor restoration and process cleanup complete atomically.

The original signal handler is restored after cleanup.

### Impact
- prevents stdout/stderr corruption
- prevents zombie tee processes
- avoids fd leaks during retries
- ensures logs are preserved for timed out tasks

### DIFF
--- a/artemis/output_redirector.py
+++ b/artemis/output_redirector.py
@@ -1,4 +1,5 @@
 import os
+import signal
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
This fixes a race where timeout signal interrupts OutputRedirector cleanup and leaves stderr pointing to a dead pipe which can break logs and leak file descriptors

kazet:
How did you cause this problem to occur in real Artemis usage?

WHOIM1205:
> How did you cause this problem to occur in real Artemis usage?

From what I saw in the flow the timeout decorator clears the timer in its finally before exit runs so the window is small. But exit restores stdout and stderr one after the other and if a signal happens to arrive between those steps the cleanup could stop midway and stderr may still point to the tee pipe. This change just makes that small cleanup section safe so the descriptors always get restored.

PR #2483, opened 2026-03-12
Implement parameter minimization to SqlInjectionDetector

### LINKED ISSUE #2410: Add parameter minimization to sql_injection_detector/lfi_detector

imraflip:
Closes #2410 

This PR implements parameter minimization for generated SQLi PoCs.

Before the change, when SQL injection (error-based or time-based) was detected, the PoC URL contained the full injected parameter batch, which often made it unnecessarily long and noisy. After detection, the module now re-tests each parameter individually and keeps only those that independently reproduce the issue. If none of them works alone, it falls back to the original batch to avoid changing detection behavior.

The same minimization approach is applied in both query-parameter handling paths (URLs with existing parameters and URLs without them), so reported PoCs stay focused while preserving current detection logic.

The tests were updated accordingly to validate behavior.

### DIFF
--- a/artemis/modules/sql_injection_detector.py
+++ b/artemis/modules/sql_injection_detector.py
@@ -137,7 +137,8 @@ class SqlInjectionDetector(ArtemisBase):
     ) -> List[str]:
         """
         Try to find the minimal set of parameters that still triggers SQLi. Currently minimizes to single parameters only.
-        Falls back to original params if none work individually.
+        Falls back to original params if none work individually. When minimized parameters are found,
+        the result is capped to SQL_INJECTION_MINIMAL_PARAMS_MAX_LEN.
         """
         if minimization_mode == "error" and baseline_payload is None:
             raise ValueError("baseline_payload is required for error-based minimization")
@@ -175,14 +176,17 @@ class SqlInjectionDetector(ArtemisBase):
                 minimal_params.append(param)
 
         if minimal_params:
+            capped_minimal_params = minimal_params[
+                : Config.Modules.SqlInjectionDetector.SQL_INJECTION_MINIMAL_PARAMS_MAX_LEN
+            ]
             mode_label = "error-based" if minimization_mode == "error" else "time-based"
             self.log.info(
                 "SQLi %s parameter minimization: %s -> %s",
                 mode_label,
                 params,
-                minimal_params,
+                capped_minimal_params,
             )
-            return minimal_params
+            return capped_minimal_params
 
         # fallback if no single param triggers SQLi
         return params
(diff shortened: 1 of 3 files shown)


artemis/modules/sql_injection_detector.py
```diff
@@ -119,6 +119,81 @@ def contains_error(self, url: str, response: Optional[HTTPResponse]) -> str | No
                 return message
         return None
 
+    def minimize_parameters_error_based(
```
kadewu: Did you look into refactoring this: extracting the common part?
Those two methods looks similar (and similar to LFI one already implemented)

imraflip: Actually, I didn't think of it. When implementing the LFI, there was only one flow, so one method was enough, unlike in SQLi, which has error-based and time-based, so I split them
But sure, let me refactor and extract the common logic into one method to keep it clean ;)

imraflip: Hello, I just pushed a refactor that merges parameter minimization into one shared method to remove the duplicated logic. I also reran the tests, and everything is still passing as before

artemis/modules/sql_injection_detector.py
```diff
@@ -119,6 +119,74 @@ def contains_error(self, url: str, response: Optional[HTTPResponse]) -> str | No
...
+        baseline_payload: Optional[str] = None,
+    ) -> List[str]:
+        """
+        Try to find the minimal set of parameters that still triggers SQLi. Currently minimizes to single parameters only.
+        Falls back to original params if none work individually.
+        """
+        if minimization_mode == "error" and baseline_payload is None:
+            raise ValueError("baseline_payload is required for error-based minimization")
+
+        minimal_params: List[str] = []
```
kadewu, line 146: Actually there will be one more thing from me. After some consideration I believe it would be great to add a limit on how that list is long. The reason is that some entities will trigger for every parameter and we will end up with a long url again - which we wanted to reduce in first place. On other hand sending just want may be insufficient; entities may just try to fix that one param and we will end up ping-ponging.

So making a limit on how long the minimal set of parameter can be, would compromise.
I would suggest that this is configurable (`artemis/config.py`) , default to 5.

If you have any questions regarding that feel free to ask.

imraflip: I think I get what you mean. If we find vulnerable parameters like a, b, c, d, e, f, g, h, i, j, k, l, m, n, sending all 14 can still make the URL quite long, but sending only one param might not be enough. So with a configurable limit (default 5), we could send a, b, c, d, e as a middle ground. Is that correct?

kadewu: Yes, exactly.

imraflip: I’ve just pushed the update, you may review it again now ;)

PR #2485, opened 2026-03-13
add delay to @retry in test_simple to stabilize WordPress bruter test

vishnudathks:
Fixes part of #2465.
test_simple in test_wordpress_bruter.py already had @retry(tries=3) but with no delay, all three retries fired instantly — before the test-wordpress-easy-password Docker container had finished starting. Adding delay=10 gives the container 10 seconds between retries to become ready.


kazet:
Hello,

thanks for the contribution! Could you remove the other change (randint mocking) that we discussed is a bad idea?

vishnudathks:
ok, will you check the discord i have few questions.

PR #2490, opened 2026-03-14
add retry to test_simple in domain expiration scanner to handle WHOIS flakiness

vishnudathks:
Fixes part of #2465.
test_simple in test_domain_expiration_scanner.py calls perform_whois_or_sleep which makes a live external WHOIS request for google.com with no retry logic. WHOIS servers can be slow or rate-limit requests causing random test failures. Adding @retry(tries=3, delay=10) gives the test three attempts with a 10-second gap between them.

### DIFF
--- a/test/modules/test_domain_expiration_scanner.py
+++ b/test/modules/test_domain_expiration_scanner.py
@@ -3,6 +3,7 @@ from test.base import ArtemisModuleTestCase
 from unittest.mock import patch
 
 from karton.core import Task
+from retry import retry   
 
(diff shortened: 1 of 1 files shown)


kazet:
did you observer flakiness of this test on real CI runs?

vishnudathks:
I haven't observed it in CI logs directly, but WHOIS servers are known to rate-limit . happy to revert if you feel it's unnecessary.

kadewu:
Closing as we are not observing flaky behavior of this test.

PR #2493, opened 2026-03-14
Fix lost stdout/stderr output on failed task attempts

WHOIM1205:
### Summary
`ArtemisBase.process_multiple` silently discards captured stdout/stderr when a task attempt raises an exception. Because `OutputRedirector.get_output()` is only called on the success path, any logs generated before a crash are lost. As a result, failed tasks are stored in the database with empty or incomplete logs.

This PR ensures that output is collected in both the success and exception paths so diagnostic logs from failed attempts are preserved.

### Problem
When a task fails inside the `with OutputRedirector:` block, execution jumps directly to the `except` block and skips:

```python
output += output_redirector.get_output()

### DIFF
--- a/artemis/module_base.py
+++ b/artemis/module_base.py
@@ -706,6 +706,7 @@ class ArtemisBase(Karton):
                             break
 
                     except Exception:
+                        output += output_redirector.get_output()
                         if i < self.num_retries - 1:
                             self.log.exception("Task(s) failed, retrying (try %d/%d)", i + 1, self.num_retries)
                         else:


WHOIM1205:
HEY @kazet 
This PR fixes an issue where stdout/stderr captured by `OutputRedirector` could be lost if a task raised an exception. Previously `get_output()` was only called on the success path, so logs from failed attempts were discarded. This change ensures output is collected in the exception path as well, preserving diagnostic logs for failed tasks.

PR #2494, opened 2026-03-14
Fix nuclei batch crash on unmatched findings

WHOIM1205:
### Problem
`Nuclei.run_multiple()` currently uses an assertion when a finding cannot be matched to a task:

```python
assert found, "Cannot match finding: %s" % finding

### DIFF
--- a/artemis/modules/nuclei.py
+++ b/artemis/modules/nuclei.py
@@ -707,7 +707,9 @@ class Nuclei(ArtemisBase):
                         findings_per_task[task.uid].append(finding)
                         found = True
                         break
-                assert found, "Cannot match finding: %s" % finding
+                if not found:
+                    self.log.warning("Cannot match finding to any task, assigning to first task: %s", finding)
+                    findings_per_task[tasks[0].uid].append(finding)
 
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
Small note: the `assert found` here may crash the entire batch if Nuclei returns a URL that doesn't exactly match the target (e.g., redirects or CNAME resolution). Maybe this could be handled more gracefully (log + continue) so valid findings aren't discarded.

kazet:
Regarding the first one, we've discussed that earlier - this assert is important as it signals a bug that we shouldn't skip silently.

Regarding the second one, assigning unmatched results to the first task is a very bad idea, as this way we may send vuln info to invalid entity.

WHOIM1205:
> Regarding the first one, we've discussed that earlier - this assert is important as it signals a bug that we shouldn't skip silently.
> 
> 
> 
> Regarding the second one, assigning unmatched results to the first task is a very bad idea, as this way we may send vuln info to invalid entity.

Thanks for the clarification. I'll take care of this and adjust the approach accordingly.

PR #2496, opened 2026-03-15
Implement ORM injection detector module

### LINKED ISSUE #2437: Detect ORM Injection vulnerabilities

imraflip:
Closes #2437 

This PR adds a module for detecting potential ORM injection through query parameter manipulation (right now only focusing on Django-style lookup patterns).

The previous version behaved more like classic SQLi detection which is triggering errors with malformed payloads. And this version instead, focuses on detecting ORM query logic manipulation, such as Django-style lookup parameters (`field__startswith`, `field__contains`, etc.).

The module probes common ORM lookup suffixes and compares responses to determine whether user-controlled parameters are influencing ORM query behavior. It also includes basic checks for filtering on potentially sensitive fields.


artemis/modules/orm_injection_detector.py
```diff
@@ -0,0 +1,234 @@
...
+            return False
+
+        # Double-check: the baseline (original URL as-is) should differ
+        # from at least one of the above to confirm the suffix is actually being processed
+        baseline_response = self.forgiving_http_get(original_url)
+        baseline_text = self._get_response_text(baseline_response)
+        likely_text = self._get_response_text(response_likely)
+        unlikely_text = self._get_response_text(response_unlikely)
+
+        if baseline_text != likely_text or baseline_text != unlikely_text:
```
kadewu: I'm not sure about this. E.g: What if the change was cause by different times send and in response we get sth like 'generated_at'? It would lead to FP am I right?

artemis/modules/orm_injection_detector.py
```diff
@@ -0,0 +1,234 @@
...
+class Statements(Enum):
+    orm_injection = "orm_injection"
+    orm_sensitive_field_access = "orm_sensitive_field_access"
+
+
+@load_risk_class.load_risk_class(load_risk_class.LoadRiskClass.HIGH)
+class OrmInjectionDetector(ArtemisBase):
+    """
+    Module for detecting basic ORM injection vulnerabilities, currently focused
+    on Django-style query parameter manipulation (the most common case).
```
kadewu: I'm okay with this starting point, but let's make it clear in code.
What I mean is to extract the method properly and named it. It's gonna be easier to read in future and as well extend with different styles.

artemis/modules/orm_injection_detector.py
kadewu: I'm not sure about this. E.g: What if the change was cause by different times send and in response we get sth like 'generated_at'? It would lead to FP am I right?
It would also apply to the previous check.

artemis/modules/orm_injection_detector.py
```diff
@@ -0,0 +1,234 @@
...
+
+        return False
+
+    def scan(self, urls: List[str], task: Task) -> List[Dict[str, Any]]:
+        self.log.info("Scanning URLs: %s", urls)
+        message: List[Dict[str, Any]] = []
+
+        for current_url in urls:
+            parsed = urlparse(current_url)
+            query_params = parse_qs(parsed.query)
```
kadewu, line 226: I may be missing it, but where we do actually try to inject some parameters?
Just relying on parsed params is not much.
Let's try to add some common ones at least to this list.

artemis/modules/orm_injection_detector.py
```diff
@@ -0,0 +1,234 @@
...
+
+                for suffix in ORM_LOOKUP_SUFFIXES:
+                    if suffix in ("__startswith", "__istartswith"):
+                        likely_value = original_value[:1] if original_value else "a"
+                    elif suffix in ("__contains", "__icontains"):
+                        likely_value = original_value[:2] if len(original_value) >= 2 else original_value or "a"
+                    elif suffix in ("__endswith", "__iendswith"):
+                        likely_value = original_value[-1:] if original_value else "a"
+                    elif suffix in ("__gt", "__gte"):
+                        likely_value = ""
```
kadewu: AFAIK Empty values may be silently dropped.

docker-compose.test.yaml
```diff
@@ -60,6 +61,15 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
 
+  orm-injection-test-postgres:
```
kadewu: A little nitpicking; let's keep it close to the service using this postgres.
And depends-on in the app.

artemis/modules/orm_injection_detector.py
```diff
@@ -0,0 +1,234 @@
...
+            for field_name, lookup, probe_value in SENSITIVE_FIELD_PROBES:
+                param_with_lookup = f"{field_name}__{lookup}"
+
+                url_with_probe = self._build_url_with_params(base_url, {param_with_lookup: probe_value})
+                url_with_unlikely = self._build_url_with_params(base_url, {param_with_lookup: UNLIKELY_VALUE})
+
+                response_probe = self.forgiving_http_get(url_with_probe)
+                response_unlikely = self.forgiving_http_get(url_with_unlikely)
+
+                if self._responses_differ(response_probe, response_unlikely):
```
kadewu: Here I can see that you are trying to reduce those FP, but still I'm not sure if it's enough - as there is different noise possible, e.g. error?
I don't think just error vs no-error is enough envidence that orm injection is possible.

imraflip:
Hello @kadewu, sorry for the late reply. I was a bit preoccupied this week
I’ll go through your review and start addressing the comments soon. Thanks a lot for the feedback :)!

artemis/modules/orm_injection_detector.py
imraflip: I extracted the django-specific detection into separate named methods like `_test_django_style_lookups()` and `_test_django_sensitive_field_probes()`, so `scan()` is now just an orchestrator. This way if we wanna add other orm styles later (sqlalchemy, sequelize, etc.) we can just drop in new `_test_*` methods without touching the existing ones. Is this exactly what you had in mind?

artemis/modules/orm_injection_detector.py
imraflip: Yeah, you're right. Although I think it's not that common for endpoints to have somthing like `generated_at`, we wanna be safe rather than dealing with FP :)

I think I'd replace the baseline approach with a dynamic content check, it sends the same request twice to the same url first, and if those already differ, we skip the endpoint entirely since differential analysis won't be reliable there. I'd apply this to both the suffix testing and the sensitive field probes

artemis/modules/orm_injection_detector.py
kadewu: Yes, exacatly.

artemis/modules/orm_injection_detector.py
imraflip: The injection happens in `_test_lookup_suffix()` - that's where we take a param name like `username`, append suffixes like `__contains` or `__startswith`, build the url with the modified param, and send the differential requests. `scan()` just loops over the parsed query params and feeds them into it, so the actual injection part isn't super visible from there. And yeah, agreed on not just relying on existing url params, I'd add a list of common param names to probe with too when the url doesn't have any

artemis/modules/orm_injection_detector.py
imraflip: For this, I'm gonna change it to "0" instead

artemis/modules/orm_injection_detector.py
imraflip: Yeah that's a fair point, just comparing responses isn't enough when it could be noise like a server error. So I'm updating `_responses_differ` to skip the comparison when either response is a 5xx - that way crashes or internal errors won't count as evidence. I'm keeping 4xx in though, since some apps return stuff like 404 for "no matching records" which could still be a legit differential. What do you think?

imraflip:
Hey @kadewu, I have updated the PR based on your reviews:

- extracted django-specific detection into named methods `(_test_django_style_lookups`, `_test_django_sensitive_field_probes`),  so it's easier to extend with other ORM styles later
- added dynamic content check to avoid FPs from non-deterministic responses (timestamps, tokens, etc.)
- added common param names list to probe endpoints that don't have existing query params
- changed empty string to `"0"` for `__gt`/`__gte` so it won't get silently dropped
- updated `_responses_differ` to skip 5xx responses since errors aren't evidence of ORM processing
- moved `orm-injection-test-postgres` next to the flask app in docker-compose.test.yaml and added `depends_on`

 Sorry the whole thing took a while, let me know if there's anything else :)!

kazet:
@kadewu pls integrate the ORM injection vuln app I sent you as a test

imraflip:
Hello @kazet @kadewu 

I already fixed the failing CI checks for this PR. The cause was that my branch was forked from main before some changes landed, that's why it broke the tests. So the fix was just merging main into the branch

I also already tested locally with `./scripts/test` and everything works now

kadewu:
Yeah I know the tests failed cause of the problem with dependencies.
I'm integrating slitghtly different app to be tested orm injection against.

.pre-commit-config.yaml
```diff
@@ -21,6 +21,7 @@ repos:
   rev: v1.19.1
   hooks:
   - id: mypy
+    exclude: ^test/data/
```
kadewu: I don't think there is any reason to run mypy on those simple apps we want to test against - that why I sneaked it in.

kadewu:
@imraflip I've changed the app we are testing against and removed the 'safe' endpoint from it + some more precise test for the first one.
Sneaked also some small adjustment like removing unused params in second commit.

Regarding the whole review, I'm still unsure around some comparissons - but I would say that we can most likely observe how it behave agaisnt real internet and then think if there are still FP. Let's see what @.kazet thinks about it.

.pre-commit-config.yaml
kazet: why? mypy may find something, even in test apps, and fixing the linter issues is not a big deal

artemis/orm_injection_data.py
```diff
@@ -0,0 +1,47 @@
+# Common parameter names to probe on endpoints that have no existing query parameters.
+# These are typical names found in Django views, DRF filters, and admin panels.
+COMMON_PARAM_NAMES = [
+    "id",
```
kazet: is there any reason not to use the param list from the SQL injection module? i.e. the params obtained from the webpage (get_injectable_parameters) + a large list of parameters?

test/data/orm_injection/src/templates/posts.html
```diff
@@ -0,0 +1,89 @@
...
+    <title>Post Search</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+    <!-- Modern font -->
+    <link href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400;600&display=swap" rel="stylesheet">
+
+    <!-- Modern minimal CSS framework (Pico.css) -->
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@picocss/pico@2/css/pico.min.css">
+
+    <style>
```
kazet: we may remove the styles as it's a test app

artemis/modules/orm_injection_detector.py
```diff
@@ -0,0 +1,279 @@
...
+        # for "no matching records".
+        status_a = response_a.status_code
+        status_b = response_b.status_code
+        if status_a >= 500 or status_b >= 500:
+            return False
+
+        text_a = self._get_response_text(response_a)
+        text_b = self._get_response_text(response_b)
+
+        return text_a != text_b
```
kazet: comparing for equality is not the best idea, as we will see cases where websites have e.g. CSP nonces, some page generation statistics etc. Let's use some similarity metric, e.g. the one used in https://github.com/CERT-Polska/Artemis/blob/main/artemis/modules/bruter.py#L110 ?

artemis/modules/orm_injection_detector.py
```diff
@@ -0,0 +1,279 @@
...
+
+        # Skip endpoints with non-deterministic content to avoid false positives
+        if self._has_dynamic_content(base_url):
+            return results
+
+        for field_name, lookup, probe_value in SENSITIVE_FIELD_PROBES:
+            param_with_lookup = f"{field_name}__{lookup}"
+
+            url_with_probe = self._build_url_with_params(base_url, {param_with_lookup: probe_value})
+            url_with_unlikely = self._build_url_with_params(base_url, {param_with_lookup: UNLIKELY_VALUE})
```
kazet: unlikely_value is not unlikely for all lookups, e.g. for __lt and __gt it's not ;)

artemis/modules/orm_injection_detector.py
```diff
@@ -0,0 +1,279 @@
...
+        message: list[dict[str, Any]] = []
+
+        for current_url in urls:
+            parsed = urlparse(current_url)
+            query_params = parse_qs(parsed.query)
+            base_url = self._strip_query_string(current_url)
+
+            # Use existing query params if present, otherwise probe with common param names
+            if query_params:
+                params_to_test = query_params
```
kazet: why not use an union: query_params + common_param_names + get_injectable_parameters() ?

artemis/reporting/modules/orm_injection_detector/template_orm_injection.jinja2
```diff
@@ -0,0 +1,41 @@
...
+                                    allowing an attacker to manipulate query logic by appending ORM-specific
+                                    suffixes (e.g. __startswith, __contains, __gt) to parameter names. This
+                                    can be used to extract sensitive data from the database by brute-forcing
+                                    field values character by character.
+                                {% endtrans %}
+                            {% elif message.code == report.additional_data.statements.orm_sensitive_field_access %}
+                                :
+                                {% trans trimmed %}
+                                    ORM sensitive field access: the endpoint allows filtering on sensitive fields
+                                    (such as password hashes, admin flags, or email addresses) through ORM lookup
```
kazet: 'admin flags' is not something easy to understand for everybody, let's skip it

test/modules/test_orm_injection_detector.py
```diff
@@ -0,0 +1,42 @@
...
+            payload={"host": "test-django-with-orm-injection", "port": 8000},
+        )
+        self.run_task(task)
+        (call,) = self.mock_db.save_task_result.call_args_list
+
+        results = call.kwargs["data"]["result"]
+        parameters = {result["parameter"] for result in results}
+
+        self.assertEqual(call.kwargs["status"], TaskStatus.INTERESTING)
+        self.assertTrue(len(results) >= 2)
```
kadewu: Just one thing from me, let's be precise here.
If because of some change we stop discovering one possible injection we should catch it in the test.

kadewu: And that would also apply to the ensuring which one with `self.assertIn('parameter', parameters)` later in the test

imraflip: Done, switched to one assertIn per detected param now, and I dropped the count check entirely

kazet (dismissed):

PR #2497, opened 2026-03-16
Fix task loss on exception in _single_iteration

WHOIM1205:
### Problem
`ArtemisBase._single_iteration()` dequeues tasks via `_take_and_lock_tasks()` but performs several Redis operations afterwards without exception safety.

If a transient Redis error (e.g., `redis.exceptions.ConnectionError`) occurs after tasks are removed from the queue, the worker process crashes before:
- releasing locks
- decrementing `analysis-num-in-progress-tasks`
- completing task processing

This leads to:
- permanently lost tasks (already removed from queue)
- inflated in-progress counters
- leaked locks until TTL expiration

Since every module inherits `ArtemisBase`, this affects all workers.

### Root Cause
Missing `try/finally` protection around the critical section between:
- task dequeue (`_take_and_lock_tasks`)
- task processing (`internal_process_multiple`)

### Fix
Wrap the critical section in `try/except/finally`:

- Track tasks for which the `in-progress` counter was incremented
- On failure:
  - decrement any partially incremented counters
  - re-queue dequeued tasks via `backend.produce_routed_task`
- Always release task locks and the resource lock in `finally`

### Result
- prevents silent task loss during transient Redis failures
- keeps analysis counters consistent
- guarantees lock cleanup
- improves resilience of the worker processing loop

No architectural changes; only adds failure safety around an existing critical section.

### DIFF
--- a/artemis/module_base.py
+++ b/artemis/module_base.py
@@ -349,53 +349,72 @@ class ArtemisBase(Karton):
         tasks, locks, num_task_removed_from_queue = self._take_and_lock_tasks(self.task_max_batch_size)
         self._log_tasks(tasks)
 
-        for task in tasks:
-            increase_analysis_num_in_progress_tasks(REDIS, task.root_uid, by=1)
+        in_progress_incremented_for: list = []
+        try:
+            for task in tasks:
+                increase_analysis_num_in_progress_tasks(REDIS, task.root_uid, by=1)
+                in_progress_incremented_for.append(task)
 
-        requests_per_second_overrides = [
-            task.payload_persistent.get("requests_per_second_override")
-            for task in tasks
-            if "requests_per_second_override" in task.payload_persistent
-        ]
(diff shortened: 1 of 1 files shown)


WHOIM1205:
heyy @kazet 
This PR adds exception safety around the critical section in `_single_iteration()` after tasks are dequeued. If a transient Redis error occurs, the worker could crash before releasing locks or finishing counter updates, causing tasks to be lost. The fix wraps the section in `try/except/finally`, re-queues tasks on failure, and guarantees lock cleanup.

kadewu:
We cannot retry stuff the way it's done here.
If you want to adjust logic on how Artemis actually process things please discuss it beforehand by creating issues.

WHOIM1205:
> We cannot retry stuff the way it's done here.
> 
> If you want to adjust logic on how Artemis actually process things please discuss it beforehand by creating issues.




Sure

PR #2498, opened 2026-03-16
Fix silent subprocess failures in API scanner

WHOIM1205:
### Problem
The API scanner was using `subprocess.call()` to execute OFFAT without checking the return code. If OFFAT failed, execution still continued and attempted to read `/tmp/output.json`. This could lead to:

- Missing vulnerability results when OFFAT fails
- `FileNotFoundError` when the output file is not created
- Stale results being read from a previous scan due to the hardcoded `/tmp/output.json` path
- Silent initialization failures if `git clone`, `git apply`, or `pip install` fail in `__init__`

### Changes
- Replaced `subprocess.call()` with `check_output_log_on_error()` when executing OFFAT to properly capture failures and logs.
- Switched from a shared `/tmp/output.json` to a unique temporary output file per scan.
- Added a `finally` block to ensure the output file is always cleaned up.
- Replaced initialization `subprocess.call()` with `subprocess.check_call()` so OFFAT setup failures are detected during module startup.

### Impact
- OFFAT execution failures are now properly detected and logged.
- Prevents stale scan results from being attributed to the wrong target.
- Ensures initialization issues are surfaced early instead of silently breaking scans.

### DIFF
--- a/artemis/modules/api_scanner.py
+++ b/artemis/modules/api_scanner.py
@@ -16,6 +16,7 @@ from artemis.module_base import ArtemisBase
 from artemis.modules.data.api_scanner_data import COMMON_SPEC_PATHS, VULN_DETAILS_MAP
 from artemis.sql_injection_data import SQL_ERROR_MESSAGES
 from artemis.task_utils import get_target_url
+from artemis.utils import check_output_log_on_error
 
 
 class APIResult(BaseModel):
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
Thanks for pointing this out. I've updated the API scanner to properly handle OFFAT failures by replacing `subprocess.call()` with `check_output_log_on_error`, using a unique output file per scan, and ensuring cleanup in a `finally` block. Initialization commands now use `subprocess.check_call` so failures are not silently ignored.

PR #2511, opened 2026-03-16
Fix distributed lock corruption caused by unconditional Redis.delete in ResourceLock.release

WHOIM1205:
### Problem
`ResourceLock.release()` unconditionally called `REDIS.delete(self.res_name)` without verifying that the caller still owns the lock.

This allowed a deterministic double-release scenario where:
1. One object releases a lock it never acquired.
2. Another process legitimately acquires the same lock.
3. The original object later deletes the Redis key during cleanup.

As a result, one module process could destroy another process's lock, breaking the distributed locking guarantees.

### Impact
This could cause multiple module containers to scan the same destination concurrently, leading to:
- Rate-limit violations against targets
- WAF triggering and scanner IP blocking
- Duplicate scan results
- Corruption of scan task coordination across modules

### Fix
Replace the unconditional `REDIS.delete` with an ownership-aware atomic check-and-delete using a Lua script.

The lock is now removed **only if the stored Redis value matches the caller's lock ID (`lid`)**, preventing cross-process lock deletion.

### Changes
- Added an atomic Redis Lua script for safe lock release
- Updated `ResourceLock.release()` to use ownership-aware deletion
- Updated `ResourceLock.release_all_locks()` to respect lock ownership
- Adjusted unit tests to mock the release script instead of `REDIS.delete`

This preserves the existing architecture while restoring correct distributed lock semantics.

### DIFF
--- a/artemis/resource_lock.py
+++ b/artemis/resource_lock.py
@@ -30,6 +30,18 @@ LOCKS_TO_SUSTAIN_LOCK = threading.Lock()
 
 REDIS = Redis.from_url(Config.Data.REDIS_CONN_STR)
 
+# Lua script for atomic check-and-delete: only deletes the key if its value matches the caller's lock ID.
+# This prevents a process from releasing a lock that was re-acquired by a different process.
+_RELEASE_SCRIPT = REDIS.register_script(
+    """
+    if redis.call("get", KEYS[1]) == ARGV[1] then
+        return redis.call("del", KEYS[1])
+    else
+        return 0
+    end
+    """
+)
+
 
 def sustain_locks() -> None:
     while True:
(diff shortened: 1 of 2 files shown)


WHOIM1205:
hey @kazet 
i noticed that ResourceLock.release() was deleting the Redis key unconditionally which could allow one process to release another process’ s lock in certain double release scenarios. This PR adds an ownership check (via atomic Lua script) before deleting the lock to preserve correct distributed lock behavior.

kadewu:
Why opening new PR instead of addressing the comments on old one?

WHOIM1205:
> Why opening new PR instead of addressing the comments on old one?



> Why opening new PR instead of addressing the comments on old one?

Sorry for the confusion I thought the previous PR was closed to open an issue first for discussing the logic change I will create an issue and discuss the approach there before submitting changes

PR #2512, opened 2026-03-16
feat: add YAML-based configuration support(#433)

### LINKED ISSUE #433: Extract config into YAML

codedByJana:
Closes #433

Changes:

- config.py
  - Added YAML support (`import yaml`, `from pathlib import Path`)
  - Implemented `load_yaml_config()` to read configuration from repository root
  - Updated `get_config()` to check YAML values before environment variables

- scripts/start
  - Added logic to create an example `config.yaml` if it does not exist

- docs/user-guide/configuration.rst
  - Added documentation section describing YAML configuration

- docs/generate_config_docs.py
  - Added YAML configuration section header


artemis/config.py
```diff
@@ -10,7 +12,30 @@
 DEFAULTS = {}
 
 
+def load_yaml_config() -> Dict[str, Any]:
+    """Load configuration from YAML file."""
+    yaml_path = Path(__file__).parent / "config.yaml"
+    if yaml_path.exists():
```
kadewu: I'd rather have proper exception raised here instead of silentely ignoring this fact.

kadewu (commented):
Thank you for the contribution.

Currently unfortunately those changes will not work cause configuration is accessed directly across codebase; please make sure to run the changes to verify that.

Developer also need a way to introduce his configuration - adjusting one big config.yaml is hard and also will be annoying to manage conflict with git - so this should be addressed as well.

kadewu:
Please let me know once you think it's ready for re-review.

codedByJana:
please review it before we run CI tests

scripts/start
```diff
@@ -15,6 +15,17 @@ else
     ENV_MESSAGE=""
 fi
 
+if ! exists config.yaml; then
```
kadewu, line 18: Let's assume that something is broken if this file is not present. I wouldn't add such check

artemis/config.py
```diff
@@ -10,7 +12,57 @@
...
+
+    base_path = root_dir / "config.yaml"
+    try:
+        if base_path.exists():
+            with open(base_path, "r", encoding="utf-8") as f:
+                result = yaml.safe_load(f)
+                if result is not None:
+                    config.update(cast(Dict[str, Any], result))
+    except (FileNotFoundError, PermissionError, OSError) as e:
+        print(f"Error reading config file {base_path}: {e}")
```
kadewu, line 29: And regarding config - same applies here. We want to throw excpetions in such cases.
The artemis should not run if config is broken and that applies to all instances where you try to just handle it with printing.

And for logging please also remember to use logger for that

scripts/start
```diff
@@ -15,6 +15,17 @@ else
...
+    cp config.yaml.example config.yaml
+    YAML_MESSAGE="\e[1mThe config.yaml file has been been created. You can access it later to configure YAML settings.\e[0m"
+else
+    YAML_MESSAGE=""
+fi
+
+if ! exists config.local.yaml; then
```
kadewu, line 25: Correcting myself on previous comment on that, as I've misread this one.
What exactly is `config.local.yaml.example`?

artemis/config.py
```diff
@@ -10,7 +12,57 @@
...
+    except Exception as e:
+        print(f"Unexpected error loading local config {local_path}: {e}")
+
+    return config
+
+
+YAML_CONFIG = load_yaml_config()
+
+
 def get_config(name: str, **kwargs) -> Any:  # type: ignore
```
kadewu, line 55: One more thing, I'm still not sure about this approach.
I'd rather have one source of truth; currently we have config definition in `config.yaml` and in this file.
(cuase of default that you did created config.yaml based on).

Let's assume that config.yaml is fully correct one and the one containing all defaults.

kadewu:
Hello do you still plan to work on that?

codedByJana:
I'm not quite sure how to continue working on this, do you have clear instructions on how to fix this?

kadewu:
The point behind introducing yml file is to have configuration there instead of python config file.
Right now the changes here only introduce yml configuration file - but the definition of variables is still present in both places - which is something we want to avoid.

As an example, if I will want to change default value of whatever variable, and I've seen that we introduced yml file with default values where should I put the change. If I was unware and put it into yml file, and due to error we silently drop it (as in current behavior) then I would get the old unchanges python variable.
The above is just one example, you can think of other multiple problems.

And the above insist on the fact that the default yml file must be present and is critical for whole Artemis working - as said in comments - if something with loading yml config files is broken - then it must broke whole Artemis.

so in summary:
- we want to have one place of truth - new config
- make sure that if config with default values is broken - artemis breaks -> so it enforces developer to fix config file
- default config is going the part of the artemis codebase, same as previously was config.py
- IMO broken additional config file should also break artemis, but this file can be absent, and is not going to be a part of artemis codebase

And I'm not sure about the env variables. Probably best approach would be: 'If additional config is not present, fallback to env variables' - so we do not break setups of open source users with the new change. WDYT?

Feel free to ask any more specific question if something is not clear.

PR #2513, opened 2026-03-17
intergrated web application firewall

### LINKED ISSUE #1708: Integrate web application firewall bypass functionality to modules of artemis
@kazet I would like to work on this

Premkumar-2004:
Added WAF detection and bypass for Artemis. `waf_detect.py` figures out what WAF a target is running (Cloudflare, AWS WAF, ModSecurity) by checking headers, cookies, and sending test probes. `waf_bypass.py` has encoding functions that re-encode payloads to get past the WAF so the scanner can actually test the app behind it. 39 tests included. 

fixes #1708


Premkumar-2004:
@kazet  can you please review it

kazet:
I don't think it's integrated into any Artemis code. Please come back with an integrated PR - BTW, Artemis already has WAF detection, what about improving that first?

PR #2517, opened 2026-03-17
fix: add pool_pre_ping=True to prevent stale DB connections

WHOIM1205:
### Problem
The SQLAlchemy engines in `artemis/db.py` are created without `pool_pre_ping=True`.  
When PostgreSQL drops idle connections (e.g., due to container restarts or timeouts), the connection pool may return a stale connection, causing `sqlalchemy.exc.OperationalError` on the next query.

This particularly affects long-running services with long idle periods (e.g., cleanup and autoarchiver workers).

### Fix
Enable connection health checks by adding `pool_pre_ping=True` to both `create_engine()` calls:

- `DB.__init__`
- `TestDB.__init__`

### Result
SQLAlchemy now validates pooled connections before use and automatically replaces stale ones, preventing crashes caused by dropped PostgreSQL connections.

### DIFF
--- a/artemis/db.py
+++ b/artemis/db.py
@@ -239,7 +239,9 @@ class DB:
         self.logger = build_logger(__name__)
 
         self._engine = create_engine(
-            Config.Data.POSTGRES_CONN_STR, json_serializer=functools.partial(json.dumps, cls=JSONEncoderAdditionalTypes)
+            Config.Data.POSTGRES_CONN_STR,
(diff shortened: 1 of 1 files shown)


kazet (approved):

WHOIM1205:
hey @kazet 
This PR enables pool_pre_ping=True for the SQLAlchemy engines
Without it stale PostgreSQL connections in the pool can cause OperationalError after idle periods or DB restarts.
pool_pre_ping ensures connections are validated before use and stale ones are automatically replaced

PR #2520, opened 2026-03-18
add js secret scanner module

### LINKED ISSUE #2516: Add JS secret scanner module
Artemis currently doesn't check for  secrets leaked in JavaScript files.
For the JS secret scanner : it would go through inline scripts and linked JS files looking for stuff that shouldn't be there. Think AWS keys, GitHub tokens, Stripe keys, private keys, that kind of thing. Secrets in the output get redacted so reports stay safe to share.

Premkumar-2004:
New module that scans JS files for leaked secrets API keys, tokens, passwords, database URIs. Checks both inline scripts and external JS files, follows webpack chunks, and flags exposed source maps. 34 detection patterns with entropy-based filtering to cut down on false positives.
fixes #2516

### DIFF
--- a/artemis/modules/data/js_secret_patterns.py
+++ b/artemis/modules/data/js_secret_patterns.py
@@ -1,0 +1,209 @@
+import re
+from typing import List, NamedTuple
+
+
+class SecretPattern(NamedTuple):
+    name: str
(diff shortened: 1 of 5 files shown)


Premkumar-2004:
@kazet can you please review it

PR #2521, opened 2026-03-18
Fix global log handler side-effect causing permanent log suppression after export

WHOIM1205:
#### Problem
`export(silent=True)` permanently modifies the global `CONSOLE_LOG_HANDLER` log level to `ERROR` without restoring it.

Since this handler is a shared singleton used across the entire application, all `INFO`, `DEBUG`, and `WARNING` logs are suppressed after the first report generation. The autoreporter (which always calls `export(silent=True)`) loses all non-error observability for the rest of its lifecycle.

#### Impact
- Autoreporter appears idle in logs after first report
- Loss of critical operational signals:
  - Task pickup logs
  - Success confirmations
  - Memory usage reporting
  - General system visibility
- Only `ERROR` logs remain visible, hiding degraded states

#### Root Cause
Global log handler level is modified but never restored:
```python
if silent:
    CONSOLE_LOG_HANDLER.setLevel(logging.ERROR)
    
   original_log_level = CONSOLE_LOG_HANDLER.level
if silent:
    CONSOLE_LOG_HANDLER.setLevel(logging.ERROR)

try:
    # existing export logic
    ...
finally:
    CONSOLE_LOG_HANDLER.setLevel(original_log_level)

### DIFF
--- a/artemis/reporting/export/main.py
+++ b/artemis/reporting/export/main.py
@@ -175,63 +175,67 @@ def export(
     skip_suspicious_reports: bool = False,
     include_only_results_since: Optional[datetime.datetime] = None,
 ) -> Path:
+    original_log_level = CONSOLE_LOG_HANDLER.level
     if silent:
         CONSOLE_LOG_HANDLER.setLevel(level=logging.ERROR)
-    blocklist = load_blocklist(Config.Miscellaneous.BLOCKLIST_FILE)
+    try:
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
This fixes a global logging side-effect in `export(silent=True)` where the shared `CONSOLE_LOG_HANDLER` level was not restored, causing all non-error logs to be permanently suppressed after the first export. The change ensures the original log level is restored to avoid impacting observability across the process.

PR #2529, opened 2026-03-19
fix: close SSH connections on failed auth to prevent socket leak

WHOIM1205:
#### Summary
Fixes a resource leak where `paramiko.SSHClient` connections were not closed on failed authentication attempts.

#### Problem
`client.close()` was only called on successful authentication. On failure, exceptions were caught and ignored, leaving SSH connections open.

This caused:
- File descriptor exhaustion (`Too many open files`)
- SSH service disruption (hitting `MaxStartups` limits)
- Worker instability under load

#### Root Cause
Cleanup logic (`client.close()`) was only present in the success path and missing in failure paths.

#### Fix
- Initialize `SSHClient` before the `try` block
- Move `client.close()` into a `finally` block to ensure it always executes

#### Impact
- Eliminates SSH connection leaks
- Prevents FD exhaustion during large scans
- Avoids unintended impact on target SSH services
- Aligns with existing FTP bruter cleanup pattern

#### Notes
Matches the cleanup approach used in `ftp_bruter.py` (`finally: ftp.close()`).

### DIFF
--- a/artemis/modules/ssh_bruter.py
+++ b/artemis/modules/ssh_bruter.py
@@ -59,8 +59,8 @@ class SSHBruter(ArtemisBase):
 
         result = SSHBruterResult()
         for username, password in BRUTE_CREDENTIALS:
+            client = paramiko.client.SSHClient()
             try:
-                client = paramiko.client.SSHClient()
                 client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr fixes missing cleanup of SSH client on failures, preventing socket/FD leaks.

PR #2533, opened 2026-03-21
fix: close streamed HTTP responses to prevent socket leak

WHOIM1205:
### Summary
Fixes a critical resource leak in `artemis/http_requests.py::_internal_request()` where streamed HTTP responses (`stream=True`) were not properly closed when the response body exceeded `max_size`.

### Problem
The function reads response data using `response.iter_content()` and exits early once `max_size` is reached. However, the response object was never explicitly closed.

With `stream=True`, this leads to:
- Connections not being released back to the pool (session reuse path)
- Open sockets accumulating (non-session path)

### Impact
- Connection pool exhaustion (default pool size = 10) → requests hang indefinitely  
- File descriptor leak → `OSError: [Errno 24] Too many open files`  
- Affects all HTTP-based modules (~20+ call sites) under normal scanning load  
- Easily triggered by responses >100KB (common in real-world targets)

### Root Cause
Missing `response.close()` when breaking early from `iter_content()`.

### Fix
Wrap response handling in a `try/finally` block to guarantee cleanup:

```python
try:
    result = b""
    for item in response.iter_content(max_size):
        result += item
        if len(result) >= max_size:
            break
finally:
    response.close()

### DIFF
--- a/artemis/autoarchiver/autoarchiver.py
+++ b/artemis/autoarchiver/autoarchiver.py
@@ -42,8 +42,7 @@ def _save_and_delete_items(old_items: List[Dict[str, Any]], path_suffix: str) ->
 
     LOGGER.info("Saved %s megabytes", os.stat(output_path).st_size / (1024 * 1024 * 1.0))
 
-    for item in old_items:
-        db.delete_task_result(item["id"])
+    db.delete_task_results_by_ids([item["id"] for item in old_items])
 
     LOGGER.info("Deleted %d documents", len(old_items))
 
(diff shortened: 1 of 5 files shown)


WHOIM1205:
Hey @kazet 
This fixes a socket/connection leak in _internal_request() caused by stream=True responses not being closed when exiting early from iter_content().  
Adding a try/finally with response.close() ensures connections are properly released and prevents pool exhaustion / FD leaks under load.

kazet (approved):

Issue #2547, opened 2026-03-24
Add pip-compile flow so that not only base requirements are freezed, but the full requirement set


kazet:

MiracleInvoker:
Hello! I would love to work on this issue. Could you assign it to me?

Based on my understanding of the codebase, my plan would be to:

1. Create a `requirements.in` file and move the direct, top-level dependencies from the current `requirements.txt` into it.
2. Set up `pip-tools` so that running `pip-compile requirements.in` automatically resolves and generates the fully frozen `requirements.txt`.
3. Also, would you like the `requirements.in` approach, or would you prefer moving the base dependencies into `pyproject.toml` under a `[project]` dependencies section and compiling from there?
4. Update any build scripts/CI workflows to include the new compilation step if necessary.

Let me know if this sounds like the right direction.

kazet:
I think 3 is redundant, the rest makes sense!

MiracleInvoker:
Awesome, thanks for the clarification! I will stick with the `requirements.in` approach and get started on the PR.

PR #2549, opened 2026-03-24
build: implement pip-compile workflow for full requirement freezing

MiracleInvoker:
**Resolves Issue:** Add pip-compile flow so that not only base requirements are freezed, but the full requirement set. #2547 

#### **Overview**
This PR introduces `pip-tools` to properly manage and freeze the full dependency tree. It prevents configuration drift by mechanically enforcing that our locked `requirements.txt` files always mathematically match our top-level declarations in `requirements.in`.

#### **Implementation Details**
* **Separation of Concerns:** Extracted top-level, unpinned dependencies into `requirements.in` (main application) and `docs/requirements.in` (documentation).
* **Automated CI/Local Enforcement:** Integrated the official `jazzband/pip-tools` hook into `.pre-commit-config.yaml`. 
  * Configured strictly scoped `files` regexes for both the main and docs environments so they compile independently without triggering infinite loops.
  * This ensures developers cannot commit out-of-sync dependencies, and the existing `lint.yml` CI workflow automatically inherits this validation without needing custom bash scripts.
* **Documentation:** Updated the `README.md` to document the native `pre-commit run pip-compile --all-files` command for managing dependencies.

#### **Important Technical Decisions & Version Pins**
To ensure this PR safely implements the new infrastructure without breaking the application runtime or static analysis, two specific versioning adjustments were made:

1. **Capped `starlette<1.0.0` in `requirements.in`:** An unconstrained upgrade pulled `starlette==1.0.0`, which introduces a breaking API change to `TemplateResponse` (requiring `request` as the first positional argument). This immediately caused 500 Internal Server Errors in the web UI. I capped the version to keep this PR strictly focused on the build infrastructure without requiring immediate application-wide refactoring.
2. **Synced `mypy` Static Analysis Environment:** Upgraded several packages in the `additional_dependencies` block of `.pre-commit-config.yaml` (e.g., `aiohttp`, `alembic`, `flask`, `karton-core`) to perfectly match the newly compiled versions in `requirements.txt`. This ensures the static analysis vacuum does not hallucinate an outdated dependency tree. *(Note: I left `fastapi` and `dnspython` at their existing `mypy` pins to prevent surfacing pre-existing type-hint errors unrelated to this PR).*

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,455 +1,155 @@
 aioboto3==13.3.0
-    # via karton-core
 aiobotocore[boto3]==2.16.0
-    # via aioboto3
 aiofiles==25.1.0
-    # via
-    #   -r docs/../requirements.in
-    #   aioboto3
 aiohappyeyeballs==2.6.1
-    # via aiohttp
 aiohttp==3.13.3
-    # via
-    #   -r docs/../requirements.in
-    #   aiobotocore
 aioitertools==0.13.0
-    # via aiobotocore
 aiosignal==1.4.0
-    # via aiohttp
 alabaster==1.0.0
-    # via sphinx
 alembic==1.18.4
-    # via -r docs/../requirements.in
 annotated-doc==0.0.4
-    # via
-    #   fastapi
-    #   typer
 annotated-types==0.7.0
-    # via pydantic
 anyio==4.13.0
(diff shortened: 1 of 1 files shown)


kazet:
Please fix the CI issues and we're good to go

PR #2550, opened 2026-03-25
fix: close MySQL connections to prevent socket leak

HarshitPal25:
**Problem**
The MySQL bruter calls `pymysql.connect()` inside `throttle_request()` but
discards the returned connection object. The connection is never explicitly
closed — it leaks until Python's garbage collector runs. Under sustained
brute-force scanning the number of open sockets grows continuously, eventually
exhausting available file descriptors and crashing the scanner process with
`OSError: [Errno 24] Too many open files`.
 
**Root Cause**
Line 43: the return value of `self.throttle_request(lambda: pymysql.connect(...))` 
is not assigned to any variable. Even if it were assigned, there is no `finally`
block in the entire function — confirmed by `grep -n "finally" mysql_bruter.py`
returning no results.
 
**Fix**
- Assign the connection object to `conn = self.throttle_request(...)`
- Add a `finally` block that calls `conn.close()` when `conn is not None`
- Wrap the close itself in `try/except` to avoid masking the original error

### DIFF
--- a/artemis/modules/mysql_bruter.py
+++ b/artemis/modules/mysql_bruter.py
@@ -41,7 +41,9 @@ class MySQLBruter(ArtemisBase):
         for username, password in BRUTE_CREDENTIALS:
             conn = None
             try:
-                conn = self.throttle_request(lambda: pymysql.connect(host=host, port=port, user=username, password=password))
+                conn = self.throttle_request(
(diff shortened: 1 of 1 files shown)


HarshitPal25:
@kazet hey kazet there was a socket leak that i have fixed in this pr. In mysql_bruter.py the MySQL connection is not closed on success

PR #2551, opened 2026-03-25
fix: add timeout and close response for Shodan API call in port scanner

HarshitPal25:
**Problem**
Line 167 in `port_scanner.py` calls the Shodan InternetDB API with a bare
`requests.get()` that has three issues:
 
1. No `timeout=` argument — if the Shodan API is slow or unreachable, the
   entire port scan worker hangs indefinitely
2. The response object is never stored or closed — the underlying TCP socket
   leaks after every scan of every IP address
3. `.json()` is called directly on the response — if Shodan returns a
   non-JSON response (rate limit page, 500 error), an unhandled
   `JSONDecodeError` or `requests.exceptions.JSONDecodeError` propagates up
   and silently kills the port scan task
 
**Root Cause**
`requests.get("https://internetdb.shodan.io/" + new_target_ip).json()`  
The chained `.json()` call means the intermediate `Response` object is
never assigned to a variable, making it impossible to call `.close()`.
 
**Fix**
- Store the response object in a variable
- Pass `timeout=Config.HTTP_TIMEOUT_SECONDS`
- Call `response.close()` in a `finally` block after JSON parsing
- This prevents the socket from leaking and caps the wait time

### DIFF
--- a/test/modules/test_port_scanner.py
+++ b/test/modules/test_port_scanner.py
@@ -78,6 +78,7 @@ class PortScannerTest(ArtemisModuleTestCase):
 
         found_ports = set(call.kwargs["data"][test_ip].keys())
         shodan_only_ports = internetdb_ports - {"80"}
+        self.assertTrue(len(shodan_only_ports) > 0)
 
         self.assertTrue(
             shodan_only_ports,


HarshitPal25:
@kazet hey i have added the timeout argument in the port_scanner.py and fix the shodan timeout error and response was not closed it is fixed now

artemis/modules/port_scanner.py
```diff
@@ -171,7 +171,16 @@ def _scan(self, target_ips: List[str]) -> Dict[str, Dict[str, Dict[str, Any]]]:
 
             if Config.Modules.PortScanner.ADD_PORTS_FROM_SHODAN_INTERNETDB:
                 for new_target_ip in new_target_ips:
-                    data = requests.get("https://internetdb.shodan.io/" + new_target_ip).json()
+                    try:
```
kazet, line 174: This part of code is not tested, could you add a test?

But let it be an integration test that actually scans a real IP. Maybe set ports to be scanned to only [80] and check that new from this db are added for e.g. 8.8.8.8 or a well-known ip?

HarshitPal25: integration test added pathched the tests and run it on the server sorry for the delay working on another issue too i found in artemis going to make a pr for it soon and thanks you for the support and guide 🫡

HarshitPal25:
@kazet check the changes i integrate the tests

PR #2552, opened 2026-03-25
fix: use context manager and handle JSONDecodeError in previous reports loader

HarshitPal25:
**Problem**
Line 16 in `previous_reports.py` has two issues:
 
1. `open(path)` is called without a context manager — the file descriptor
   is never explicitly closed. Each call to this function leaks a file handle.
   In long-running report generation workers this accumulates until the process
   hits the open file descriptor limit.
 
2. `json.load()` has no `JSONDecodeError` handler — if the previous report
   file is truncated or corrupted (disk-full crash, SIGKILL during write),
   the exception propagates uncaught and silently kills the entire report
   generation pipeline for all subsequent targets.
 
**Root Cause**
`vulnerability_reports = json.load(open(path))` — the `open()` call is not
wrapped in a `with` statement, so `f.close()` is never called. The chained
pattern also makes it impossible to add error handling around the JSON parse.
 
**Fix**
- Replaced `json.load(open(path))` with a `with open(path) as f:` block
- Added `except json.JSONDecodeError` to log the error and return `[]`
  instead of crashing, so report generation continues for other targets
 
Addresses two of the issues found by static analysis.

### DIFF
--- a/artemis/reporting/export/previous_reports.py
+++ b/artemis/reporting/export/previous_reports.py
@@ -1,5 +1,6 @@
 import datetime
 import json
+import logging
 import os
 from pathlib import Path
 from typing import List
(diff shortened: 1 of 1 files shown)


HarshitPal25:
Hi @kazet  I've opened this PR to fix a file handle leak and a potential JSON crash risk in the previous_reports.py loader that were flagged during static analysis.

Could you please take a look when you have a moment? Thank you!

kazet:
This should crash. If someone has broken report files on disk they should see an exception instead of silently skipping them

HarshitPal25:
ohh okey thank for the remark bro understood

PR #2555, opened 2026-03-26
fix: make Dependabot work with pip-compile flow

### LINKED ISSUE #2553: Make Dependabot work with pip-compile flow

Sumit-ai-dev:
Fixes #2553

The issue was that Dependabot was unable to keep requirements up to date because of the `--no-header` flag in the pip-compile pre-commit hooks. However, simply removing that flag causes CI to break — the generated header contains an environment-specific Python version (e.g. `Python 3.11` in CI vs `Python 3.13` on a dev machine), so the lockfile keeps changing between runs.

The fix is to move `no-header = true` into `pyproject.toml` under `[tool.pip-tools]`. pip-tools reads this config natively, so it applies to every invocation — pre-commit hooks, local runs, and Dependabot's runner — without needing it as a CLI flag. The result is that all environments always produce identical, headerless lockfiles.

Changes:
- `pyproject.toml` — added `[tool.pip-tools]` config section with `no-header = true`
- `.pre-commit-config.yaml` — removed the now-redundant `--no-header` flag from both pip-compile hooks

Verified by running `pre-commit run pip-compile --all-files` inside a `python:3.11-slim` Docker container (matching the CI environment exactly). Both hooks pass cleanly on repeated runs with no file modifications.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,9 +65,9 @@ repos:
   hooks:
   - id: pip-compile
     name: pip-compile main
-    args: [requirements.in, -o, requirements.txt, --no-header]
+    args: [requirements.in, -o, requirements.txt]
     files: ^requirements\.(in|txt)$
   - id: pip-compile
     name: pip-compile docs
-    args: [docs/requirements.in, -o, docs/requirements.txt, --no-header, --no-annotate]
+    args: [docs/requirements.in, -o, docs/requirements.txt, --no-annotate]
     files: ^docs/requirements\.(in|txt)$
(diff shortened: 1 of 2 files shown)


Sumit-ai-dev:
@kazet , @kadewu  pls review it

kadewu:
How did you verify compatibility with Dependabot?

Sumit-ai-dev:
Can't trigger Dependabot locally, so I checked the dependabot core source  it detects --no-header from the file contents, not the CLI args. Since pip-tools reads 
pyproject.toml natively, Dependabot picks up no-header = true when it runs pip compile.
Also ran pre-commit run pip-compile --all-files twice in a python:3.11-slim Docker container  passed clean both times. :)

kazet:
Unfortunately, it doesn't work yet (https://github.com/CERT-Polska/Artemis/pull/2574), so I reopened the issue.

Do you have an idea what is the problem?

PR #2562, opened 2026-03-29
 remove live DNS lookup in test_classifier to stabilize flaky test

vishnudathks:
Test_parsing called gethostbyname("cert.pl") only to build a test input string like "188.184.100.20:80". The classifier itself never resolves DNS for IP inputs, so the lookup was unnecessary and made the test fragile — it would fail if DNS was slow, unreachable, or cert.pl's IP changed.
Replaced with a hardcoded stable IP. Part of #2465.

### DIFF
--- a/test/modules/test_classifier.py
+++ b/test/modules/test_classifier.py
@@ -1,4 +1,3 @@
-from socket import gethostbyname
 from test.base import ArtemisModuleTestCase
 from typing import Any, Dict, List, NamedTuple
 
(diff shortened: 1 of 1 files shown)


kazet:
Hello,

did you encounter this problem in real test runs?

vishnudathks:
Hi @kazet, I didn't observe it failing in real CI runs — I identified it through static analysis of the code. Since you closed it, I'll skip this one and focus only on fixes for tests that have been observed to fail. Thanks for the feedback!

PR #2563, opened 2026-03-29
add retry to subdomain enumeration tests to stabilize flaky tests

vishnudathks:
Test_simple and test_get_subdomains_from_subfinder both make live calls to external services (subfinder, CT logs) that can be slow or rate-limited, causing random failures. test_get_subdomains_from_gau already had @retry(tries=3, delay=10) for the same reason — this applies the same pattern consistently to the other two tests.
Part of #2465.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -87,7 +87,10 @@ click==8.3.1
 click-plugins==1.1.1.2
     # via shodan
 colorama==0.4.6
-    # via shodan
+    # via
+    #   click
+    #   shodan
+    #   tqdm
 coverage==7.13.5
     # via unittest-parallel
 cryptography==46.0.5


test/modules/test_classifier.py
```diff
@@ -48,7 +47,7 @@ def test_support(self) -> None:
         self.assertFalse(Classifier.is_supported("ssh://127.0.0.1"))
 
     def test_parsing(self) -> None:
-        cert_ip = gethostbyname("cert.pl")
+        cert_ip = "188.184.100.20"
```
kazet: I don't think this is needed

vishnudathks:
Hi @kazet! Cleaned up the PR — it now contains only the @retry additions for test_simple and test_get_subdomains_from_subfinder, consistent with the existing @retry already on test_get_subdomains_from_gau. Both tests make live external calls that are known to be flaky. Would appreciate a review when you get a chance!

vishnudathks:
Sorry for the confusion with #2583. I had submitted #2563 without running the tests locally first, which is why the missing import wasn't caught before opening the PR. I've since set up the local Docker test environment and am running tests before submitting. I'll make sure this doesn't happen again.

vishnudathks:
Hi @kazet! The `from retry import retry` import is now present in the branch — it came in via rebasing onto the latest upstream main (where it was added separately). The diff doesn't show it as a new line because it already exists in main. The file now has both the import and `@retry` decorators on all three test methods. Sorry for the confusion!
Screenshot 2026-04-01 224241

vishnudathks:
Hi @kazet! The requirements.txt has been updated via pip-compile. All checks should now be passing. Could you re-review and approve when you get a chance?

PR #2584, opened 2026-04-02
Implement header minimization for SQL injection detection

### LINKED ISSUE #2575: Http headers minimazation
Similarly to parameter minimazation logic(https://github.com/CERT-Polska/Artemis/pull/2483), we want to introduce it for http headers reporting introduced here:
https://github.com/CERT-Polska/Artemis/pull/2424

imraflip:
Closes #2575 

This PR implements header minimization for generated SQLi PoCs reported through HTTP headers.                                                                                     
   
Before the change, when SQL injection (error-based or time-based) was detected via HTTP headers, the PoC included all ~40 injected headers, making it noisy and hard to identify the actual entry point. After detection, the module now re-tests each header individually and keeps only those that independently reproduce the issue. If none of them works alone, it falls back to the original header set to avoid changing detection behavior.                                                                                             
                                                             
The same minimization approach is applied in both header-based detection paths (error-based and time-based).                                                                      
   
Unlike parameter minimization, no cap is applied to the minimized header set. The header pool is fixed and small (~40 entries), so hitting multiple vulnerable headers simultaneously is extremely unlikely - a cap would add complexity for no practical benefit. If the maintainers prefer a cap for consistency, I can add one along with its corresponding test.


imraflip:
Yeah exactly, the tests only trigger `User-Agent`, so there was nothing to trim. I already added the cap now reusing `SQL_INJECTION_MINIMAL_PARAMS_MAX_LEN` for both params and headers, and a test that verifies it (7 headers, 6 vulnerable, capped to 5). Let me know if you'd prefer a separate config for headers or rename the config to something more general like `SQL_INJECTION_MINIMAL_RESULTS_MAX_LEN`

kadewu:
Yes, let's introduce the config for that.

imraflip:
Just to confirm, should I introduce a separate config for headers, or rename the existing one to `SQL_INJECTION_MINIMAL_RESULTS_MAX_LEN` so it covers both?

kadewu:
I believe introducing new config is cleaner.

PR #2586, opened 2026-04-03
feat(docs): add REST API documentation and integration tests

### LINKED ISSUE #2195: Documentation: example on how to add a scan using the REST API

MiracleInvoker:
Closes #2195


test/e2e/test_rest_api_docs.py
```diff
@@ -0,0 +1,188 @@
...
+
+class RestApiDocsTestCase(BaseE2ETestCase):
+    """Test the REST API workflow as documented in rest-api.rst."""
+
+    def test_full_scan_workflow_via_api(self) -> None:
+        """Test the complete documented workflow: add targets, check analyses,
+        monitor queue, retrieve results, and delete the analysis."""
+
+        # Step 1: POST /api/add - Add a target to scan
+        add_response = requests.post(
```
kazet: These tests will soon get inconsistent with what's written in the docs. What do you think about writing tests that extract code snippets from the docs and run them? That way we will always make sure the docs are up to date

MiracleInvoker:
@kazet I have updated the PR based on your review. The tests now extract and execute curl commands directly from the RST docs instead of duplicating them manually.

### How it works:

- Code blocks in `rest-api.rst` are annotated with `.. test-id:: <label>` comments (invisible in rendered docs)
- The tests parse the RST, extract the labeled curl blocks, convert them to Python `requests` calls, and run them against the test instance
- If someone changes a curl example in the docs, the tests automatically pick up the change - so docs and tests can't drift apart

Export endpoints (create / download / delete) are intentionally not annotated here since they require preconditions already covered by `test_exporting.py`.

Let me know if this is what you had in mind or if you would like any changes.

MiracleInvoker:
@kazet just a quick ping, does the updated approach look good to you?

PR #2617, opened 2026-04-14
Prevent wordpress_plugins crash-loop on upstream API failure

WHOIM1205:
### Issue
`WordpressPlugins.__init__` could raise unhandled exceptions when fetching data from WordPress API (e.g. `InvalidResponseException`, `KeyError`, JSON errors).  
This caused the worker to crash during startup, leading to a Docker/Karton restart loop.

### Fix
- Wrapped API fetch + parsing logic in `try/except`
- Handled:
  - `InvalidResponseException`
  - `ValueError` (JSON issues)
  - `KeyError`
  - `RequestException`
- Added safe fallback: `_top_plugins = []`
- Logged a warning instead of crashing

### Impact
- Worker starts reliably even if API is unavailable or malformed
- Prevents crash-loops and queue stalls
- No change in behavior when API works correctly

### Reference
See detailed bug analysis: :contentReference[oaicite:0]{index=0}

### DIFF
--- a/artemis/modules/wordpress_plugins.py
+++ b/artemis/modules/wordpress_plugins.py
@@ -16,7 +16,7 @@ from artemis.binds import TaskStatus, TaskType, WebApplication
 from artemis.config import Config
 from artemis.crawling import get_links_and_resources_on_same_domain
 from artemis.domains import is_subdomain
-from artemis.fallback_api_cache import FallbackAPICache
+from artemis.fallback_api_cache import FallbackAPICache, InvalidResponseException
 from artemis.module_base import ArtemisBase
 
 # Some readmes are long, longer than the default 100kb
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
Handled potential exceptions during startup when fetching WordPress plugins list.  
Added safe fallback to prevent worker crash-loop and ensure service continues even if API fails.

artemis/modules/wordpress_plugins.py
```diff
@@ -234,16 +234,24 @@ class WordpressPlugins(ArtemisBase):
 
     def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
-        response = FallbackAPICache.Urls.WORDPRESS_PLUGINS_LIST.value.get()
```
kazet, line 237: We use FallbackAPICache precisely to have a backup value in case of API failure

PR #2618, opened 2026-04-14
Skip whois heuristic when IP resolution returns None

WHOIM1205:
### Summary
Fixes an issue where `dangling_dns_detector` crashes due to an unhandled `ValueError` when `get_ip_from_domain` returns `None`.

### Problem
- `IPWhois(None)` raises a `ValueError`
- Exception was not caught (only `HTTPLookupError` handled)
- Caused Karton tasks to be marked as **CRASHED** instead of completing normally :contentReference[oaicite:0]{index=0}

### Fix
- Added a guard clause to check if `ip` is `None`
- Logs the failure and safely returns `False` before calling `IPWhois`

### Impact
- Prevents task crashes and retry loops
- Ensures dangling DNS findings are still reported
- No change in behavior for valid IP resolution paths

### Changes
```diff
ip = self.get_ip_from_domain(domain)
+ if not ip:
+     self.log.info("Could not resolve %s to an IP...", domain)
+     return False

### DIFF
--- a/artemis/modules/dangling_dns_detector.py
+++ b/artemis/modules/dangling_dns_detector.py
@@ -360,6 +360,13 @@ class DanglingDnsDetector(ArtemisBase):
         filters = ["university", "educational", "academic", "education", "institute of technology"]
 
         ip = self.get_ip_from_domain(domain)
+        if not ip:
+            self.log.info(
+                "Could not resolve %s to an IP for public-entity whois check; "
+                "skipping heuristic and keeping existing findings.",
+                domain,
+            )
+            return False
         try:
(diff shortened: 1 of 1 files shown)


kadewu:
The crashing is intended behavior.

WHOIM1205:
> The crashing is intended behavior.

understood thanks for the quick review out of curiosity is there a documented place where artemis distinguishes "should crash and surface via Karton" vs "should be 
caught and saved as ERROR/OK"? 
would help me triage future contributions.

kadewu:
There is none, cause there is no general way to describe that.

PR #2620, opened 2026-04-14
Avoid per-row KartonBackend in analyses table

WHOIM1205:
### Problem
`/api/analyses-table` was creating multiple `KartonBackend` instances per row inside `render_analyses_table_row`, leading to:
- Redis connection pool leaks (no cleanup)
- N+1 query pattern (`karton.binds` fetched repeatedly)
- Increased latency and risk of hitting Redis `maxclients` limit

### Fix
- Instantiate a single `KartonBackend` per request in `get_analyses_table`
- Precompute:
  - `all_bind_identities`
  - `disableable_bind_identities`
- Pass these sets to `render_analyses_table_row` instead of recomputing per row

### Impact
- Redis calls reduced from `2N` → `2` per request
- Eliminates connection leak under repeated dashboard refresh
- Improves response latency significantly
- No functional changes to API response

### Notes
- Output remains identical
- Fix is fully backward-compatible

### DIFF
--- a/artemis/api.py
+++ b/artemis/api.py
@@ -294,7 +294,10 @@ def get_analyses_table(
     search_query = _get_search_query(request)
 
     result = db.get_paginated_analyses(start, length, ordering, search_query=search_query)
-    num_pending_tasks = get_num_pending_tasks(KartonBackend(config=KartonConfig()))
+    backend = KartonBackend(config=KartonConfig())
+    num_pending_tasks = get_num_pending_tasks(backend)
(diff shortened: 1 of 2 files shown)


WHOIM1205:
hey @kazet 
this change removes per-row kartonbackend instantiation and avoids repeated redis calls by computing binds once per request

should fix the connection leak and reduce n+1 queries without changing any output

kazet (approved):

PR #2621, opened 2026-04-15
Prevent tempdir leak in handle_single_task

WHOIM1205:
Title: Fix tempdir leak in report generation when skip_previously_exported is enabled

Summary:
This PR fixes a temp directory leak caused by improper placement of the try/finally block in handle_single_task. Previously, tempfile.mkdtemp() was executed before entering the try block, allowing exceptions raised during file copy or DB access to bypass cleanup.

Changes:
- Moved temp directory creation inside the try block
- Ensured cleanup always runs via finally
- Added ignore_errors=True to shutil.rmtree for safer deletion
- Introduced Optional typing for previous_reports_directory

Impact:
Prevents accumulation of leaked temp directories in /tmp during repeated failures, avoiding disk or inode exhaustion in long-running workers

Root Cause:
Exceptions like OSError, PermissionError, or DB failures occurring before the try/finally block caused cleanup to be skipped, leaving temp directories behind

Testing:
- Simulated failures in shutil.copy and DB calls
- Verified temp directory is always cleaned up

### DIFF
--- a/artemis/reporting/task_handler.py
+++ b/artemis/reporting/task_handler.py
@@ -8,6 +8,7 @@ import threading
 import time
 import traceback
 from pathlib import Path
+from typing import Optional
 
 import psutil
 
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
fixed the tempdir leak by ensuring cleanup always runs across all exception paths let me know if any changes are needed

PR #2622, opened 2026-04-15
Avoid caching transient failures in get_injectable_parameters

WHOIM1205:
## Title
fix: prevent caching of transient failures in injectable parameters fetch

## Description
This PR fixes an issue where transient network or parsing failures were being cached due to lru_cache. The function returned an empty list on exception, which was cached as a valid result, causing stale data.

The fix separates the cached logic into a helper function so that exceptions are not cached. Only successful responses are stored, while failures are handled outside.

## Changes
- added _fetch_injectable_parameters as cached helper
- moved exception handling outside cached function
- switched internal return type to tuple for caching
- ensured only successful responses are cached

## Impact
prevents silent false negatives in vulnerability detection caused by stale cached results

### DIFF
--- a/artemis/crawling.py
+++ b/artemis/crawling.py
@@ -1,6 +1,6 @@
 import functools
 import urllib
-from typing import List
+from typing import List, Tuple
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 import requests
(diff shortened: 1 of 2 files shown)


WHOIM1205:
hey @kazet 
this change ensures transient failures are not cached and only successful responses are stored let me know if you want me to adjust anything further

kazet:
Hello,

thank you for the PR! could you write a test for that?

WHOIM1205:
hey @kazet
added regression tests to cover the caching behavior and the original bug scenario (failure → success). ensured transient errors are not cached while successful responses still are.

happy to extend coverage further if needed

PR #2624, opened 2026-04-16
try nuclei host-spray strategy

kadewu:


kadewu:
It seems that in our current setup, in terms of memory, template-spray behaves similarly to host-spray, while significant increase in RPS per host is being observed - therefore I'm closing.

PR #2628, opened 2026-04-17
Add timeout to WordFence feed fetch to prevent worker deadlock

WHOIM1205:
#### Summary
Adds a request timeout to the WordFence vulnerability feed call to prevent indefinite blocking when the external service is slow or unresponsive.

#### Root Cause
`requests.get()` was called without a timeout inside a global `ResourceLock`, causing all workers to hang indefinitely if the request stalled.

#### Fix
- Added `timeout=Config.Limits.REQUEST_TIMEOUT_SECONDS` to the request.

#### Impact
Prevents cluster-wide worker blockage and ensures WordPress plugin processing does not stall due to external API delays.

Refs: :contentReference[oaicite:0]{index=0}


WHOIM1205:
hey @kazet 
this pr Adds a timeout to the WordFence request to prevent workers from hanging under the global lock if the feed is unresponsive.

PR #2629, opened 2026-04-17
fix: count only current chunk in nuclei METRIC_WORK_UNITS

WHOIM1205:
### Summary
Fixes an issue where `METRIC_WORK_UNITS` was over-counted within the per-chunk processing loop in `nuclei.py`.

### Problem
The metric increment used:

### Impact
- Ensures accurate work unit tracking per scan
- Prevents inflated metrics in Prometheus
- Improves reliability of dashboards and alerting


WHOIM1205:
hey @kazet 
Fixes over-counting of METRIC_WORK_UNITS by using per-chunk size instead of full template list. This ensures accurate metrics during chunked execution.

PR #2632, opened 2026-04-17
Restart lock-sustain thread after fork in parallel_loop workers

WHOIM1205:
### Summary
Fixes an issue where the lock sustain daemon thread (`LOCK_SUSTAIN_THREAD`) is not present in forked worker processes, causing Redis-backed locks to expire prematurely.

### Root Cause
The sustain thread is started at import time in the parent process. When `multiprocessing.Process.start()` uses `fork()`, only the calling thread is preserved in the child process, resulting in the absence of the sustain thread in workers.

### Impact
- Locks acquired in worker processes are not refreshed and expire after `LOCK_HEARTBEAT_TIMEOUT`
- Leads to duplicate concurrent scans on the same target
- Breaks request rate limits and may trigger remote bans
- Causes inconsistent queue state and incorrect lock releases

### Fix
- Register `os.register_at_fork(after_in_child=...)` in `resource_lock.py`
- In child process:
  - Reset `LOCKS_TO_SUSTAIN` and reinitialize its lock
  - Start a new sustain daemon thread to handle TTL refresh

### Result
Each worker process independently maintains lock heartbeats, preventing premature expiration and ensuring correct synchronization across parallel execution.

### DIFF
--- a/artemis/resource_lock.py
+++ b/artemis/resource_lock.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import sys
 import threading
 import time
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
 This pr fixes an issue where forked worker processes were not running the lock sustain thread, causing Redis locks to expire during execution. This change ensures each child process reinitializes its own sustain thread so locks are properly refreshed and duplicate work is avoided.

PR #2633, opened 2026-04-17
Add HTTP client timeout and response size limit to wappalyzer

WHOIM1205:
### Summary
The Wappalyzer Go HTTP client was created without a timeout, causing requests to hang indefinitely on unresponsive targets. This could block the subprocess and stall Karton workers until the global timeout is hit.

Additionally, response bodies were read without any size limit, risking excessive memory usage or OOM.

### Changes
- Added a 30s timeout to `http.Client`
- Limited response body size to 10MB using `io.LimitReader`

### Impact
- Prevents indefinite hangs in `webapp_identifier`
- Avoids worker stalls and task crashes
- Reduces risk of OOM from large responses

### DIFF
--- a/artemis/modules/utils/wappalyzer/main.go
+++ b/artemis/modules/utils/wappalyzer/main.go
@@ -10,11 +10,18 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	wappalyzer "github.com/projectdiscovery/wappalyzergo"
 )
 
+const (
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr adds a timeout to the Go HTTP client to prevent indefinite hangs on unresponsive targets, and limits response body size to avoid excessive memory usage.

PR #2635, opened 2026-04-18
Open today's log file in append mode in FileLogger

WHOIM1205:
### Summary
Fixes an issue where the current day's log file was being truncated on every service restart due to opening the file in write (`"w"`) mode.

### Changes
- Switched file open mode from `"w"` to `"a"` to ensure logs are appended instead of overwritten.

### Impact
- Prevents loss of log data across restarts
- Maintains correct log rotation behavior

### Notes
- Rotation logic already handles old logs, so append mode is the correct behavior for the active log file

### DIFF
--- a/artemis/karton_logger.py
+++ b/artemis/karton_logger.py
@@ -26,7 +26,7 @@ class FileLogger(LogConsumer):
                     self.opened_file.close()
                 self._rotate_logs()
                 self.opened_file_date = datetime.datetime.now().date()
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr fix: opening today's log file in `"w"` truncates existing logs on every restart. Switching to `"a"` preserves logs while keeping current rotation logic intact. :contentReference[oaicite:0]{index=0}

PR #2636, opened 2026-04-18
Fix: Prevent worker stall with subprocess timeout in humble.py

HarshitPal25:
### Description
This PR resolves an issue in the `humble` module where `subprocess.check_output()` ran the external `humble.py` tool with no timeout. Previously, if the target was unresponsive or humble hung, the worker would block indefinitely. 

### Changes
- Introduced the existing `Config.Limits.REQUEST_TIMEOUT_SECONDS` config parameter as a timeout to the `humble.py` subprocess.
- Handled the resulting `subprocess.TimeoutExpired` exception so that if the subprocess runs out of time, the incomplete task is safely caught and cleanly marked as `ERRORED` without stalling the rest of the worker.

### DIFF
--- a/artemis/modules/humble.py
+++ b/artemis/modules/humble.py
@@ -100,19 +100,29 @@ class Humble(ArtemisBase):
 
         base_url = get_target_url(current_task)
 
-        data = subprocess.check_output(
-            [
-                "python3",
-                "humble.py",
-                "-u",
-                base_url,
-                "-b",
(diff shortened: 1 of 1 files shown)


HarshitPal25:
hey @kazet
this pr adds a timeout to humble.py to prevent indefinite hangs.

kazet:
I don't think this problem occurs in real life

PR #2637, opened 2026-04-18
Fix: karton logger gzip timeout

HarshitPal25:
### Description
This PR resolves an issue in `artemis/karton_logger.py` where the background log rotation operation ran `subprocess.call(["gzip", ...])` without any timeout constraints. For exceptionally large log files or during heavy I/O conditions, this process could lead to indefinite blocking.

### Changes
- Added `timeout=300` to the `gzip` subprocess command to ensure the system is highly resilient against slow disk operations while actively preventing infinite hangs.
- Catching `subprocess.TimeoutExpired` to cleanly recover if a timeout occurs without propagating an error and crashing the logger background loop.

### DIFF
--- a/artemis/karton_logger.py
+++ b/artemis/karton_logger.py
@@ -61,7 +61,7 @@ class FileLogger(LogConsumer):
                         # Do nothing if failed to acquire the lock
                         continue
                     try:
-                        subprocess.call(["gzip", LOGS_PATH / file_name], timeout=3600)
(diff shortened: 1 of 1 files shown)


HarshitPal25:
@kazet please review it

kazet:
Killing gzip if it takes longer than 300s is not a good idea

HarshitPal25:
@kazet umm ok what if i cange it back to 3600

PR #2638, opened 2026-04-19
Add timeout to joomla-scanner subprocess to prevent worker hang

WHOIM1205:
### Summary
Adds a timeout to the `subprocess.check_output` call in `joomla_extensions` to prevent workers from being blocked by long-running scans.

### Bug
The module invokes `/joomla-scanner/myscanner.py` without a timeout, allowing slow targets to keep the subprocess alive for extended periods, potentially several hours. :contentReference[oaicite:0]{index=0}

### Fix
- Introduced a fixed timeout (`JOOMLA_SCANNER_TIMEOUT_SECONDS`)
- Wrapped subprocess call in try/except
- On timeout:
  - Log warning
  - Mark task as `ERROR`
  - Exit cleanly to release worker resources

### Impact
Prevents worker starvation, frees locks/resources promptly, and avoids task queue buildup under slow target conditions

### DIFF
--- a/artemis/modules/joomla_extensions.py
+++ b/artemis/modules/joomla_extensions.py
@@ -10,6 +10,8 @@ from artemis.binds import TaskStatus, TaskType, WebApplication
 from artemis.config import Config
 from artemis.module_base import ArtemisBase
 
+JOOMLA_SCANNER_TIMEOUT_SECONDS = 600
+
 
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr add a timeout to the joomla_extensions subprocess call to prevent workers from getting stuck on slow targets. On timeout, the task now exits cleanly with an error and releases resources.

PR #2639, opened 2026-04-19
Add timeouts to wappalyzer outer subprocess calls to prevent worker hang

WHOIM1205:
## Summary

Fix potential worker hangs in web_technology_identification caused by unbounded subprocess calls.

## Changes

- Added explicit timeouts to:
  - `go get -u ./...` (60s)
  - `go run main.go` (300s)
- Extended exception handling to include `subprocess.TimeoutExpired`
- Return empty tech results on timeout instead of blocking execution

## Impact

Prevents indefinite blocking of Karton workers during network issues or slow Go module resolution, improving stability of the webapp discovery pipeline.

### DIFF
--- a/artemis/web_technology_identification.py
+++ b/artemis/web_technology_identification.py
@@ -6,6 +6,8 @@ import tempfile
 from typing import Any, List
 
 WAPPALYZER_PATH = "/opt/artemis/modules/utils/wappalyzer/"
+GO_GET_TIMEOUT_SECONDS = 60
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
This pr adds timeouts to the outer subprocess calls that were previously unbounded, as noted in #2633. Should prevent worker starvation during Go module fetch issues. Let me know if you prefer different timeout values.

kazet:
If go get takes longer than expected, it's better to wait than to kill it in the middl

PR #2641, opened 2026-04-19
Handle empty blocklist file in load_blocklist

WHOIM1205:
### Summary
Fixes a crash when the blocklist file is empty or contains only comments/null. `yaml.safe_load()` returns `None`, causing a `TypeError` during iteration.

### Root Cause
`load_blocklist()` assumed parsed YAML is always iterable:
``python
data = yaml.safe_load(file)
for item in data:
Empty or comment-only files return None.

Fix

Default to an empty list if YAML parsing returns None:
data = yaml.safe_load(file) or []
Impact
Prevents worker crash loops on startup
Keeps API and reporting functional
Treats empty/comment-only blocklist as "no entries"
Scope

Minimal, safe change with no impact on valid non-empty blocklist files.

### DIFF
--- a/artemis/blocklist.py
+++ b/artemis/blocklist.py
@@ -54,7 +54,7 @@ def load_blocklist(file_path: Optional[str]) -> List[BlocklistItem]:
         return []
 
     with open(file_path, "r") as file:
-        data = yaml.safe_load(file)
+        data = yaml.safe_load(file) or []
 
     expected_keys = {
         "mode",


WHOIM1205:
hey @kazet 
handled the case where the blocklist file is empty or contains only comments/null.

`yaml.safe_load()` returns `None` for such files, which caused a `TypeError` during iteration and led to worker/API/reporting crashes.

Defaulting to an empty list ensures it behaves like "no blocklist" instead of crashing. Minimal and safe change with no impact on valid configurations.

PR #2642, opened 2026-04-20
Use SCAN instead of KEYS in cleanup to avoid blocking Redis

WHOIM1205:
### Summary
Replaces `backend.redis.keys()` with `scan_iter()` in cleanup to avoid blocking the Redis instance during key enumeration.

### Root Cause
Using `KEYS *` scans the entire keyspace in a single blocking operation. Since Redis is single-threaded, this stalls all other operations (queues, locks, cache) during execution.

### Fix
Switched to `scan_iter(match=...)` for `karton.task*` and `karton.queue*` keys. SCAN is cursor-based and non-blocking, ensuring smooth operation even with large keyspaces.

### Impact
Prevents periodic Redis stalls during cleanup without changing cleanup behavior. :contentReference[oaicite:0]{index=0}

### DIFF
--- a/artemis/cleanup.py
+++ b/artemis/cleanup.py
@@ -23,20 +23,20 @@ def _cleanup_tasks_not_in_queues() -> None:
     # Until https://github.com/CERT-Polska/karton/issues/262 gets fixed, let's have our own cleanup routine
     backend = KartonBackend(config=KartonConfig())
 
-    keys = backend.redis.keys()
+    # Use SCAN (non-blocking, cursor-based) instead of KEYS (blocks Redis for
+    # the entire enumeration): this cleanup shares Redis with Karton queues and
+    # ResourceLock, so a blocking scan of the full keyspace stalls every worker.
     tasks = set()
-    for key in keys:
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr replaces the blocking `KEYS *` call with `scan_iter()` to avoid stalling Redis during cleanup. Since the instance is shared with queues and locks, KEYS could block all operations as the keyspace grows. SCAN keeps the behavior the same while making it non-blocking and scalable. Let me know if you'd like any adjustments.

artemis/cleanup.py
```diff
@@ -23,20 +23,20 @@ def _cleanup_tasks_not_in_queues() -> None:
     # Until https://github.com/CERT-Polska/karton/issues/262 gets fixed, let's have our own cleanup routine
     backend = KartonBackend(config=KartonConfig())
 
-    keys = backend.redis.keys()
+    # Use SCAN (non-blocking, cursor-based) instead of KEYS (blocks Redis for
```
kadewu: Unnecessary comment

PR #2657, opened 2026-04-21
Handle None cvss value in get_severity to prevent TypeError

WHOIM1205:
### Summary
Fixes a crash in `get_severity` when CVE entries contain `cvss=None`. Ensures safe handling by defaulting to `0` during aggregation.

### Root Cause
`item.get("cvss", 0)` returns `None` if the key exists with a `None` value. This causes `max()` to fail with a `TypeError` when comparing `NoneType` and `int`.

### Fix
Replace direct access with a fallback:
```python
cvss = max((item.get("cvss") or 0) for item in report.additional_data["cves"])

### DIFF
--- a/artemis/reporting/severity.py
+++ b/artemis/reporting/severity.py
@@ -99,7 +99,7 @@ def get_severity(report: Any) -> Severity:
         and "cves" in report.additional_data
         and report.additional_data["cves"]
     ):
-        cvss = max(item.get("cvss", 0) for item in report.additional_data["cves"])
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr fixes a crash in `get_severity` when CVE entries have `cvss=None`. 
Uses a safe fallback to `0` to avoid `TypeError` in max()` and keep report stable.

kazet (approved):

PR #2661, opened 2026-04-21
Skip unparseable files in karton log rotation to prevent ingestion crash

WHOIM1205:
### Summary
Fixes a crash in `FileLogger._rotate_logs` when encountering files that do not match the expected `YYYY-MM-DD.log[.gz]` format, which previously broke log ingestion.

### Root Cause
The implementation assumed all filenames were valid and directly unpacked and parsed them. Unexpected files caused `ValueError` during split or date parsing, aborting the entire rotation process. :contentReference[oaicite:0]{index=0}

### Fix
- Skip files without a `.` in the name  
- Wrap date parsing in `try/except`  
- Ignore invalid entries instead of failing the loop  

### Impact
Prevents ingestion failure and ensures log rotation continues even with stray or malformed files.

### DIFF
--- a/artemis/karton_logger.py
+++ b/artemis/karton_logger.py
@@ -38,8 +38,13 @@ class FileLogger(LogConsumer):
 
     def _rotate_logs(self) -> None:
         for file_name in os.listdir(LOGS_PATH):
-            log_date_str, _ = tuple(file_name.split(".", 1))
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
This change makes log rotation more robust by safely skipping invalid or unexpected files instead of crashing the entire process. It ensures ingestion continues even if stray files are present in the logs directory. Let me know if you'd like stricter validation or logging for skipped entries.

Issue #1200, opened 2024-08-12
Figure out some new ways to obtain parameters for XSS and SQLi modules - paramminer? or other ones?


kazet:

kshitijk4poor:
param miner is an extension for burp-site so idts that is possible to implement. there are still python/go options to work with tho

kazet:
@dr4g0n369

dr4g0n369:
Please assign this to me

Fatima-Raees:
Hey @kazet , I'd like to work on this issue!

Before I start, I wanted to check which approach you'd prefer to go with first:

1. Arjun integration — subprocess call, JSON output, easy to wire into existing XSS/SQLi modules
2. x8 — faster (Rust-based), better for large-scale scans, slightly more complex to integrate
3. Passive first (ParamSpider/Wayback CDX) — zero requests to target, mines historical params
4. JS/HTML crawl parsing — extend the existing crawler to extract params from forms and JS files

Based on my research and taking help from AI I think we can start with Arjun as a lightweight enrichment step inside the existing XSS module's `process()` before spinning up a dedicated `param_discovery` Karton module — but happy to go whichever direction makes most sense for the project.

Also, is there a preferred way to emit discovered params downstream — new task type, or enriching the existing task payload?

Thanks!

kazet:
Could you provide links and information what these tools take as input - do they process a single URL or something else?

To answer your second question: we already have a function that takes a url and returns parameter names: https://github.com/CERT-Polska/Artemis/blob/main/artemis/crawling.py#L38 - the goal would be to improve it ;)

Fatima-Raees:
Thanks for the pointer @kazet !

I Checked the `crawling.py` and I think it makes sense to extend that function rather than build a separate module.

Here's what I'm thinking for improving it, in order of complexity:

1. Wayback CDX API : query `web.archive.org/cdx` for historically seen params on the domain. Fully passive (zero requests to target), pure Python, no new dependencies.
2. HTML form/input parsing : extract `name` attributes from forms, inputs, selects from the page response we're already fetching anyway.
3. Arjun (`--passive` mode) : pulls param names from CommonCrawl, OTX, and Wayback without touching the target. Could be a nice complement to number 1 option.- [Arjun](https://github.com/s0md3v/Arjun) — single URL. Pure Python, JSON output.
4. Arjun (active mode) :  brute-forces params against the live URL. More effective but heavier; could be behind a config flag.


[...]

PR #2662, opened 2026-04-21
Add Wayback CDX passive parameter discovery

elaliv:
Add Wayback CDX passive parameter discovery

Mines historical query parameters from the Wayback Machine CDX API (zero requests to target).
Parameters are extracted per domain, cached, and merged with existing HTML input param extraction
in get_injectable_parameters, feeding directly into XSS/SQLi modules.

Unit tests added for parameter extraction, deduplication, caching behavior, and error handling.

The next step is to integrate ParamSpider as an additional passive source for broader historical param coverage.

Part of #1200

### DIFF
--- a/artemis/crawling.py
+++ b/artemis/crawling.py
@@ -7,7 +7,12 @@ import requests
 from bs4 import BeautifulSoup
 from bs4.exceptions import ParserRejectedMarkup  # type: ignore
 
-from artemis import http_requests
+from artemis import http_requests, utils
+
+logger = utils.build_logger(__name__)
+
+WAYBACK_CDX_URL = "https://web.archive.org/cdx/search/cdx"
+WAYBACK_CDX_LIMIT = 5000
 
 
 def get_links_and_resources_on_same_domain(url: str) -> List[str]:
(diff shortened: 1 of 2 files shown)


artemis/crawling.py
```diff
@@ -54,7 +59,50 @@ def _fetch_injectable_parameters(url: str) -> Tuple[str, ...]:
...
+    except Exception:
+        html_params = []
+
+    wayback_params = get_wayback_parameters(url)
+    return list(set(html_params + wayback_params))
+
+
+@functools.lru_cache(maxsize=1024)
+def _fetch_wayback_parameters(domain: str) -> Tuple[str, ...]:
+    try:
```
kazet: let's move this try...except block to get_injectable_parameters, so that transient failures won't get cached

artemis/crawling.py
```diff
@@ -54,7 +59,50 @@ def _fetch_injectable_parameters(url: str) -> Tuple[str, ...]:
...
+    except Exception:
+        logger.exception("Failed to fetch Wayback CDX data for %s", domain)
+        return ()
+
+    params: set = set()
+    for row in rows[1:]:  # skip header row
+        try:
+            parsed = urlparse(row[0])
+            params.update(parse_qs(parsed.query).keys())
+        except Exception:
```
kazet: I'd be more specific what exceptions we're catching here

test/unit/test_wayback_parameters.py
```diff
@@ -0,0 +1,97 @@
...
+        ])
+
+        get_wayback_parameters("http://example.com/")
+        get_wayback_parameters("http://other.com/")
+
+        self.assertEqual(mock_get.call_count, 2)
+
+    def test_invalid_url_returns_empty_list(self) -> None:
+        result = get_wayback_parameters("not-a-url")
+        self.assertEqual(result, [])
```
kazet, line 103: I'd add at least one real integration test that actually connects to Wayback to get the data to check that the format doesn't change

test/unit/test_wayback_parameters.py
```diff
@@ -0,0 +1,123 @@
...
+        _fetch_wayback_parameters.cache_clear()
+
+    def tearDown(self) -> None:
+        _fetch_wayback_parameters.cache_clear()
+
+    def test_real_wayback_cdx_response_format(self) -> None:
+        result = get_wayback_parameters("http://example.com/")
+        self.assertIsInstance(result, list)
+        for param in result:
+            self.assertIsInstance(param, str)
```
kazet: what about finding a real well-known website where you can also check the parameter names?

artemis/crawling.py
```diff
@@ -54,8 +59,48 @@ def _fetch_injectable_parameters(url: str) -> Tuple[str, ...]:
 
 def get_injectable_parameters(url: str) -> List[str]:
     try:
-        return list(_fetch_injectable_parameters(url))
+        html_params = list(_fetch_injectable_parameters(url))
+    except Exception:
+        html_params = []
+
+    wayback_params = get_wayback_parameters(url)
```
kazet: you can add the try...except here as it's with _fetch_injectable_parameters so that the function becomes symmetrical

artemis/crawling.py
```diff
@@ -54,8 +59,48 @@ def _fetch_injectable_parameters(url: str) -> Tuple[str, ...]:
...
+    )
+    response.raise_for_status()
+    rows = response.json()
+
+    params: set[str] = set()
+    for row in rows[1:]:  # skip header row
+        try:
+            parsed = urlparse(row[0])
+            params.update(parse_qs(parsed.query).keys())
+        except (TypeError, IndexError, ValueError):
```
kazet: is it needed to have such generic exce3ptions?

artemis/crawling.py
```diff
@@ -7,7 +7,12 @@
 from bs4 import BeautifulSoup
 from bs4.exceptions import ParserRejectedMarkup  # type: ignore
 
-from artemis import http_requests
+from artemis import http_requests, utils
+
+logger = utils.build_logger(__name__)
+
+WAYBACK_CDX_URL = "http://web.archive.org/cdx/search/cdx"
```
kazet:
```suggestion
WAYBACK_CDX_URL = "https://web.archive.org/cdx/search/cdx"
```

PR #2664, opened 2026-04-22
Ignore blank lines when loading placeholder-page keywords

WHOIM1205:
### Summary
Prevents false-positive placeholder detection caused by blank or whitespace-only lines in the keyword file.

### Root Cause
The loader included empty lines, and `keywords.strip() == ""` always matches any HTML content, making all domains appear as placeholders. :contentReference[oaicite:0]{index=0}

### Fix
Skip empty/whitespace-only lines while loading keywords to ensure only valid entries are used.

### Impact
- Restores correct domain filtering behavior
- Prevents entire scan pipeline from being silently skipped
- Improves robustness against manual file edits


WHOIM1205:
hey @kazet 
This pr change skips blank/whitespace-only lines when loading placeholder keywords, preventing unintended matches where an empty string would classify all domains as placeholders. This ensures the detector behaves correctly even if the keyword file contains trailing or accidental empty lines.

PR #2668, opened 2026-04-22
Don't release scan-destination lock from a non-owning ResourceLock object

WHOIM1205:
### Summary
Fixes a bug where `check_connection_to_base_url_and_save_error` released a scan-destination lock using a non-owning `ResourceLock` instance. This caused unintended deletion of the actual lock and broke locking guarantees.

### Changes
- Removed creation of redundant `ResourceLock` instance
- Removed invalid `lock.release()` calls in WAF and exception branches
- Updated log messages for clarity

### Impact
- Prevents premature deletion of active locks
- Ensures lock ownership integrity across workers
- Avoids concurrent scans on the same destination
- Maintains correct heartbeat behavior for lock sustain mechanism

Ref: :contentReference[oaicite:0]{index=0}

### DIFF
--- a/artemis/module_base.py
+++ b/artemis/module_base.py
@@ -842,7 +842,6 @@ class ArtemisBase(Karton):
     def check_connection_to_base_url_and_save_error(self, task: Task) -> bool:
         base_url = get_target_url(task)
         scan_destination = self._get_scan_destination(task)
-        lock = ResourceLock(f"lock-{scan_destination}", max_tries=Config.Locking.SCAN_DESTINATION_LOCK_MAX_TRIES)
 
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
This PR removes an invalid `lock.release()` on a non-owning `ResourceLock` in `check_connection_to_base_url_and_save_error`.
The actual lock is handled by the outer cleanup, so releasing it here was breaking lock integrity and could allow concurrent scans.
Happy to add tests if needed.

kadewu:
https://github.com/CERT-Polska/Artemis/pull/2370

PR #2669, opened 2026-04-22
Compare cleanup task age against UTC now, not local now

WHOIM1205:
## Summary
Fix incorrect task cleanup threshold caused by mixing UTC and local time comparisons.

## Problem
_cleanup_tasks_not_in_queues compares a UTC timestamp (`utcfromtimestamp`) with a local time (`datetime.now()`), leading to inconsistent cleanup behavior depending on deployment timezone. This results in either premature deletion or delayed cleanup of tasks.

## Fix
Replaced `datetime.now()` with `datetime.utcnow()` to ensure both sides of the comparison use the same UTC reference.

## Impact
Ensures consistent task retention across all timezones and prevents unintended data loss or Redis key buildup.

### DIFF
--- a/artemis/cleanup.py
+++ b/artemis/cleanup.py
@@ -45,7 +45,7 @@ def _cleanup_tasks_not_in_queues() -> None:
         task = json.loads(value)
         if (
             datetime.datetime.utcfromtimestamp(task["last_update"])
-            < datetime.datetime.now() - datetime.timedelta(days=DONT_CLEANUP_TASKS_FRESHER_THAN__DAYS)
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr fixes a timezone mismatch in `_cleanup_tasks_not_in_queues` where UTC and local times were being compared. This could lead to premature or delayed task cleanup depending on deployment timezone. Switched to `datetime.utcnow()` to keep the comparison consistent. Let me know if you'd prefer a timezone-aware approach instead.

PR #2670, opened 2026-04-22
Fix stale resource lock release ownership

HarshitPal25:
## Summary
  - protect `ResourceLock.release()` with an atomic Redis compare-and-delete so stale lock holders cannot delete a lock reacquired by another worker
  - harden `release_all_locks()` the same way because it had the same ownership bug
  - add a real-Redis regression test that reproduces the expiry/reacquire race and verifies the new owner keeps the lock

 ## Bug:

The lock lifecycle used unsafe Redis operations:

sustain_locks() blindly executed:
REDIS.set(key, value, ex=60)

release() blindly executed:
REDIS.delete(key)

If a worker missed heartbeats (e.g., due to CPU starvation or I/O delays),
the lock would expire and be correctly acquired by another worker.
However, when the original worker resumed:

It overwrote the new owner's lock (blind SET)
It deleted the lock entirely on release (blind DEL)
This caused multiple workers to concurrently scan the same target,
breaking rate limits and potentially generating DDoS-like traffic.

## Fix:

Replaced unsafe operations with atomic Redis Lua CAS checks:

Sustain:
Only extend TTL if GET(key) == worker_id

Release:
Only delete key if GET(key) == worker_id

This ensures only the current lock owner can modify or release the lock.

### DIFF
--- a/artemis/resource_lock.py
+++ b/artemis/resource_lock.py
@@ -31,6 +31,14 @@ LOCKS_TO_SUSTAIN_LOCK = threading.Lock()
 
 REDIS = Redis.from_url(Config.Data.REDIS_CONN_STR)
 
+_RELEASE_SCRIPT = """
+if redis.call("get", KEYS[1]) == ARGV[1] then
+    return redis.call("del", KEYS[1])
+else
+    return 0
+end
+"""
+
 
 def sustain_locks() -> None:
     while True:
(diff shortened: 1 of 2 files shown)


HarshitPal25:
hi @kadewu as requested in #2560 i created the integrated tests that passed and fix the bug

kadewu:
The reason why I was asking for integration tests was because I wasn't sure if the changes are not gonna break current task flow.
While I agree that hardening is not bad idea it needs to be done carefully with consideration for all logic, therefore I'm closing.

HarshitPal25:
ohh okey okey thanks

PR #2671, opened 2026-04-22
Treat href-less anchors as empty href in is_log_file classifier

WHOIM1205:
## Summary
Fix crash in `is_log_file` when parsing `<a>` tags without an `href` attribute.

## Bug
`link.get("href")` can return `None`, causing a `TypeError` during substring checks and breaking the entire report generation flow. :contentReference[oaicite:0]{index=0}

## Fix
Default `href` to an empty string before performing checks to prevent runtime errors.

## Impact
Prevents report export failures on directory index pages containing anchor tags without `href`, ensuring stable report generation.


WHOIM1205:
hey @kazet 
This pr fix prevents a crash in `is_log_file` caused by `<a>` tags without an `href`. Defaulting to an empty string ensures the checks run safely and report generation doesn’t fail.

kazet (approved):

PR #2672, opened 2026-04-22
Split add_protocol_if_needed host:port from the right so IPv6 doesn't crash it

WHOIM1205:
Fix IPv6 host:port parsing crash in add_protocol_if_needed

- Replace split(":") with rsplit(":", 1) to correctly handle IPv6 addresses
- Add validation for non-integer ports to prevent ValueError
- Ensure function safely returns original data when parsing fails

Impact:
Prevents report export failures caused by uncaught exceptions when processing IPv6 host:port inputs :contentReference[oaicite:0]{index=0}

### DIFF
--- a/artemis/reporting/utils.py
+++ b/artemis/reporting/utils.py
@@ -98,8 +98,12 @@ def add_protocol_if_needed(data: str) -> str:
     if "://" in data:
         return data
     elif ":" in data:
-        host, port = data.split(":")
-        port_int = int(port)
+        # rsplit so IPv6 host:port (where host contains colons) splits only on the final colon.
+        host, port = data.rsplit(":", 1)
+        try:
+            port_int = int(port)
+        except ValueError:
+            return data
         try:
             service_name = getservbyport(port_int)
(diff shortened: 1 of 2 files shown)


WHOIM1205:
hey @kazet 
This pr fixes a crash in add_protocol_if_needed when handling IPv6 host:port values. Using rsplit ensures only the port separator is considered, and added validation avoids unexpected exceptions. Let me know if you'd like tests added for more edge cases.

artemis/reporting/utils.py
kazet, line 95: if you modify this function, would you add an unittest?

WHOIM1205:
hey @kazet 
added unit tests for `add_protocol_if_needed` covering IPv4, IPv6 (including the reported crash case), existing scheme inputs, and invalid/non-integer ports.
tests follow the current unittest structure in `test/unit/`. They pass within the project's docker test setup.
let me know if you'd like any additional edge cases covered.

kazet (approved):

PR #2676, opened 2026-04-24
Handle None spf/dmarc in mail_dns_scanner reporter instead of crashing export

WHOIM1205:
### Description

Fixes a crash in `mail_dns_scanner` reporter caused by unsafe chained `.get()` calls when `spf` or `dmarc` values are present but set to `None`.

The previous implementation assumed missing keys, but `dict.get()` returns `None` for existing keys with `None` values, leading to `AttributeError` during nested access.

### Fix

Replaced chained `.get(key, {})` patterns with safe fallbacks using `(x.get(key) or {})` and introduced local variables for `spf_dmarc`, `spf`, and `dmarc` to prevent crashes.

### Impact

Prevents report generation failures when `spf` or `dmarc` fields are `None`, ensuring stable exports without requiring manual data cleanup. :contentReference[oaicite:0]{index=0}

### DIFF
--- a/artemis/reporting/modules/mail_dns_scanner/reporter.py
+++ b/artemis/reporting/modules/mail_dns_scanner/reporter.py
@@ -33,10 +33,16 @@ class MailDNSScannerReporter(Reporter):
         MessageWithTarget = namedtuple("MessageWithTarget", "message target type is_warning")
 
         messages_with_targets = []
-        if task_result["result"].get("spf_dmarc_scan_result", {}):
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr fixes a crash caused by `.get()` chains when `spf` or `dmarc` are present but set to `None`. Switched to `(x.get(...) or {})` to safely handle both missing and null values. Let me know if you’d like tests added for this case.

kazet (approved):

PR #2678, opened 2026-04-24
Regenerate the CSRF secret file if it exists but is empty, and write atomically

WHOIM1205:
## Summary
Fixes an issue where `generate_csrf_secret` returned an empty string if `/data/csrf_secret` existed but was zero-length, resulting in an invalid CSRF signing key.

## Fix
- Treat empty secret file as missing and regenerate it
- Use atomic writes (tmp file + `os.replace`) to prevent partial writes

## Impact
Prevents scenarios where interrupted writes could silently disable CSRF protection.

### DIFF
--- a/artemis/csrf.py
+++ b/artemis/csrf.py
@@ -20,14 +20,26 @@ def generate_csrf_secret() -> str:
 
     if os.path.exists(csrf_secret_path):
         with open(csrf_secret_path) as f:
-            return f.read()
+            existing = f.read()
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr handled edge case where `/data/csrf_secret` exists but is empty by regenerating it instead of using an invalid key. Also updated write logic to be atomic to prevent partial-write corruption.

PR #2679, opened 2026-04-24
Implement authorization layer for frontend routes

### LINKED ISSUE #2623: Add some authorization layer so that it's not possible to access Artemis just by browsing to [ip]:5000

imraflip:
Closes #2623 

This PR implements a frontend authorization layer for Artemis web UI routes.

Before this change, frontend pages could be accessed without proper auth enforcement. After this change, protected frontend routes require authentication, unauthenticated users are redirected to the login page, and login/logout flow is integrated into the UI flow.

I also manually verified the behavior end-to-end:

- Visit `http://localhost:5000` without creds set in `.env` -> redirected to `/login`, page shows blue info alert: `No frontend credentials configured...`.
- Add `FRONTEND_USERNAME=admin` and `FRONTEND_PASSWORD=s3cret` to `.env`, then restart the stack.
- Visit `http://localhost:5000` -> redirected to `/login`, and the info alert is no longer shown (credentials are configured).
- Submit wrong credentials -> remain on `/login` with red error alert.
- Submit correct credentials -> redirected to `/`, then frontend pages are accessible.


artemis/auth.py
```diff
@@ -0,0 +1,62 @@
...
+UNPROTECTED_PATH_PREFIXES = ("/api/", "/static/")
+UNPROTECTED_PATHS = ("/login", "/logout")
+
+
+def generate_session_secret() -> str:
+    session_secret_path = "/data/session_secret"
+
+    if os.path.exists(session_secret_path):
+        with open(session_secret_path) as f:
+            return f.read()
```
kazet: let's check that the secret is not empty in case somthing went wrong during saving ;)

env.example
```diff
@@ -2,5 +2,7 @@ POSTGRES_CONN_STR=postgresql://postgres:postgres@postgres/artemis
 DB_CONN_STR=mongodb://root:root@db
 REDIS_CONN_STR=redis://redis:6379/1
 CUSTOM_USER_AGENT=
+FRONTEND_USERNAME=admin
```
kazet: I'm not a big fan of default passwords - more lazy users will leave the default password thus making ths instance vulnerable.

What do you think of just skipping the default username and password and making `./script/start` **not start** if the credentials aren't configured?

artemis/config.py
```diff
@@ -213,6 +213,18 @@ class Miscellaneous:
             "API_TOKEN", default=None
         )
 
+        FRONTEND_USERNAME: Annotated[
+            str,
+            "Username used to log in to the Artemis web interface. Leave empty together with FRONTEND_PASSWORD to "
```
kazet: I think we can even modify ./scripts/start so that Artemis doesn't start with these credentials

env.example
imraflip: TThat's a good idea :), I already removed the defaults from env.example and added a check in ./scripts/start that refuses to start if the credentials aren't set

PR #2681, opened 2026-04-26
Don't assert in vcs reporter when both SVN and Git are detected on one host

WHOIM1205:
### Problem
The reporter asserted that both `svn` and `git` cannot be true simultaneously. However, valid scenarios (e.g. migrations, mirrors, honeypots) can expose both repositories. This assertion caused the DataLoader to crash during export, blocking all future exports until the offending row was manually removed.

### Changes
- Removed the assertion blocking dual VCS detection.
- Added safe handling using `.get()` for `svn` and `git` keys.
- Generate reports for both SVN and Git when both are present.
- Added a warning log when dual repositories are detected.
- Ensured export pipeline continues without interruption.

### Impact
Prevents export failures caused by valid dual-repository cases and improves robustness of the reporting pipeline.

### DIFF
--- a/artemis/reporting/modules/vcs/reporter.py
+++ b/artemis/reporting/modules/vcs/reporter.py
@@ -28,17 +28,20 @@ class VCSReporter(Reporter):
         if not task_result["status"] == "INTERESTING":
             return []
 
-        assert not (
-            task_result["result"]["svn"] and task_result["result"]["git"]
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr removed the assert that caused export crashes when both SVN and Git were detected. Now logs a warning and reports both cases safely without breaking the DataLoader.

PR #2682, opened 2026-04-26
Don't crash removed_domain_existing_vhost reporter when original_domain is None

WHOIM1205:
## Summary
Fix crash in removed_domain_existing_vhost reporter when original_domain is missing.

## Bug
The reporter used `payload_persistent.get("original_domain")` and directly concatenated it with `"www."`. When the key was missing, this resulted in a TypeError (`"www." + None`), causing the export pipeline to fail.

## Fix
- Extract `original_domain` into a local variable
- Add a guard to ensure it exists before comparison
- Skip `www.` concatenation when `original_domain` is None

## Impact
Prevents report generation failures caused by malformed or legacy task data and avoids persistent export crashes.

### DIFF
--- a/artemis/reporting/modules/removed_domain_existing_vhost/reporter.py
+++ b/artemis/reporting/modules/removed_domain_existing_vhost/reporter.py
@@ -24,13 +24,15 @@ class RemovedDomainExistingVhostReporter(Reporter):
         if not task_result["status"] == "INTERESTING":
             return []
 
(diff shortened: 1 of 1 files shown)


WHOIM1205:
hey @kazet 
this pr handled a potential None case for `original_domain` to prevent crashes during report generation. Let me know if you'd like me to add a test for this scenario as well.

kazet (approved):

PR #2696, opened 2026-04-30
Add ParamSpider as additional passive parameter source

elaliv:
Added ParamSpider as an additional passive source for parameter discovery, alongside the existing HTML form scraping and Wayback CDX. The parameters found are merged in get_injectable_parameters() and fed into the XSS/SQLi modules.

Part of #1200

### DIFF
--- a/artemis/crawling.py
+++ b/artemis/crawling.py
@@ -1,5 +1,8 @@
 import functools
+import subprocess
+import tempfile
 import urllib
+from pathlib import Path
 from typing import List, Tuple
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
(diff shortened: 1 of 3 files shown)


artemis/crawling.py
```diff
@@ -108,6 +116,39 @@ def get_wayback_parameters(url: str) -> List[str]:
         return []
 
 
+@functools.lru_cache(maxsize=1024)
+def _fetch_paramspider_parameters(domain: str) -> Tuple[str, ...]:
+    with tempfile.TemporaryDirectory() as tmpdir:
+        output_file = Path(tmpdir) / f"{domain}.txt"
+        subprocess.run(
+            ["paramspider", "-d", domain, "--output", str(output_file)],
```
kazet, line 124: what paramspider does is query Wayback CDX. What is the difference over what we already have?

kazet, line 124: what paramspider does is query Wayback CDX. What is the difference over what we already have?

kazet:
Hello,

if you plan to further develop this PR, let me know.

PR #2709, opened 2026-05-05
Nuclei routing

kadewu:
Introducing routing karton module.

> [!WARNING]
> Changed nuclei module identify from `nuclei` to `nuclei-module`. This affects `module_runtime_configuration` which needs to be adjusted if used. While the previous identity would still be properely processed it should be changed to new value when creating new entries.

Ideas behind separate routing module:
- run WAF before collecting tasks, so we use max_batch more efficiently
- grouping outside of nuclei modules allows us to use concurrency more efficiently (we scan batch of x targets with the same set of templates)

### DIFF
--- a/artemis/module_base.py
+++ b/artemis/module_base.py
@@ -567,6 +567,10 @@ class ArtemisBase(Karton):
     def run_multiple(self, tasks: List[Task]) -> None:
         raise NotImplementedError()
 
+    def get_batch_group_key(self, _task: Task) -> str | None:
+        """Return a grouping key used when taking a batch of tasks from queue."""
+        return None
+
     def internal_process_multiple(self, tasks: List[Task]) -> None:
         if not tasks:
             return
@@ -738,7 +742,7 @@ class ArtemisBase(Karton):
                 result = self._get_ip_for_locking(host)
             except UnknownIPException:
                 result = host
-        elif task.headers["type"] == TaskType.URL:
+        elif task.headers["type"] == TaskType.URL or task.headers["type"] == TaskType.NUCLEI_TARGET:
             host = urllib.parse.urlparse(task.payload["url"]).hostname
             try:
                 result = self._get_ip_for_locking(host)
--- a/artemis/module_base.py
+++ b/artemis/module_base.py
@@ -742,13 +742,13 @@ class ArtemisBase(Karton):
                 result = self._get_ip_for_locking(host)
             except UnknownIPException:
                 result = host
-        elif task.headers["type"] == TaskType.URL or task.headers["type"] == TaskType.NUCLEI_TARGET:
+        elif task.headers["type"] == TaskType.URL:
             host = urllib.parse.urlparse(task.payload["url"]).hostname
             try:
                 result = self._get_ip_for_locking(host)
             except UnknownIPException:
                 result = host
-        elif task.headers["type"] == TaskType.SERVICE:
+        elif task.headers["type"] == TaskType.SERVICE or task.headers["type"] == TaskType.NUCLEI_TARGET:
             try:
                 result = self._get_ip_for_locking(task.payload["host"])
             except UnknownIPException:
--- a/artemis/module_base.py
+++ b/artemis/module_base.py
@@ -567,10 +567,6 @@ class ArtemisBase(Karton):
     def run_multiple(self, tasks: List[Task]) -> None:
         raise NotImplementedError()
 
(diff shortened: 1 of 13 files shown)


artemis/modules/nuclei.py
```diff
@@ -192,14 +193,30 @@ class Nuclei(ArtemisBase):
     """
 
     num_retries = Config.Miscellaneous.SLOW_MODULE_NUM_RETRIES
-    identity = "nuclei"
+    identity = "nuclei-module"
```
kazet, line 194: I see a slight problem with this rename: we already have many tasks in the queue with runtime configuration with the old name. Can we modify the runtime configuration module to understand the old identity?

kazet, line 194: I see a slight problem with this rename: we already have many tasks in the queue with runtime configuration with the old name. Can we modify the runtime configuration module to understand the old identity?

artemis/module_base.py
```diff
@@ -435,8 +435,9 @@ def _take_and_lock_tasks(self, num_tasks: int) -> Tuple[List[Task], List[Optiona
             return [], [], 0
 
         try:
-            tasks = []
+            tasks: list[Task] = []
             locks: List[Optional[ResourceLock]] = []
+            selected_batch_group_key: str | None = None
```
kazet, line 406: I would add a docstring somewhere what this does and what's the purpose

artemis/modules/nuclei.py
```diff
@@ -192,14 +193,30 @@ class Nuclei(ArtemisBase):
...
+            return []
+
+        if any(task.payload.get(NUCLEI_ROUTER_FLAGS_PAYLOAD_KEY, []) != first_task_flags for task in tasks[1:]):
+            self.log.warning("Nuclei picked up tasks from different groups")
+            return []
+
+        return [item for item in first_task_flags if isinstance(item, str)]
+
+    def get_batch_group_key(self, task: Task) -> str | None:
+        return "\x1f".join(self._get_nuclei_router_flags([task]))
```
kazet: why \x1f? Won't it make some problems with debugging later?

artemis/modules/nuclei_router.py
```diff
@@ -0,0 +1,93 @@
...
+from karton.core import Task
+
+from artemis import load_risk_class
+from artemis.binds import Service, TaskStatus, TaskType
+from artemis.module_base import ArtemisBase
+from artemis.task_utils import get_target_host, get_target_url
+from artemis.web_technology_identification import run_tech_detection
+
+TECHNOLOGY_DETECTION_TAGS_TO_EXCLUDE = {"wordpress": ["wordpress"]}
+# it's seperate with module configuration cause flags with values defined in router serve
```
kazet: it's separate from - and I will put the two purposes in bullet points

artemis/modules/nuclei_router.py
```diff
@@ -0,0 +1,93 @@
...
+# two purposes, configuration runtime for nuclei command and define grouping key to pickup tasks
+NUCLEI_ROUTER_FLAGS_PAYLOAD_KEY = "nuclei-routing-additional-flags"
+
+
+@load_risk_class.load_risk_class(load_risk_class.LoadRiskClass.LOW)
```
kazet: Hmm, as it pairs with Nuclei, maybe high? I won't die on that hill though

artemis/modules/nuclei_router.py
```diff
@@ -0,0 +1,93 @@
...
+            tags_to_exclude.update(TECHNOLOGY_DETECTION_TAGS_TO_EXCLUDE[tech_name])
+        return sorted(tags_to_exclude)
+
+    def get_nuclei_additional_flags_for_task(self, target_url: str) -> list[str]:
+        flags = []
+
+        # Nuclei module implementation currently only supports '-etags'
+        tags_to_exclude = self.get_tags_to_exclude(target_url)
+        if tags_to_exclude:
+            flags.extend(["-etags", ",".join(tags_to_exclude)])
```
kazet, line 56: Does etags override both itags and the `-it` flag we use?

artemis/modules/nuclei.py
kadewu: Was a random string, changed that

artemis/modules/nuclei_router.py
kadewu: `-it` and `-itags` override `-etags` but the impact of 'unnecessary' included templates will be currently low

kazet (approved):

PR #2738, opened 2026-05-17
Implement katana + uro crawling pipeline with shared redis cache

imraflip:
This PR implements the Katana + uro crawling pipeline, replacing the old single-page HTML link extractor.

Before the change, `get_links_and_resources_on_same_domain` just fetched the seed URL's HTML and returned the links found on that one page, missing JS-rendered routes, paginated content, and anything reachable only by following further links. After the change, `crawl_and_filter` walks the target with Katana (with `-jc` for static JS parsing and `-jsonl` output), pipes the result through uro for structure-aware deduplication so `/post/1` and `/post/2` collapse to one entry, and filters the result back to the target host.

The result is cached in a shared Redis namespace (`cache.crawl_pipeline:*`, DB 1) so concurrent modules scanning the same target read one crawl instead of spawning their own Katana process. Successful crawls get a 24-hour TTL; crawls that hit the Katana subprocess timeout get a shorter 1-hour TTL so a partial result is retried sooner. The cache key is sha256 over the normalized URL plus the knobs that actually affect crawl output (depth, max URLs, user agent) - the timeout is deliberately excluded so changing it doesn't invalidate prior results. A ResourceLock serializes concurrent crawls of the same target, and a re-check inside the lock makes the losing worker pick up the winner's cache entry instead of running Katana redundantly.

Katana is installed via `go install pkg@v1.6.1` using a Go 1.25 toolchain mounted at `/usr/local/go125/`, separate from the go.mod-tracked tools (naabu, fingerprintx, gau, subfinder, nuclei) which stay on Go 1.24 with the existing pattern unchanged. Adding katana to `go.mod` was attempted first but requires bumping the go directive from 1.24 to 1.25, which in turn forces a nuclei bump from v3.4.7 to v3.8.0+ because v3.4.7 doesn't compile against the newer transitive deps Go 1.25 resolution pulls in. That nuclei jump is unrelated to crawling and was deferred to a follow-up PR after mentor consultation.

Module tests mock `crawl_and_filter` because Katana's URL validator rejects Docker Compose service names (no TLD) - production targets have real domains so this only affects the test environment. Each mock returns exactly what the old HTML extractor would have returned from parsing the corresponding test target's index page, so the existing assertions are unchanged.


test/modules/test_sql_injection_detector.py
```diff
@@ -17,7 +17,14 @@ def test_sql_injection_detector(self) -> None:
             {"type": TaskType.SERVICE.value, "service": Service.HTTP.value},
             payload={"host": "test-apache-with-sql-injection-postgres", "port": 80},
         )
-        self.run_task(task)
+        with patch(
+            "artemis.modules.sql_injection_detector.crawl_and_filter",
+            return_value=[
```
kazet: why do we mock that? IMO let's check whether sql_injection_detector properly integrates with the crawling pipeline

test/modules/test_nuclei.py
```diff
@@ -11,6 +11,15 @@ class NucleiTest(ArtemisModuleTestCase):
     # The reason for ignoring mypy error is https://github.com/CERT-Polska/karton/issues/201
     karton_class = Nuclei  # type: ignore
 
+    def setUp(self) -> None:
+        self.crawl_patcher = patch(
```
kazet: why do we mock that? IMO let's check whether nuclei_detector properly integrates with the crawling pipeline

test/modules/test_lfi_detector.py
```diff
@@ -17,31 +17,33 @@ def test_lfi_detector_with_rce(self) -> None:
...
-                call.kwargs["status_reason"],
-            )
-
-            self.assertIn(
-                "%0a/bin/cat%20/etc/passwd",
-                call.kwargs["status_reason"],
-            )
+        with patch(
+            "artemis.modules.lfi_detector.crawl_and_filter",
+            return_value=["http://test-apache-with-lfi-and-rce:80/page.php"],
```
kazet: why do we mock that? IMO let's check whether lfi_detector properly integrates with the crawling pipeline

artemis/config.py
```diff
@@ -370,6 +370,35 @@ class Bruter:
                 "doesn't exist, thus decreasing the number of false positives at the cost of losing some true positives.",
             ] = get_config("BRUTER_FOLLOW_REDIRECTS", default=True, cast=bool)
 
+        class Crawling:
+            KATANA_DEPTH: Annotated[int, "Crawl depth passed to Katana (-d)."] = get_config(
+                "KATANA_DEPTH", default=3, cast=int
```
kazet: 3 is very much, it can slow down the scanning. What about 2?

artemis/config.py
```diff
@@ -370,6 +370,35 @@ class Bruter:
...
+            )
+
+            KATANA_MAX_URLS: Annotated[int, "Hard cap on URLs collected by Katana per target (passed as -mdp)."] = (
+                get_config("KATANA_MAX_URLS", default=2000, cast=int)
```
kazet: now it's 20 after deduplication, I won't make the value much higher in order not to slow down scanning.

test/unit/test_crawling_pipeline.py
```diff
@@ -0,0 +1,186 @@
...
+        self.lock_patcher = patch.object(crawling, "ResourceLock")
+        self.mock_lock_cls = self.lock_patcher.start()
+        self.mock_lock = MagicMock()
+        self.mock_lock_cls.return_value = self.mock_lock
+
+    def tearDown(self) -> None:
+        self.cache_patcher.stop()
+        self.lock_patcher.stop()
+
+    def test_cache_hit_skips_katana_and_lock(self) -> None:
```
kazet, line 49: what I am missing here is running the method on a real website (e.g. run as a Docker container) and checking whether the pipeline returns good results

test/modules/test_sql_injection_detector.py
imraflip: I mocked it because katana can't resolve our dotless docker hostnames, it somehow rejects hostnames with no dots before even doing the lookup. But katana works fine against IPs and FQDNs, but on names like `test-apache-with-sql-injection-postgres`, it just returns nothing. The old crawler handled those names naturally - that's why the previous test had no mock

With katana in the pipeline, the crawler returns `[]` and the module has nothing to scan. The mock just substitutes the URLs the old crawler used to find, so the rest of the module logic still gets exercised. Production scans FQDNs so I thought it wasn't an issue there

imraflip: image

artemis/config.py
imraflip: Sure, I'm gonna bump it down to 2 ;)

artemis/config.py
imraflip: Well you're right, the gap between 25 and 2000 is huge. What about 50–100?

test/unit/test_crawling_pipeline.py
imraflip: Same situation as the mocked tests, katana can't resolve the dotless hostnames. But I can add a `.local` alias to one of the containers and write an integration test in test/modules that hits the real pipeline. What do you think?

kazet: you can add .local or scan by IP - provide gethostbyname(domain) as the target to scan

artemis/config.py
kazet: let's use 50 :)

test/modules/test_sql_injection_detector.py
kazet: you can add .local or scan by IP - provide gethostbyname(domain) as the target to scan

test/unit/test_crawling_pipeline.py
imraflip: Done, I have replaced the mocks with real integration tests ;)

kazet (dismissed):

PR #2740, opened 2026-05-17
 Add ProxyHeadersMiddleware to fix mixed-content when using SSL termination

### LINKED ISSUE #2715: Static content (css, favicon) is blocked as mixed-content when using SSL termination (apache2, nginx)
Tried both, nginx and apache2 but the issue stays the same: following URLs are generated with http instead of https which leads to "blocked mixed-content" and entire page looks ugly (no css):
* http://artemis.example.com/static/bootstrap.min.css?v=2
* http://artemis.example.com/static/style.css?v=3
* http://artemis.example.com/static/theme.js
* http://artemis.example.com/static/images/favicon.ico

I'm attaching anonymized (domain, IP) HAR file
[artemis_static_http.json](https://github.com/user-attachments/files/27516427/artemis_static_http.json)

LLM suggestion was to add `PREFERRED_URL_SCHEME=https` to `.env`, which I did and run `./scripts/run_docker_compose --mode=production down` and `./scripts/start --mode=production` after that - still no luck.

relevant part of my web servers configuration files
**nginx**
```
location / {
        proxy_pass http://127.0.0.1:5000;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_set_header X-Forwarded-Port 443;
        proxy_buffering off;
        proxy_redirect off;
    }
```

**apache2**
```
    ProxyPreserveHost On
    ProxyRequests Off
    RequestHeader set X-Forwarded-Proto "https"
    RequestHeader set X-Forwarded-Port "443"
    RequestHeader set X-Forwarded-For %{REMOTE_ADDR}s
    ProxyPass / http://127.0.0.1:5000/
    ProxyPassReverse / http://127.0.0.1:5000/
```

vishnudathks:
When Artemis is deployed behind nginx or apache2 with SSL termination,
static files (CSS, JS, favicon) are served with `http://` URLs instead
of `https://`, causing browsers to block them as mixed content and
breaking the UI.

The proxy correctly sets `X-Forwarded-Proto: https` but Artemis was
not reading it.

## Fix
Add `ProxyHeadersMiddleware` from uvicorn (already a dependency) so
that Artemis respects the `X-Forwarded-Proto` header and generates
correct `https://` URLs for static content.

## Changes
- Add `ProxyHeadersMiddleware` import and middleware in `artemis/main.py`

Closes #2715

### DIFF
--- a/artemis/main.py
+++ b/artemis/main.py
@@ -15,6 +15,7 @@ from artemis.config import Config
 from artemis.db import DB
 from artemis.frontend import error_content_not_found
 from artemis.frontend import router as router_front
+from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 
 faulthandler.register(signal.SIGUSR1)
 
@@ -35,6 +36,7 @@ app.add_middleware(
     same_site="strict",
     https_only=False,
 )
+app.add_middleware(ProxyHeadersMiddleware, trusted_hosts="*")
 
 db = DB()
 
--- a/artemis/main.py
+++ b/artemis/main.py
@@ -7,6 +7,7 @@ from fastapi import FastAPI
 from fastapi.staticfiles import StaticFiles
 from fastapi_csrf_protect.exceptions import CsrfProtectError
 from starlette.middleware.sessions import SessionMiddleware
+from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 
 from artemis import csrf
 from artemis.api import router as router_api
@@ -15,7 +16,6 @@ from artemis.config import Config
 from artemis.db import DB
 from artemis.frontend import error_content_not_found
 from artemis.frontend import router as router_front
-from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 
 faulthandler.register(signal.SIGUSR1)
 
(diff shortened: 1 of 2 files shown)


artemis/main.py
```diff
@@ -35,6 +36,7 @@
     same_site="strict",
     https_only=False,
 )
+app.add_middleware(ProxyHeadersMiddleware, trusted_hosts="*")
```
kazet: Let's make trusted_hosts configurable in order not to introduce a vulnerability - maybe a variable that if is not set then ProxyHeadersMiddleware is not added?

And if we do that, let's add a mention in the docs (e.g. Troubleshooting section)

kazet: BTW, with this change, the static file URLs in Artemis will have proper protocol. Could you also document how to set proper hostnames when Artemis is behind a reverse proxy?

vishnudathks:
Hi @kazet! Updated the PR with the following changes:

- Made `TRUSTED_PROXY_HOSTS` configurable via an environment variable — `ProxyHeadersMiddleware` is only added when this variable is set, avoiding introducing a vulnerability when not needed
- Added `# type: ignore[arg-type]` for the mypy incompatibility with `ProxyHeadersMiddleware`

Still need to add the docs update (Troubleshooting section + reverse proxy hostname guide) — will push that shortly. Would appreciate a re-review when ready!

PR #2839, opened 2026-06-12
deduplicate original host for listed modules

kadewu:
As per title, some modules don't require data regarding original_domain/original_ip, therfore we can deduplicate them.

### DIFF
--- a/artemis/db.py
+++ b/artemis/db.py
@@ -664,6 +664,11 @@ class DB:
             del task_as_dict["payload"]["last_domain"]
         if "created_at" in task_as_dict["payload"]:
             del task_as_dict["payload"]["created_at"]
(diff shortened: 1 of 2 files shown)


artemis/db.py
```diff
@@ -664,6 +664,11 @@ def _get_task_deduplication_data(self, task: Task) -> str:
             del task_as_dict["payload"]["last_domain"]
         if "created_at" in task_as_dict["payload"]:
             del task_as_dict["payload"]["created_at"]
+        if task.receiver in Config.Miscellaneous.MODULES_WHITELIST_FOR_ORIGINAL_HOST_DEDUPLICATION:
```
kazet, line 667: These tasks are unrouted - i.e. they contain type, but the receiver is always None. It's Karton who will decide, based on type, what kartons to route the tasks to. So this idea won't work.

kadewu:
closing on favor of https://github.com/CERT-Polska/Artemis/pull/2921

PR #2842, opened 2026-06-13
Add CVE lookup for detected web technologies

imraflip:
This PR adds a `cve_lookup` module that maps the web technologies Artemis detects to known CVEs from the NVD. To feed it, `run_tech_detection` now returns structured `Technology` objects (name, version, CPE, categories) instead of plain tag strings, the Wappalyzer wrapper emits the CPE via `FingerprintWithInfo` - and the Nuclei router keeps its old string-based behaviour through a small `to_tag_strings` helper.

Two NVD quirks drove the implementation. The CPE Wappalyzer returns usually has a wildcard version, so when the version is known I rewrite that slot to query the specific version. And NVD's `cpeName` query does range matching ("version X and earlier"), which means a result really means "known CVEs exist for this product/version, verify the patch level" rather than confirmed-exploitable - the report wording reflects that. I also filter to `vulnerable: true` so CPEs that are merely the platform a vulnerable app runs on aren't reported as the technology's own vulnerabilities. Queries are throttled to NVD's unauthenticated limit (5 per 30s, raisable with an API key) and cached.


artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,230 @@
...
+
+        url = f"{NVD_API_URL}?{urlencode({'cpeName': cpe})}"
+        try:
+            response = http_requests.get(
+                url,
+                requests_per_second=Config.Modules.CveLookup.CVE_LOOKUP_NVD_REQUESTS_PER_SECOND,
+                max_size=NVD_RESPONSE_MAX_BYTES,
+                timeout=NVD_REQUEST_TIMEOUT_SECONDS,
+            )
+        except Exception as e:
```
kadewu: I'm not sure about hiding this behind only warning log.
As this is critical component of the module to work, I think it's worth to save the task as error.
Or at least not save it as 'OK' 'no cve found'

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,230 @@
...
+            return []
+
+        try:
+            payload = response.json()
+        except ValueError:
+            self.log.warning(f"NVD returned non-JSON for cpe={cpe}")
+            return []
+
+        cves = _extract_cves(payload, _cpe_product_key(cpe))
+        self.cache.set(cpe, json.dumps(cves).encode("utf-8"), timeout=NVD_CACHE_TTL_SECONDS)
```
kadewu, line 325: We can also introduce cache for failed lookups - but with shorter TTL

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,230 @@
...
+NVD_RESPONSE_MAX_BYTES = 5 * 1024 * 1024
+NVD_CACHE_TTL_SECONDS = 24 * 60 * 60
+# NVD's public API is frequently slow to first byte, so it needs a longer timeout than the
+# default used for scanning targets directly.
+NVD_REQUEST_TIMEOUT_SECONDS = 30
+
+
+def _fill_cpe_version(cpe: str, version: Optional[str]) -> str:
+    """
+    Wappalyzer keeps the version slot as ``*`` even when the version is known.
```
kadewu: Do we know the reason for that?
Is it possible that wappalyzer returns behind `version` sth like 'latest' and now we try to send it that way?

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,230 @@
...
+
+def _cpe_product_key(cpe: str) -> str:
+    """Return the ``part:vendor:product`` portion of a cpe:2.3 URI (e.g. ``a:apache:http_server``)."""
+    parts = cpe.split(":")
+    if len(parts) < 5:
+        return ""
+    return ":".join(parts[2:5])
+
+
+def _is_product_vulnerable(cve: Dict[str, Any], product_key: str) -> bool:
```
kadewu, line 121: I believe this check is too shallow.
e.g: The NVD configurations can have multiple nodes with operator 'AND'

example: https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=CVE-2021-44228

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,230 @@
...
+                if isinstance(decoded, list):
+                    return decoded
+            except (ValueError, UnicodeDecodeError):
+                pass  # fall through to a fresh fetch
+
+        url = f"{NVD_API_URL}?{urlencode({'cpeName': cpe})}"
+        try:
+            response = http_requests.get(
+                url,
+                requests_per_second=Config.Modules.CveLookup.CVE_LOOKUP_NVD_REQUESTS_PER_SECOND,
```
kadewu: Throttle request works only if we would not scale that module.
We should either fix that now, or leave it but have a good comment about that.

I doubt that we will scale that, so I'm fine with just leaving comment.

artemis/web_technology_identification.py
```diff
@@ -33,7 +73,24 @@ def run_tech_detection(urls: List[str], logger: logging.Logger) -> Any:
...
 
-        return json.loads(wappalyzer_output)
+        raw = json.loads(wappalyzer_output)
+        parsed: Dict[str, List[Technology]] = {}
+        for url, items in raw.items():
+            if not isinstance(items, list):
+                parsed[url] = []
+                continue
+            parsed[url] = [_parse_tech(item) for item in items if isinstance(item, dict)]
+        return parsed
```
kadewu, line 84: Are all `urls` provided to this method always present in `raw` response?
If not, we will not return url -> list as per comment said.

artemis/modules/cve_lookup.py
imraflip: I have split the two cases so a failed lookup doesn't read as a clean scan, so it's saved as `ERROR` instead of `OK` now

artemis/modules/cve_lookup.py
imraflip: Done, failed lookups now get a short 5 min cache under the CPE key, so we're not re-hitting NVD during an outage

artemis/web_technology_identification.py
imraflip: They're not always all there actually. Wappalyzer can quietly drop a URL it found nothing on, so I now pre-seed the result with every input URL (`{url: [] for url in urls}`) and lay its output on top, which keeps the url -> list contract intact

artemis/modules/cve_lookup.py
imraflip: And for this I reworked it to actually walk the config tree and respect `negate` at both the config and node level, so for the CVE-2021-44228 AND case we only flag our product when it's genuinely the vulnerable one and not the platform it runs on. I also added tests for the negated and AND shapes too

artemis/modules/cve_lookup.py
imraflip: It's by design in `wappalyzergo`, not a bug. It returns the version separately from a static template CPE that keeps `*` in the version slot, so it never injects "latest" or anything else into the CPE itself

The version it detects is the separate piece, and that's the one we drop into the CPE ourselves, so that's where a value like "latest" could realistically sneak in. To cover exactly that, `_fill_cpe_version` only fills the version in when it looks like a real one; anything else and we keep the `*`, so we never send junk to NVD

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,287 @@
...
+        skips re-querying NVD for this CPE until the short failure TTL expires.
+        """
+        self.cache.set(
+            cpe,
+            json.dumps({NVD_FAILURE_MARKER: True}).encode("utf-8"),
+            timeout=NVD_FAILURE_CACHE_TTL_SECONDS,
+        )
+
+    def _query_nvd(self, cpe: str) -> Optional[List[Dict[str, Any]]]:
+        # NVD is a slow, rate-limited public API (~5 requests / 30s without a key). We deliberately
```
kadewu: I find this comment a little bit too long, can we make it as comprehensive as possible?

Also it's most likely not the best place to put comment about scaling, it should be in place where we will scale it.

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,287 @@
...
+        self.db.save_task_result(
+            task=current_task,
+            status=status,
+            status_reason=status_reason,
+            data={"url": url, "findings": findings},
+        )
+
+
+if __name__ == "__main__":
+    CveLookup.parallel_loop()
```
kadewu: It shouldn't be parallel_loop cause of the scaling problem (and here we can place the comment)

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,287 @@
...
+                decoded = json.loads(cached.decode("utf-8"))
+            except (ValueError, UnicodeDecodeError):
+                decoded = None
+            if isinstance(decoded, dict) and decoded.get(NVD_FAILURE_MARKER):
+                return None  # recently failed - don't re-hammer NVD until the short TTL expires
+            if isinstance(decoded, list):
+                return decoded
+            # anything else: fall through to a fresh fetch
+
+        url = f"{NVD_API_URL}?{urlencode({'cpeName': cpe})}"
```
kadewu: I can see that CVE API mentions pagination. Shouldn't we cover this as well? E.g when version is not present?

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,287 @@
...
+
+        description = ""
+        for desc in cve.get("descriptions") or []:
+            if isinstance(desc, dict) and desc.get("lang") == "en":
+                description = str(desc.get("value", ""))
+                break
+
+        cvss_score: Optional[float] = None
+        metrics = cve.get("metrics") or {}
+        for key in ("cvssMetricV31", "cvssMetricV30", "cvssMetricV2"):
```
kadewu: What about v4? If someone will not publish v3 then we may miss that

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,287 @@
...
+            metric_entries = metrics.get(key)
+            if isinstance(metric_entries, list) and metric_entries:
+                first = metric_entries[0]
```
kadewu: We are only extracting first, but maybe we should get the highest one instead?
Or focus on NVD provided?

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,378 @@
...
+        return False
+    return any(isinstance(match, dict) and match.get("vulnerable") for match in node.get("cpeMatch") or [])
+
+
+def _configuration_implicates_product(configuration: Dict[str, Any], product_key: str) -> bool:
+    """
+    Whether ``product_key`` is a vulnerable component of this single configuration.
+
+    A configuration joins its sibling nodes with an ``operator``. Ignoring that operator is
+    what made the earlier check too shallow: under ``AND`` the CVE only applies when *every*
```
kadewu: Let's refrain from referencing the 'github discussion'. Comment should be understandable by a person that will read it in the future.

artemis/config.py
```diff
@@ -402,6 +402,23 @@ class Crawling:
                 "retry sooner.",
             ] = get_config("KATANA_TIMEOUT_CACHE_TTL_SECONDS", default=60 * 60, cast=int)
 
+        class CveLookup:
+            CVE_LOOKUP_NVD_REQUESTS_PER_SECOND: Annotated[
+                float,
+                "Rate limit for NVD API queries. NVD allows 5 requests per 30 seconds without an API key "
+                "(~0.166 r/s); raise this if you configure a key with NIST.",
```
kadewu: The comment about key can be removed, as we not support it right?

artemis/modules/cve_lookup.py
```diff
@@ -0,0 +1,378 @@
...
+                return decoded
+            # anything else: fall through to a fresh fetch
+
+        product_key = _cpe_product_key(cpe)
+        api_url = Config.Modules.CveLookup.CVE_LOOKUP_NVD_API_URL
+        cves: List[Dict[str, Any]] = []
+        start_index = 0
+        for _ in range(NVD_MAX_PAGES):
+            url = f"{api_url}?{urlencode({'cpeName': cpe, 'startIndex': start_index})}"
+            try:
```
kadewu, line 288: In such scenario:
- 1 page finds something
- 2 page fails

we cache whole response as fail, I think it should be okay to return what we found?

kadewu (commented):
a couple more small things from me

kadewu (dismissed):

PR #2844, opened 2026-06-15
Build all go tools with one go 1.25 toolchain

imraflip:
The image used to include two Go toolchains: 1.24 for most of the tools and a second 1.25 just for katana. I removed the second one. I bumped nuclei to v3.9.0, which moves the dependency graph onto a version of bytedance/sonic that works with Go 1.25, and that let me add katana to docker/go.mod and build everything with one Go 1.25 toolchain.


kadewu:
Bumping nuclei currently is on hold due too investigation why we see such spike in time.
(https://github.com/CERT-Polska/Artemis/pull/2684)

kadewu:
Nuclei got updated on main branch - this PR can be rebased against main branch.

imraflip:
got it, on it, sir

kazet (approved):

PR #2857, opened 2026-06-18
improve memory perfromance in archiver

kadewu:


artemis/autoarchiver/autoarchiver.py
```diff
@@ -69,20 +97,16 @@ def archive_old_results(interesting: bool) -> None:
...
-        max_length=Config.Data.Autoarchiver.AUTOARCHIVER_PACK_SIZE,
-        interesting=interesting,
+    _save_and_delete_items(
+        db.iter_oldest_task_results_before(
+            time_to=datetime.datetime.now() - archive_age_timedelta,
+            max_length=Config.Data.Autoarchiver.AUTOARCHIVER_PACK_SIZE,
+            interesting=interesting,
+        ),
+        "_interesting" if interesting else "_not_interesting",
+        min_count=Config.Data.Autoarchiver.AUTOARCHIVER_PACK_SIZE,
```
kazet: would you write a test for that?

artemis/autoarchiver/autoarchiver.py
```diff
@@ -15,16 +15,49 @@
...
+
+    if not ids:
+        os.remove(temp_path)
         LOGGER.info("Nothing to save")
-        return
+        return 0
 
-    date_from = old_items[0]["created_at"]
-    date_to = old_items[-1]["created_at"]
+    if len(ids) < min_count:
```
kazet: Here we write the file to disk and immediately remove it - it's not optimal, IMO counting using one SQL query and saving using the second would be cleaner

test/unit/test_autoarchiver.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class TestAutoArchiverWithDB(unittest.TestCase):
+    def setUp(self) -> None:
+        self.db = DB()
+        self.tmpdir = tempfile.mkdtemp()
+        self.tag = f"autoarchiver_test_{uuid.uuid4().hex}"
+
+    def tearDown(self) -> None:
+        import shutil
```
kazet: why do we import shutil inside the method (here and in multiple other places) and not at the top?

test/unit/test_autoarchiver.py
```diff
@@ -0,0 +1,212 @@
...
+            raise RuntimeError("simulated DB failure mid-stream")
+
+        with patch("artemis.autoarchiver.autoarchiver.db"), patch(
+            "artemis.autoarchiver.autoarchiver.Config"
+        ) as mock_cfg:
+            mock_cfg.Data.Autoarchiver.AUTOARCHIVER_OUTPUT_PATH = self.tmpdir
+            with self.assertRaises(RuntimeError):
+                _save_and_delete_items(bad_iterator(), "_test")
+
+        self.assertEqual(os.listdir(self.tmpdir), [])
```
kazet, line 94: Let's also assert that no items are removed in this case

test/unit/test_autoarchiver.py
```diff
@@ -0,0 +1,212 @@
...
+
+        with self._patch_config(pack_size=2, min_age_seconds=1):
+            archive_old_results(interesting=True)
+
+        remaining = _ids_in_db_with_tag(self.db, self.tag)
+        self.assertNotIn(interesting_id_1, remaining)
+        self.assertNotIn(interesting_id_2, remaining)
+        self.assertIn(ok_id_1, remaining)
+
+    def test_archive_old_results_only_archives_matching_not_interesting_flag(self) -> None:
```
kazet, line 176: I would also check that the file gets created on disk - and what are its contents

artemis/autoarchiver/autoarchiver.py
```diff
@@ -15,16 +15,49 @@
...
+    date_from: datetime.datetime | None = None
+    date_to: datetime.datetime | None = None
+
+    try:
+        with gzip.open(temp_path, "wt", encoding="utf-8") as f:
+            f.write("[\n")
+            for i, item in enumerate(items):
+                if date_from is None:
+                    date_from = item["created_at"]
+                date_to = item["created_at"]
```
kazet: here we assume that the dates increase monotonically - this is probably always the case, but can we write this more defensively?

kadewu: It was sorted in db query, but I guess it won't hurt to add a check.

test/unit/test_autoarchiver.py
kadewu: It should not be needed here I will move it out.
I remember that couple years ago I needed to import inside teardown, and just automatically done that.

artemis/autoarchiver/autoarchiver.py
```diff
@@ -15,16 +15,40 @@
...
+    date_from: datetime.datetime | None = None
+    date_to: datetime.datetime | None = None
+
+    try:
+        with gzip.open(temp_path, "wt", encoding="utf-8") as f:
+            f.write("[\n")
+            for i, item in enumerate(items):
+                if date_from is None:
+                    date_from = item["created_at"]
+                if date_to is None or date_to > item["created_at"]:
```
kazet: is the comparison good?

kadewu: good catch

artemis/autoarchiver/autoarchiver.py
```diff
@@ -15,16 +15,40 @@
...
+            f.write("\n]")
+    except Exception:
+        if os.path.exists(temp_path):
+            os.remove(temp_path)
+        raise
 
-    date_from = old_items[0]["created_at"]
-    date_to = old_items[-1]["created_at"]
+    if date_from is None or date_to is None:
+        LOGGER.warning("Coudln't properly extract date range.")
```
kazet:
```suggestion
        LOGGER.warning("Couldn't properly extract date range.")
```

PR #2864, opened 2026-06-21
Removing template blocklist as the FPs have been fixed

kazet:


artemis/config.py
```diff
@@ -641,10 +641,7 @@ class Nuclei:
                         "http/cves/2025/CVE-2025-49113.yaml",
                         "http/cves/2024/CVE-2024-42009.yaml",
                         # Too many FPs
-                        "http/exposed-panels/aveva-intouch-access-anywhere-panel.yaml",
```
kadewu, line 644: I think we should still leave that one
\- "{{BaseURL}}/AccessAnywhere/start.html"
and matching 
\- 'contains_any(body, "Access Anywhere", "AccessAnywhere")'

will produce FPs if url is found somewhere in html body.

But I think we can unblock janitza panel

kazet: But now it's with an `and` operator and also  'contains_all(body, "AppName", "OnClickConnect", "ericom", "InTouch")' must be matched, or am I misreading the code?

kadewu: But finding 'AppName' or 'OnClickConnect' is not that uncommon.

I was checking this version of template and it still did produce FP due to above, but maybe I will recheck that.

kazet: Let's recheck as I think `contains_all` should match all of them - or maybe they use it in a wrong way in this template?

kadewu: I probably run nuclei with the older version of that template - the previous sites are not matching.

kadewu (approved):

PR #2865, opened 2026-06-21
[WIP] [DO NOT MERGE] Running Nuclei on non-HTTP ports as well

snicket2100:
Two modules got changed in order to achieve it:

1. `nuclei-router` now creates `nuclei-target` tasks for non-HTTP services as well. These `nuclei-target` tasks contain a new field in the payload (`"nuclei-routing-scan-mode"`) which can be either `http` or `other`. We default to `http` for existing tasks, for backwards compatibility reason.

2. When `nuclei-routing-scan-mode = http`, Nuclei runs the way it used to. For `nuclei-routing-scan-mode = other`, we add `-ept http` so that we don't run tests targeting the HTTP(S) protocol - as they take a lot of time - plus we skip some features  which only make sense for HTTP (`-dast`, crawling for links, etc.). Also the initial WAF check doesn't happen for non-HTTP Nuclei runs. I have also updated the grouping logic (`get_batch_group_key()`) so that `http` and `other` tasks are processed in separate groups.

The goal was to keep the behaviour for HTTP ports unchanged so that we don't accidentaly break something. There was some change in how the findings are assigned back to the tasks they belong to, but that should be backwards compatible.

This pull request looks super long, but most of it is just changed indentation which gets displayed in a messy way by GitHub unfortunately.

---

image

(there is plenty of redundant templates matching for Redis and PostgreSQL here, I just haven't ignored them yet)

### DIFF
--- a/artemis/modules/nuclei.py
+++ b/artemis/modules/nuclei.py
@@ -26,7 +26,7 @@ from artemis.crawling import (
 )
 from artemis.module_base import ArtemisBase
 from artemis.modules.data.static_extensions import STATIC_EXTENSIONS
-from artemis.modules.nuclei_router import NUCLEI_ROUTER_FLAGS_PAYLOAD_KEY
+from artemis.modules.nuclei_router import NUCLEI_ROUTER_FLAGS_PAYLOAD_KEY, NUCLEI_ROUTER_SCAN_MODE_KEY, NUCLEI_ROUTER_SCAN_MODE_VALUE_HTTP
 from artemis.modules.runtime_configuration.nuclei_configuration import (
     NucleiConfiguration,
     SeverityThreshold,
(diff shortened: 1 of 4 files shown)


artemis/modules/nuclei.py
```diff
@@ -734,98 +753,44 @@ def run_multiple(self, tasks: List[Task]) -> None:
...
+            for task in tasks:
+                param_url = get_target_url(task)
+                for _, template_data in DAST_SCANNING.items():
+                    param_url = add_injectable_params_and_common_params_from_wordlist(
+                        param_url, template_data["params_wordlist"], template_data["param_default_value"]
+                    )
+                dast_targets.append(param_url)
 
+            # Running all dast templates at once on all dast targets constructed
             all_dast_templates = []
```
kadewu, line 786: The PR may appear long cause git is not handling well this code duplication.
Actually even the duplicated code is a dead code, I've pushed separate pr to remove it, it should clear the diff.

artemis/modules/nuclei_router.py
```diff
@@ -11,6 +11,9 @@
 
 TECHNOLOGY_DETECTION_TAGS_TO_EXCLUDE = {"wordpress": ["wordpress"]}
 NUCLEI_ROUTER_FLAGS_PAYLOAD_KEY = "nuclei-routing-additional-flags"
+NUCLEI_ROUTER_SCAN_MODE_KEY = "nuclei-routing-scan-mode"
+NUCLEI_ROUTER_SCAN_MODE_VALUE_HTTP = "http"
```
kadewu, line 15: I know it's WP, but definitely let's make it an ENUM (http, other), maybe we will be distinguishing more in future.

kadewu:
Closing due to inactivity.

PR #2883, opened 2026-06-24
Fix false positives from flaky services and save the probe URLs

imraflip:
A flaky service can return a slightly different response that has nothing to do with ORM injection, which is enough to trip the detector into a false finding. To avoid that, each check now runs a few times (3 by default, set by `ORM_INJECTION_NUM_CONFIRMATIONS`) and only counts as a finding if the same parameter keeps showing a difference every time, so a one-off fluke gets dropped. Checks that don't match still stop after the first run, so it doesn't add extra requests for normal cases.

I also added the two URLs we actually tested (the matching one and the baseline) to each result under `tested_urls`, so it's easy to see what triggered the finding and reproduce it.

### DIFF
--- a/artemis/config.py
+++ b/artemis/config.py
@@ -1181,7 +1181,7 @@ class Config:
                 "How many times a differential response must reproduce before it is reported as a finding. "
                 "Guards against flaky services where a one-off difference is not actually caused by ORM "
                 "injection.",
-            ] = get_config("ORM_INJECTION_NUM_CONFIRMATIONS", default=3, cast=int)
+            ] = get_config("ORM_INJECTION_NUM_CONFIRMATIONS", default=10, cast=int)
 
         class SqlInjectionDetector:
             SQL_INJECTION_STOP_ON_FIRST_MATCH: Annotated[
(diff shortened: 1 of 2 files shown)


artemis/config.py
```diff
@@ -1171,6 +1176,12 @@ class OrmInjectionDetector:
                 bool,
                 "Whether to stop scanning after the first ORM injection finding.",
             ] = get_config("ORM_INJECTION_STOP_ON_FIRST_MATCH", default=True, cast=bool)
+            ORM_INJECTION_NUM_CONFIRMATIONS: Annotated[
+                int,
+                "How many times a differential response must reproduce before it is reported as a finding. "
+                "Guards against flaky services where a one-off difference is not actually caused by ORM "
+                "injection.",
+            ] = get_config("ORM_INJECTION_NUM_CONFIRMATIONS", default=3, cast=int)
```
kazet:
```suggestion
            ] = get_config("ORM_INJECTION_NUM_CONFIRMATIONS", default=10, cast=int)
```

kazet (approved):

PR #2899, opened 2026-06-29
Implement direct URL scanning

imraflip:
This change teaches the classifier to accept a root URL such as `https://example.com/` or `ssh://example.com:22/`. When it sees one, it reads the scheme, host and port and emits a `TaskType.SERVICE` task directly, the same task `port_scanner` would eventually produce - so every service-consuming module (`sql_injection_detector`, `nuclei`, `lfi_detector`, etc) picks it up immediately with no changes on their side. The scheme selects the service and the SSL flag, and the default port is resolved with `socket.getservbyname()`, the same way the reporting code already derives ports from schemes, so no port numbers are hardcoded.

Only root URLs are accepted. A URL carrying a path, query, fragment or userinfo is rejected with a clear validation error rather than being silently stripped, because a root URL identifies a service to scan rather than a specific resource, and the HTTP modules crawl from the root on their own. Because a root URL already yields a SERVICE task without port scanning, the add-targets form no longer forces `port_scanner` to be enabled when every submitted target is a root URL. Mixed submissions that still contain a bare domain or IP continue to require it, since those targets rely on the recon path.

### DIFF
--- a/artemis/frontend.py
+++ b/artemis/frontend.py
@@ -28,7 +28,7 @@ from karton.core.task import TaskPriority, TaskState
 from redis import Redis
 from starlette.datastructures import Headers
 
-from artemis import auth, csrf
+from artemis import auth, csrf, direct_url_scanning
 from artemis.binds import TaskType
 from artemis.config import Config
 from artemis.db import DB, ColumnOrdering, ReportGenerationTaskStatus, TaskFilter
@@ -216,7 +216,7 @@ async def post_add(
 
     # When every target is a root URL, the classifier emits the SERVICE task directly
     # from the scheme, so port_scanner isn't needed to reach SERVICE/WEBAPP modules.
-    all_targets_are_urls = bool(total_list) and all(Classifier.is_url(task) for task in total_list)
+    all_targets_are_urls = bool(total_list) and all(direct_url_scanning.is_scannable_url(task) for task in total_list)
 
     for dependency, task_types in [
         ("port_scanner", [TaskType.SERVICE.value, TaskType.WEBAPP.value]),
--- a/artemis/frontend.py
+++ b/artemis/frontend.py
@@ -211,13 +211,19 @@ async def post_add(
     for task in total_list:
         if not Classifier.is_supported(task):
             validation_messages.append(
-                f"{task} is not supported - Artemis supports domains, IPs or IPv4 ranges. Domains and IPs may also optionally be followed by port number."
+                f"{task} is not supported - Artemis supports domains, IPs, IPv4 ranges or root URLs such as https://example.com/ or ssh://example.com:22/. Domains and IPs may also optionally be followed by port number. A URL must be a root URL (scheme://host[:port]/) without a path."
             )
+
+    # When every target is a root URL, the classifier emits the SERVICE task directly
+    # from the scheme, so port_scanner isn't needed to reach SERVICE/WEBAPP modules.
+    all_targets_are_urls = bool(total_list) and all(Classifier.is_url(task) for task in total_list)
 
     for dependency, task_types in [
         ("port_scanner", [TaskType.SERVICE.value, TaskType.WEBAPP.value]),
         ("device_identifier", [TaskType.DEVICE.value]),
     ]:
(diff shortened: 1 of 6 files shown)


artemis/modules/classifier.py
```diff
@@ -17,6 +19,34 @@
 ASN_REGEX = "[aA][sS][0-9][0-9]*"
 
 
+class ParsedURL(NamedTuple):
```
kadewu: So this class would be moved to the mentioned file

artemis/modules/classifier.py
```diff
@@ -83,8 +113,63 @@ def _clean_ipv6_brackets(data: str) -> str:
             data = data[1:-1]
         return data
 
+    @staticmethod
+    def _parse_url(data: str) -> Optional[ParsedURL]:
```
kadewu: and this method

artemis/modules/classifier.py
```diff
@@ -109,6 +194,11 @@ def is_supported(data: str) -> bool:
         data = Classifier._clean_ipv6_brackets(data)
         return Classifier._is_ip_or_domain(data)
 
+    @staticmethod
+    def is_url(data: str) -> bool:
```
kadewu: this method is not exactly checking if something is url, rather 'is this url accepted for direct scanning feature?' let's rename it 
probably can be also moved outside this class, wdyt?

artemis/modules/classifier.py
```diff
@@ -143,6 +258,15 @@ def run(self, current_task: Task) -> None:
             )
             return
 
+        parsed_url = Classifier._parse_url(data)
```
kadewu:
```suggestion
        direct_scanning_url = Classifier._parse_url(data)
```

test/modules/test_classifier.py
```diff
@@ -29,9 +30,22 @@ def test_skipping_public_suffixes(self) -> None:
         self.assertTasksEqual(results, [])
 
     def test_support(self) -> None:
```
kadewu, line 33: let's also add some invalid ones like 'http:/test.com'

kadewu (commented):
Small sugestion, which I believe would make the code be easier to read.
Cause fingerprinting etc. is crucial part of classifier, sudden appearance of scheme mapping etc. makes code of this module little more confusing imo.
Can we extract all the utility to single file 'direct_url_scanning_utility.py' (just randon name I came up with)

more in details in comments

You can also add small information about that to our documentation.

imraflip:
Moved the scheme->service mapping and the URL parsing out of the classifier into a new `artemis/direct_url_scanning.py` so it stays focused on fingerprinting like you suggested. `is_url` is now `is_scannable_url` and lives outside the class, since it's really asking whether the URL is accepted for direct scanning rather than whether it's a URL at all, and I renamed the local `parsed_url` to `direct_scanning_url` too. Also added the malformed-separator cases (`http:/cert.pl`, `http:cert.pl`) to the tests, plus a short note about direct URL scanning in the quick-start docs

kadewu (dismissed):

artemis/frontend.py
```diff
@@ -211,13 +211,19 @@ async def post_add(
     for task in total_list:
         if not Classifier.is_supported(task):
             validation_messages.append(
-                f"{task} is not supported - Artemis supports domains, IPs or IPv4 ranges. Domains and IPs may also optionally be followed by port number."
+                f"{task} is not supported - Artemis supports domains, IPs, IPv4 ranges or root URLs such as https://example.com/ or ssh://example.com:22/. Domains and IPs may also optionally be followed by port number. A URL must be a root URL (scheme://host[:port]/) without a path."
             )
 
+    # When every target is a root URL, the classifier emits the SERVICE task directly
+    # from the scheme, so port_scanner isn't needed to reach SERVICE/WEBAPP modules.
+    all_targets_are_urls = bool(total_list) and all(direct_url_scanning.is_scannable_url(task) for task in total_list)
```
kazet: We can also disable the port scanner check if all targets have ports (e.g. example.com:22)

kazet (approved):

PR #2902, opened 2026-07-01
Shorten Nuclei DAST PoC URLs via single-mode re-fuzzing

X-R3Y24:
## What
Shorten Nuclei DAST PoC URLs by re-fuzzing confirmed findings in `-fuzzing-mode single`, keeping only the parameters that reproduce the vulnerability on their own and falling back to the full URL when single-mode confirms nothing.

## Why
DAST targets are built by stacking every parameter from all six wordlists (ssrf/redirect/lfi/cmdi/sqli/xss) onto one URL — 116 parameters. The scan runs `-fuzzing-mode multiple`, so Nuclei fuzzes all of them in a single request and the resulting `matched-at` carries the payload across many parameters at once (e.g. 51/116 for a redirect hit, the rest being scaffolding at their default values). From a multiple-mode hit there is no way to tell which parameter is actually vulnerable — that is the root cause of the very long PoC URLs. The old `minimize_nuclei_matched_at_url` relied on `fuzzing_parameter`, which multiple-mode leaves empty/ambiguous, so it never minimized and emitted the full URL.
`multiple` mode itself stays (intentional, for scan performance).

## Fix
Re-fuzz each multi-parameter finding in `-fuzzing-mode single` (one parameter per request — verified against nuclei source `pkg/fuzz/parts.go` and on a live v3.9.0 instance). Single-mode reports a non-empty `fuzzing_parameter` per hit; we collect those and keep exactly the confirmed parameters, dropping the rest.
Key design decisions (happy to discuss any of these):
- **Re-fuzz on default values, not payloads.** Injected (wordlist) parameters are reset to their family default from `DAST_SCANNING`, so Nuclei re-fuzzes from a clean base — same as the original scan — rather than re-fuzzing a value that already carries a payload. Site parameters not in any wordlist keep their `matched-at` value.
- **Payloads preserved byte-for-byte.** The query is split manually (no `parse_qs`/`urlencode` round-trip), because `%2F%2F` must not decay to `//` — an encoded-slash filter-bypass payload has to survive intact.
- **interactsh is passed to the re-fuzz** when configured, otherwise OOB findings (e.g. open-redirect via oast.me) would not reproduce in single mode.
- **The re-fuzz command mirrors the main scan's flags** (user agent, rate limit, timeout, system resolvers, response-size read) so the finding is reproduced under the same conditions the original scan ran in.
- **Fallback to the full PoC** when single confirms nothing — the PoC still works, just stays long.

## Tests
- `test_multiple_confirmed_all_kept` / `test_single_confirmed_only_that_one` — the URL is reduced to exactly the parameters single-mode confirms (one or several).
- `test_nothing_confirmed_falls_back_to_full` / `test_no_refuzz_fn_returns_full` — the fallback: an empty confirmation set (or absent re-fuzz) yields the original URL unchanged, so the PoC never breaks.
- `test_payload_preserved_byte_for_byte` / `test_equals_sign_in_value_preserved` / `test_param_without_value_preserved` / `test_original_order_preserved` — the byte-for-byte guarantee.
- `test_few_params_not_minimized` / `test_no_query_returned_unchanged` / `test_confirmed_name_absent_from_url_ignored` — edge cases.
- Manually verified single-mode against a live target: single isolates the vulnerable parameter and emits `fuzzing_parameter` correctly on v3.9.0.

## Notes / Out of scope
- **Open question for review:** combinational vulnerabilities (needing several parameters together) can't be reproduced one-at-a-time, so they hit the fallback and keep the full PoC. Rare for the reflected-XSS/open-redirect families that dominate long PoCs. I added a fallback log (without the URL) so we can measure how often this actually happens; **ablation** (greedy minimal-subset search) would handle these but at N× the nuclei cost — I'd propose deciding on it after ~1 week of observing the fallback rate rather than building it now. Open to doing it differently.
- **Command flags are currently duplicated** between `_scan` and the re-fuzz. This PR mirrors the relevant flags by hand (user agent, rate limit, timeout, resolvers, response-size); batch-only flags (rate-limit-duration, bulk-size, concurrency, stats/trace) are intentionally omitted as meaningless for a single-URL re-fuzz. The cleaner fix is a shared command-builder used by both paths so flags can't drift again — kept out of this PR to avoid refactoring the production scan path here, planned as a follow-up.
- Did **not** touch `multiple` scanning mode.
- Removed the now-dead `_verify_with_nuclei` (replaced by `_refuzz_single_with_nuclei`).

### DIFF
--- a/artemis/modules/nuclei.py
+++ b/artemis/modules/nuclei.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import collections
 import enum
+import functools
 import itertools
 import json
 import logging
@@ -11,7 +12,7 @@ import subprocess
 import time
 import urllib
 from statistics import StatisticsError, quantiles
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Set
 
 import more_itertools
 from karton.core import Task
@@ -106,38 +107,6 @@ METRIC_SCAN_DURATION = Histogram(
 logger = logging.getLogger(__name__)
 
 
-def _verify_with_nuclei(url: str, template_path: str) -> bool:
-    try:
-        command = [
-            "nuclei",
-            "-disable-update-check",
-            "-u",
-            url,
-            "-t",
-            template_path,
-            "-dast",
-            "-fuzzing-mode",
-            "multiple",
-            "-jsonl",
-            "-silent",
-        ]
-        result = subprocess.check_output(command, stderr=subprocess.DEVNULL)
-        for line in result.strip().splitlines():
-            try:
-                hit = json.loads(line)
-                if hit.get("matched-at"):
-                    return True
-            except json.JSONDecodeError:
-                continue
-        return False
-    except FileNotFoundError:
-        logger.warning("Nuclei binary not found, skipping URL verification")
-        return False
-    except subprocess.CalledProcessError as e:
-        logger.warning("Nuclei verification failed on %s: %s", url, e)
-        return False
-
-
 # It is important to keep ssrf, redirect and lfi at the top so that their params get the correct default values
 DAST_SCANNING: Dict[str, Dict[str, Any]] = {
     "ssrf": {  # ssrf dast templates work only when the param is of the form http://...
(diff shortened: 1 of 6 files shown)


artemis/modules/nuclei.py
```diff
@@ -169,6 +138,94 @@ def _verify_with_nuclei(url: str, template_path: str) -> bool:
...
+            "-u",
+            refuzz_url,
+            "-t",
+            template_path,
+            "-dast",
+            "-fuzzing-mode",
+            "single",
+            "-jsonl",
+            "-silent",
+        ]
```
kazet: Let's share as much of the commandline as possible -e.g. in this case we forget user agent or rate limit

X-R3Y24: Done. The re-fuzz now mirrors the main scan's flags (user agent, rate limit, timeout, resolvers, response-size read), verified on a live target. Batch-only flags (rate-limit-duration, bulk-size, etc.) omitted as they're meaningless for a single-URL re-fuzz. Left the shared command-builder as a follow-up rather than refactoring the scan path here,  noted in the description.

artemis/modules/nuclei.py
```diff
@@ -169,6 +138,109 @@ def _verify_with_nuclei(url: str, template_path: str) -> bool:
...
+
+    # These flags mirror the main scan's command (_scan) so the re-fuzz
+    # reproduces the finding under the same conditions (user agent, rate
+    # limit, timeout, resolvers). Batch-only flags are intentionally
+    # omitted: rate-limit-duration/bulk-size/concurrency/stats are
+    # meaningless for a single-URL re-fuzz. Sharing a common command
+    # builder with _scan is a planned follow-up (see PR description).
+    try:
+        command = [
+            "nuclei",
```
kazet: I'd insist on sharing as much common flags (e.g. as a function, nothing fancy) as possible

test/unit/test_minimize_nuclei_matched_at_url.py
```diff
@@ -6,102 +6,65 @@
...
+    def test_no_refuzz_fn_returns_full(self) -> None:
+        result = minimize_nuclei_matched_at_url(self.URL, refuzz_fn=None)
+        self.assertEqual(result, self.URL)
+
+    def test_confirmed_name_absent_from_url_ignored(self) -> None:
+        result = minimize_nuclei_matched_at_url(self.URL, refuzz_fn=lambda _: {"next", "doesnotexist"})
+        self.assertEqual(_param_names(result), ["next"])
+
+    def test_few_params_not_minimized(self) -> None:
+        short = "http://ex.com/?a=1&b=2"
```
kazet: nit: example.com is IMO better

artemis/modules/nuclei.py
```diff
@@ -169,6 +138,109 @@ def _verify_with_nuclei(url: str, template_path: str) -> bool:
...
+            try:
+                hit = json.loads(line)
+            except json.JSONDecodeError:
+                continue
+            # NOTE: assumes single-mode hits carry a non-empty "fuzzing_parameter";
+            # verify against live Nuclei output.
+            param = hit.get("fuzzing_parameter")
+            if param:
+                confirmed.add(param)
+        return confirmed
```
kazet, line 258: here we take the fuzzing_parameter at face value, without checking that actually nuclei invocation with this single parameter returns the vuln - maybe if you add all other parameters it works, but if you omit them providing only the single one, it doesn't?

let's add an integration test for this one: if (isset($_GET["login"])) { echo "Login failed: " . $_GET["user"]; }

artemis/modules/nuclei.py
X-R3Y24: Done — extracted build_common_nuclei_command() with the flags shared by every invocation (user agent, timeout, jsonl, system-resolvers, rate-limit, response-size-read, interactsh). Both _scan and the single-mode re-fuzz now build on it, so they can't drift. Batch-only flags (-rate-limit-duration, -bulk-size, -concurrency, -stats-json, -trace-log) stay in _scan since they're meaningless for a single-URL re-fuzz.

test/unit/test_minimize_nuclei_matched_at_url.py
X-R3Y24: Done — replaced ex.com with example.com across the unit tests.

artemis/modules/nuclei.py
X-R3Y24: Good catch — this was a real gap, fixed. Single mode reports the parameter it mutated, but it keeps sending the others, so a gated finding (one that needs a second parameter) is still reported against a single name. minimize_nuclei_matched_at_url now re-fuzzes the shortened URL and only keeps it if the finding still reproduces without the parameters we dropped; otherwise it falls back to the full PoC. Cost is one extra single-mode nuclei run per finding.

Added your scenario as an integration test — test/data/php_xss_requiring_second_param/: search is reflected only when login is present, so single mode reports search alone, and the test asserts Artemis keeps login in the PoC instead of shortening to ?search=…. One deviation from your snippet: I used $_GET["search"]/$_GET["login"] rather than user, because both names have to be in dast_params/xss.txt or Artemis never puts them in the DAST target in the first place. Also covered at the unit level: test_reported_param_needing_a_gate_falls_back_to_full and test_verification_runs_on_the_shortened_url.

PR #2921, opened 2026-07-07
Refactor deduplication and add more specified deduplication

kadewu:
PR contains:
- refactor, which is moving the deduplication to the higher layer in terms of Karton, so we deduplicate routed tasks, not unrouted, this way we can have more module specific deduplication
- cherry-picked the commit from: https://github.com/CERT-Polska/Artemis/pull/2839
that needed module specific 

Additionaly while adding lock release this PR also adjusted lock releasing in WAF checking mechanism.


artemis/cleanup.py
```diff
@@ -118,12 +118,11 @@ def _cleanup_scheduled_tasks() -> None:
...
         for i in range(0, len(finished_analyses_ids), BATCH):
             analysis_ids = finished_analyses_ids[i : i + BATCH]
-            rows = db.delete_scheduled_tasks_for_analyses(analysis_ids)
-            removed_rows += rows
+            removed_rows += db.delete_module_processed_tasks_for_analyses(analysis_ids)
 
             logger.debug("Cleaned up ScheduledTask table for analyses: %s", ",".join(analysis_ids))
         logger.info(
-            "Removed %d rows in ScheduleTask table for %d finished analyses. "
+            "Removed %d rows in ModuleProcessedTask table for %d finished analyses. "
```
kazet: I would call it ModuleStartedTask and create it as soon as possible (preferably even before starting the task) to decrease the probability that we consume multiple identical tasksin parallel

artemis/config.py
```diff
@@ -331,6 +331,16 @@ class Miscellaneous:
             bool, "Raise error in case cleanup task did not found unfinished analyses."
         ] = get_config("CLEANUP_RAISE_ERROR_ON_NON_UNFINISHED_ANALYSES", default=False, cast=bool)
 
+        MODULES_WHITELIST_FOR_ORIGINAL_HOST_DEDUPLICATION: Annotated[
+            list[str],
+            "List of modules not depending on original_ip/original_domain. "
+            "The purpose of the list is solely for task deduplication. Can be extended with custom modules.",
```
kazet: hmm I think the docs may be made more verbose. For me it's not clear atm

PR #2924, opened 2026-07-09
Abort admin_panel_login_bruter scan when target returns HTTP 429

X-R3Y24:
## What

Abort the whole-host scan when a target responds with HTTP 429, instead of continuing to send login attempts to a host that is already rate-limiting.

## Why

`admin_panel_login_bruter` tries every credential pair against every discovered path with no reaction to rate limiting. Against real infrastructure, once a target starts returning 429 the module keeps hammering it — which can lock out real accounts, deepen the rate limit, or trip an incident on the scanned host. Root cause: none of the request sites inspected the status for 429, and there was no way for a "stop" signal to propagate out of the nested credential loops. There is no global handling to reuse either — `throttle_request` is a fixed-rate pacer and the `module_base` retry catches Karton task errors, not HTTP status codes — so this is introduced here.

## Fix

`_abort_if_rate_limited(response)` raises a `RateLimitedError` at every site where a 429 can come back — the brute-force requests (JSON-API POST, Basic-auth GET, main GET, form POST) and path discovery (`_is_login_path`, renamed from `check_url`). It propagates through `brute_force_login_path` — with an explicit `except RateLimitedError: raise` guard so the existing broad `except Exception` does not swallow it — up to `scan`, which catches it, stops, and returns `(results, rate_limited)`. Discovery runs before `scan`, so `run` wraps discovery + scan in the same handler and a 429 during discovery aborts cleanly there too. `run` then reports OK with a "rate-limited (HTTP 429)" reason when nothing was found, or INTERESTING with a note that the scan was cut short if credentials were found before the 429.

Key design decisions:
- **Abort the whole host, not just the path.** Rate limiting is per-host, so moving on to other paths would only make it worse.
- **Do not honor Retry-After.** For a mass scanner, sleeping on one host blocks a worker; aborting the host is cheaper and still respects the target's signal.
- **429 only** — an unambiguous status code, no phrase matching.

## Tests

- `test_rate_limited_aborts_scan` — verifies a target that serves the login page (GET 200) but returns 429 on every login attempt (POST) ends as OK, with a rate-limit reason and no reported credentials.
- `test_rate_limited_during_discovery_aborts` — verifies a target that returns 429 already during path discovery aborts cleanly (caught in `run`, no crash), also ending as OK with a rate-limit reason.
- The rest of the module suite (basic-auth, JSON API, redirect, simple, both vendor-pair tests, both email-field tests) stays green, confirming the abort path does not change behavior when no 429 is present.

## Notes / Out of scope

- Follow-up (not here): lockout-phrase detection (e.g. "account locked", "too many attempts") on 401/403. That needs a language-fragile phrase list like `common_failure_messages` plus its own tuning, so it is deliberately deferred; this PR covers only the unambiguous 429 signal.
- No change to the credential set, the HTTP layer, or throttling.

### DIFF
--- a/artemis/modules/admin_panel_login_bruter.py
+++ b/artemis/modules/admin_panel_login_bruter.py
@@ -4,7 +4,7 @@ import itertools
 import os
 import random
 import urllib.parse
-from typing import IO, List, Optional, Tuple
+from typing import IO, Any, List, Optional, Tuple
 
 import requests
 from bs4 import BeautifulSoup
@@ -17,6 +17,18 @@ from artemis.config import Config
 from artemis.module_base import ArtemisBase
 from artemis.password_utils import get_passwords
 from artemis.task_utils import get_target_url
+
+
+class RateLimitedError(Exception):
+    """Raised when a target responds with HTTP 429. Rate limiting is per-host, so
+    there is no point in trying other credentials or paths — the whole scan of the
+    host is aborted."""
+
+
+def _abort_if_rate_limited(response: Any) -> None:
+    if response is not None and response.status_code == 429:
+        raise RateLimitedError()
+
 
 COMMON_USERNAMES: List[str] = ["admin"]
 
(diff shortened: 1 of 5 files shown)


artemis/modules/admin_panel_login_bruter.py
kazet: I would rename this one and add _abort_if_rate-limited also here

kazet (approved):

PR #2947, opened 2026-07-16
Implement OS command injection detector module

imraflip:


artemis/modules/command_injection_detector.py
```diff
@@ -0,0 +1,245 @@
+import datetime
+import random
+import urllib
+from enum import Enum
+from timeit import default_timer as timer
+from typing import Any, Dict, List, Literal, Optional
```
kadewu: I sometimes forget to mention it, but many typing classes can be omitted
e.g List[Optional[str]] <=> list[str | None]

artemis/modules/command_injection_detector.py
```diff
@@ -0,0 +1,245 @@
...
+
+    @staticmethod
+    def _strip_query_string(url: str) -> str:
+        url_parsed = urllib.parse.urlparse(url)
+        return urllib.parse.urlunparse(url_parsed._replace(query="", fragment=""))
+
+    @staticmethod
+    def _url_with_payload(url: str, params: tuple[Any, ...], payload: str) -> str:
+        separator = "&" if "?" in url else "?"
+        assignments = "&".join(f"{name}={payload}" for name in params)
```
kadewu: Why not url-encode? Will it work correctly with `"\n{cmd}"` template?

artemis/modules/command_injection_detector.py
```diff
@@ -0,0 +1,245 @@
...
+        Find the minimal set of parameters that still proves the injection, capped at
+        COMMAND_INJECTION_MINIMAL_PARAMS_MAX_LEN. Falls back to the full set if none prove it alone.
+        """
+        threshold = Config.Modules.CommandInjectionDetector.COMMAND_INJECTION_TIME_THRESHOLD
+        baseline_payload = self.change_sleep_to_0(payload) if minimization_mode == "time" else ""
+
+        minimal_params: List[str] = []
+        for param in params:
+            if minimization_mode == "output":
+                assert marker is not None  # required for output-based minimization
```
kadewu: I don't think this should be as assert, if will be much cleaner.

artemis/modules/command_injection_detector.py
```diff
@@ -0,0 +1,245 @@
...
+                    self.measure_request_time(self._url_with_payload(url, (param,), baseline_payload)) < threshold / 2
+                    and self.measure_request_time(self._url_with_payload(url, (param,), payload)) >= threshold
+                )
+            if confirmed:
+                minimal_params.append(param)
+            if len(minimal_params) >= Config.Modules.CommandInjectionDetector.COMMAND_INJECTION_MINIMAL_PARAMS_MAX_LEN:
+                break
+        return minimal_params or params
+
+    def scan(self, urls: List[str], task: Task) -> List[Dict[str, Any]]:
```
kadewu: task is not used

artemis/modules/command_injection_detector.py
```diff
@@ -0,0 +1,245 @@
...
+        return message
+
+    @staticmethod
+    def create_status_reason(message: Any) -> str:
+        return ", ".join(sorted({f"{item.get('url')}: {item.get('statement')}" for item in message}))
+
+    def create_data(self, message: Any) -> Dict[str, Any]:
+        deduplicated_message = []
+        for item in message:
+            if item not in deduplicated_message:
```
kadewu: rather nothing big, but it's O(n**2), you can use set instead to track keys

PR #2978, opened 2026-07-30
Add CPE information to detected web technologies

imraflip:
This is the CPE half of #2842, split out so the technology detection can be reviewed and merged on its own. #2842 keeps the cve_lookup module.

`run_tech_detection` returns structured `Technology` objects (name, version, CPE, categories) instead of tag strings, and `webapp_identifier` carries them on the `WEBAPP` task payload. The CPE Wappalyzer gives back keeps a wildcard in the version slot and reports the version separately, so both are kept as their own fields - `cpe:2.3:a:apache:http_server:*:*:*:*:*:*:*:*` with `version="2.4.53"`. Nothing here reads the technologies field yet; it's there for downstream consumers, the CVE lookup in #2842  being the first.

### DIFF
--- a/artemis/modules/nuclei_router.py
+++ b/artemis/modules/nuclei_router.py
@@ -78,7 +78,7 @@ class NucleiRouter(ArtemisBase):
         if current_task.headers.get("receiver", None) == "nuclei":
             current_task.headers["receiver"] = "nuclei-router"
             current_task.headers["original_receiver"] = "nuclei"
-        self.db.save_task_result(
+        self.save_task_result(
             task=current_task,
             status=TaskStatus.OK,
             data={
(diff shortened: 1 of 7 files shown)


kadewu (dismissed):

artemis/modules/nuclei_router.py
```diff
@@ -78,7 +78,7 @@ def run(self, current_task: Task) -> None:
         if current_task.headers.get("receiver", None) == "nuclei":
             current_task.headers["receiver"] = "nuclei-router"
             current_task.headers["original_receiver"] = "nuclei"
-        self.save_task_result(
+        self.db.save_task_result(
```
kadewu: Actually maybe one small things - the conflicts were not resolved correctly - it should be 'self.save_task_result` as we have changed the invocation on main. Same below

kadewu (approved):

PR #2996, opened 2026-08-04
subdomain enumeration improvment

kadewu:


kazet (approved):

artemis/modules/subdomain_enumeration.py
```diff
@@ -260,6 +260,36 @@ def get_subdomains_by_dns_brute_force(self, domain: str) -> Optional[Set[str]]:
 
         return subdomains
 
+    def _filter_wildcard_subdomains(self, subdomains: set[str], parent_domain: str) -> set[str]:
```
kazet: why can't we extract the common code from get_subdomains_by_dns_brute_force ?

artemis/modules/subdomain_enumeration.py
```diff
@@ -322,50 +351,53 @@ def run(self, current_task: Task) -> None:
...
-                            "domain": subdomain,
-                        },
-                    )
-                    self.add_task(current_task, task)
-
-            valid_subdomains.update(valid_subdomains_from_tool)
+                valid_subdomains.add(subdomain)
+
+        threshold = Config.Modules.SubdomainEnumeration.LARGE_SUBDOMAIN_COUNT_VERIFICATION_THRESHOLD
+        if threshold > 0 and len(valid_subdomains) > threshold:
```
kazet: nit: should be >= threshold

artemis/config.py
```diff
@@ -1134,6 +1134,15 @@ class SubdomainEnumeration:
                 "SUBDOMAIN_ENUMERATION_GAU_ADDITIONAL_OPTIONS", default="", cast=decouple.Csv(str, delimiter=" ")
             )
 
+            LARGE_SUBDOMAIN_COUNT_VERIFICATION_THRESHOLD: Annotated[
```
kazet, line 1137: we might already set the threshold in internal env so that it won't get missed

kazet (changes requested):

kazet (dismissed):

PR #3011, opened 2026-08-05
Record a CPE for assets identified by Nuclei and Wappalyzer

X-R3Y24:
## What

Adds a `cpe` field to the `Asset` model and fills it from two modules: Nuclei (for exposed panels, from the
matched template) and `webapp_identifier` (for detected technologies and CMSes, from Wappalyzer).

## Why

Assets describe the software they found in `additional_type`, which is free-form and source-specific:
`wordpress-login` from a Nuclei template path, `Apache HTTP Server` from Wappalyzer, `wordpress` from the CMS
scanners. Three spellings of the same fact, none of which can be matched against a vulnerability database.
CPE is the standard identifier for exactly that, and it is what a CVE lookup keys on.

The data already exists at both sources and was being dropped on the floor:

- Nuclei templates carry it in `info.classification.cpe`, and since findings are stored verbatim it was
already sitting in the task results.
- #2978 brought Wappalyzer CPEs into the `webapp_identifier` task result, but nothing read them: the reporter built its assets from `(name, version)` pairs and discarded the rest.

## Fix

- `Asset.cpe`, documented with what a consumer has to expect from it (see below).
- `artemis/reporting/base/cpe.py` with `to_cpe`, used by both reporters.
- The Nuclei reporter reads `info.classification.cpe` for the exposed panel assets it already creates.
- The `webapp_identifier` reporter passes through the CPE that was already in the task result. Its
`_iter_technologies` now returns `Technology` objects instead of `(name, version)` tuples, so that adding a third field didn't mean a third element in an anonymous tuple.

Two design decisions worth stating explicitly:

**The CPE is stored exactly as the source reports it, and the field is documented as such.** Two consequences
for a consumer. First, the version slot is usually a wildcard even when the version is known — Wappalyzer reports `cpe:2.3:a:apache:http_server:*:*:*:*:*:*:*:*` together with `version="2.4.53"`, so `cpe` and `version` stay separate fields and whoever needs a versioned CPE combines them. Second, the binding isn't guaranteed: this is normally a CPE 2.3 formatted string, but the older 2.2 URI (`cpe:/a:vendor:product`) exists in the wild. Both are accepted rather than one being silently dropped.

**Validation is shared and deliberately shallow.** Both sources are outside our control: Nuclei templates are updated upstream and can be added locally, and the Wappalyzer fingerprint database is community-maintained. `to_cpe` therefore checks the binding prefix and that something follows it, and drops anything else rather than exporting it as if it were a CPE. It does not parse the structure — validating 13 components would reject tomorrow's legitimate value for no benefit. Keeping it in one place is what makes the two paths agree; they started out with different rules, and that asymmetry is exactly what a reviewer would rightly flag.

## Tests

- `test_to_cpe` — verifies which values count as a CPE name: both bindings, surrounding whitespace, and the rejected cases (empty, a bare prefix with nothing after it, `cpe:whatever`, a non-string).
- `test_nuclei_asset_cpe` — verifies the CPE reaches the asset, and that everything a template can put in `classification` instead of a CPE yields `None`: no classification at all, `null`, a string, a list, a malformed value, a finding with no `info` section.
- `test_webapp_identifier_asset_cpe` — verifies the CPE reaches both the `TECHNOLOGY` and the `CMS` asset, that a technology Wappalyzer has no CPE for stays `None`, and that task results saved before #2978 (only `technology_tags`, no CPEs) still produce assets.
- Both autoreporter integration tests now assert the CPEs as well. They run a real Nuclei scan and a real Wappalyzer detection against the test containers, so these are the values the sources actually return rather than fixtures: `cpe:2.3:a:wordpress:wordpress:*:*:*:*:*:*:*:*` for the WordPress login panel, `cpe:2.3:o:debian:debian_linux:*:*:*:*:*:*:*:*` for the operating system, and so on.
- `test/e2e/test_exporting.py` — the exported asset gained a field. The export path serializes the dataclass reflectively, so nothing else had to change.

## Notes / Out of scope

**Expect `cpe` to be `None` most of the time, and not because something is broken.** Only 316 of the ~7500 technologies Wappalyzer knows carry a CPE (~4%); for Nuclei exposed panels the coverage is much better, 602 of ~1350 (44%), as of now. An asset with an empty `cpe` is the normal case, not a symptom.

**Assets are deduplicated in `artemis/reporting/export/db.py` by an explicit
`(asset_type, additional_type, name, last_domain)` tuple.** That tuple does not include the CPE and
shouldn't, since otherwise the same asset reported once with a CPE and once without would end up as two rows. The flip side is that on a collision the first occurrence wins, so an asset first seen without a CPE keeps `cpe=None` even if a later task result carries one.

That collision is reachable: `webapp_identifier` and `wp_scanner` both produce a `CMS` / `wordpress` asset for the same URL, and the query feeding the export has no `ORDER BY`, so which of the two is kept isn't deterministic. This PR doesn't create that behaviour; `cpe` just makes it routine rather than occasional. The field is asymmetric by construction: `webapp_identifier` has it whenever Wappalyzer knows one, `wp_scanner` never does, so the collision now means `None` instead of data on every WordPress scan. `version` is already affected today, less often: `wp_scanner` only records `wp_version` when it manages to read it, so when it doesn't and Wappalyzer does, the version is lost the same way. A follow-up in `db.py` will fill missing fields from the discarded duplicate instead of dropping it.

### DIFF
--- a/artemis/reporting/base/cpe.py
+++ b/artemis/reporting/base/cpe.py
@@ -1,0 +1,25 @@
+from typing import Any, Optional
+
+# A CPE name may come in one of two bindings: the current 2.3 formatted string
+# (cpe:2.3:a:wordpress:wordpress:*:*:*:*:*:*:*:*) or the older 2.2 URI (cpe:/a:redhat:infinispan).
+# Both are accepted
+CPE_PREFIXES = ("cpe:2.3:", "cpe:/")
+
+
+def parse_cpe(value: Any) -> Optional[str]:
+    """Returns the value if it looks like a CPE name (https://en.wikipedia.org/wiki/Common_Platform_Enumeration).
+
+    Only the binding prefix is checked, not the structure - the point is to drop values that clearly
+    aren't CPE names, since they come from sources we don't control and would otherwise be exported as if they were.
+    Anything (not only a string) may be passed here.
+    """
+    if not isinstance(value, str):
+        return None
+
+    cpe: str = value.strip()
+
+    # A bare prefix with nothing after it names no product, so it's of no use to anybody reading the asset.
+    if not any(cpe.startswith(prefix) and len(cpe) > len(prefix) for prefix in CPE_PREFIXES):
+        return None
+
+    return cpe
--- a/artemis/reporting/base/cpe.py
+++ b/artemis/reporting/base/cpe.py
@@ -6,7 +6,7 @@ from typing import Any, Optional
 CPE_PREFIXES = ("cpe:2.3:", "cpe:/")
 
 
-def to_cpe(value: Any) -> Optional[str]:
+def extract_cpe(value: Any) -> Optional[str]:
     """Returns the value if it looks like a CPE name (https://en.wikipedia.org/wiki/Common_Platform_Enumeration).
 
     Only the binding prefix is checked, not the structure - the point is to drop values that clearly
--- a/artemis/reporting/base/cpe.py
+++ b/artemis/reporting/base/cpe.py
@@ -6,7 +6,7 @@ from typing import Any, Optional
 CPE_PREFIXES = ("cpe:2.3:", "cpe:/")
 
 
-def parse_cpe(value: Any) -> Optional[str]:
+def to_cpe(value: Any) -> Optional[str]:
     """Returns the value if it looks like a CPE name (https://en.wikipedia.org/wiki/Common_Platform_Enumeration).
 
     Only the binding prefix is checked, not the structure - the point is to drop values that clearly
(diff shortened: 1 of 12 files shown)


artemis/reporting/base/cpe.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any, Optional
+
+# A CPE name may come in one of two bindings: the current 2.3 formatted string
+# (cpe:2.3:a:wordpress:wordpress:*:*:*:*:*:*:*:*) or the older 2.2 URI (cpe:/a:redhat:infinispan).
+# Both are accepted
+CPE_PREFIXES = ("cpe:2.3:", "cpe:/")
+
+
+def parse_cpe(value: Any) -> Optional[str]:
```
kazet: if this doesn't parse, maybe a different name would be ok?

X-R3Y24: Fair point. Renamed to `to_cpe`

PR #3028, opened 2026-08-10
Extract shared crawl helpers from injection detectors

imraflip:
The SQL, ORM, and LFI detectors each had their own copy of the same crawl prelude at the top of `run()`, along with a private `_strip_query_string` helper that was identical across all three. Any change to how the list of links to scan gets built had to be made in every module separately, and the copies had already started drifting apart.

This moves both shared pieces into `crawling.py`, next to `crawl_and_filter` where the rest of the crawling logic already lives. `strip_query_string` drops the query and fragment from a URL, and `prepare_links` runs the whole prelude - it crawls the target, adds the seed URL back in, unions in the query-stripped variants, filters out links pointing at static assets, shuffles, and caps the result at `MAX_URLS_TO_SCAN`. Each detector now calls `prepare_links(url)` in place of that block.

Nothing changes behaviorally - the relocated code is identical to what the detectors ran before, so their existing tests cover it without modification. The command injection detector (#2947) and the NoSQLi detector I'm working on next will adopt the same helpers in their own PRs rather than copying the prelude again.

### DIFF
--- a/artemis/modules/orm_injection_detector.py
+++ b/artemis/modules/orm_injection_detector.py
@@ -12,7 +12,7 @@ from artemis.binds import Service, TaskStatus, TaskType
 from artemis.config import Config
 from artemis.crawling import (
     get_injectable_parameters,
-    prepare_links,
+    get_links_to_scan,
     strip_query_string,
 )
 from artemis.http_requests import HTTPResponse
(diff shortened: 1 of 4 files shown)


artemis/modules/orm_injection_detector.py
```diff
@@ -308,20 +301,9 @@ def create_data(message: Any) -> dict[str, Any]:
 
     def run(self, current_task: Task) -> None:
         url = get_target_url(current_task)
+        links = prepare_links(url)
```
kazet: what do you think about a more descriptive name?

kazet (approved):

PR #3032, opened 2026-08-13
more optimization around batching

kadewu:
Additionaly added tests as the logic started to grow.
- units
- one e2e for the whole mechanism of batch picking tasks

### DIFF
--- a/artemis/module_base.py
+++ b/artemis/module_base.py
@@ -149,9 +149,13 @@ class ArtemisBase(Karton):
     def get_runtime_configuration(self, _task: Task) -> ModuleRuntimeConfiguration:
         return self.get_default_configuration()
 
-    def get_batch_group_key(self, _task: Task) -> str | None:
+    def _get_requests_per_second_batch_key(self, task: Task) -> str | None:
+        override = task.payload_persistent.get("requests_per_second_override")
+        return str(override) if override is not None else None
+
+    def get_batch_group_key(self, task: Task) -> str | None:
         """Return a grouping key used when taking a batch of tasks from queue."""
-        return None
+        return self._get_requests_per_second_batch_key(task)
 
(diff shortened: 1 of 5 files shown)


artemis/module_base.py
```diff
@@ -149,9 +149,13 @@ def get_default_configuration(self) -> ModuleRuntimeConfiguration:
     def get_runtime_configuration(self, _task: Task) -> ModuleRuntimeConfiguration:
         return self.get_default_configuration()
 
-    def get_batch_group_key(self, _task: Task) -> str | None:
+    def _get_requests_per_second_batch_key(self, task: Task) -> str | None:
+        override = task.payload_persistent.get("requests_per_second_override")
+        return str(override) if override is not None else None
```
kazet: else I would take the default

artemis/module_base.py
```diff
@@ -353,12 +357,12 @@ def _single_iteration(self) -> int:
...
             increase_analysis_num_in_progress_tasks(REDIS, task.root_uid, by=1)
 
-        requests_per_second_overrides = [
-            task.payload_persistent.get("requests_per_second_override")
-            for task in tasks
-            if "requests_per_second_override" in task.payload_persistent
-        ]
+        # batching guarantees all tasks share the same override, so reading the first is sufficient
+        override_from_payload = tasks[0].payload_persistent.get("requests_per_second_override") if tasks else None
+        requests_per_second_overrides = [override_from_payload] if override_from_payload is not None else []
```
kazet: hmm, why we here take the value for the first task, not all of them?

kadewu: As per comment, it's the same value for every task (as we no longer mix different rps).
I didn't add any check if that's true in code, only tested in tests, but if we want to have some defence mechanism here, I'm fine with it.

kazet: let's add some defense mechanism

PR #3051, opened 2026-08-18
Add nosql injection detector

imraflip:
This PR adds a module for detecting NoSQL (MongoDB-style) injection, where a framework like Express parses `?q[$ne]=1` into a nested object `{q: {$ne: "1"}}` and passes it straight into the query, so a user can control the query structure instead of just a value.

The module detects this two ways. The error-based check injects an operator no engine knows (`$artemisProbe`) together with a baseline that has the `$` stripped, so the same key is read as a plain field name. If the operator reaches the query, MongoDB raises an `unknown operator` error that shows up with the operator but not in the baseline, which a reflection or a WAF can't reproduce. The same check also runs through a JSON request body for login-style endpoints.

The blind check covers what the error-based one misses, like an authentication bypass with `username[$ne]=x` where the app processes the operator but never leaks an error. It compares `[$ne]=<random>` (matches almost everything) against `[$eq]=<random>` (matches nothing) and reports when the responses differ. This is the same idea the ORM detector already uses, so the two now share the response similarity check, the dynamic-content guard, and the confirmation loop (`NOSQL_INJECTION_NUM_CONFIRMATIONS`, 10 by default) that only reports a difference if it reproduces every time.

I've also bundled the shared-helper extraction into this PR. The result-building moves into `artemis/injection_helpers.py` and the parameter collection into `crawling.collect_parameters`, and the SQL and ORM detectors are wired onto them so there's one copy instead of three. It sits at the top level next to `crawling.py` and `task_utils.py` because `artemis/modules/base/` holds base classes that modules subclass, while these are loose functions. `sql_injection_detector` keeps its own `create_status_reason` - it appends the `Headers used:` detail, so it isn't the same function.

The test app is a real Express + MongoDB service with an endpoint that leaks database errors, one that swallows them so only the blind check catches it, and coercing/noisy endpoints to confirm the module doesn't false-positive.

### DIFF
--- a/artemis/modules/nosql_injection_detector.py
+++ b/artemis/modules/nosql_injection_detector.py
@@ -1,0 +1,313 @@
+import json
+import re
+import uuid
+from enum import Enum
+from typing import Any, Callable, TypeVar
+
+import more_itertools
+from karton.core import Task
+
+from artemis import load_risk_class
+from artemis.binds import Service, TaskStatus, TaskType
+from artemis.config import Config
+from artemis.crawling import collect_parameters, get_links_to_scan, strip_query_string
+from artemis.http_requests import HTTPResponse
+from artemis.module_base import ArtemisBase
+from artemis.modules.base.injection_helpers import (
+    build_result_data,
+    create_status_reason,
+    responses_differ,
+)
+from artemis.nosql_injection_data import (
+    BLIND_FALSE_OPERATOR,
+    BLIND_TRUE_OPERATOR,
+    BRACKET_OPERATOR_PAYLOADS,
+    JSON_OPERATOR_PAYLOADS,
+    NOSQL_ERROR_MESSAGES,
+    PARAMS_PER_BATCH,
+)
+from artemis.task_utils import get_target_url
+
+JSON_HEADERS = {"Content-Type": "application/json"}
+
+# A random value no record holds ("$ne" then matches every record, "$eq" none), kept to 8 hex chars
+# so that a full batch of bracket parameters stays under the URL-length limit.
+UNLIKELY_VALUE = uuid.uuid4().hex[:8]
+
+T = TypeVar("T")
+
+
+class Statements(Enum):
+    nosql_injection = "nosql_injection"
+    nosql_injection_json_body = "nosql_injection_json_body"
+    nosql_injection_blind = "nosql_injection_blind"
+
+
+@load_risk_class.load_risk_class(load_risk_class.LoadRiskClass.HIGH)
+class NoSqlInjectionDetector(ArtemisBase):
+    """
+    Module for detecting MongoDB-style NoSQL injection vulnerabilities.
+    """
+
+    num_retries = Config.Miscellaneous.SLOW_MODULE_NUM_RETRIES
+    identity = "nosql_injection_detector"
+    filters = [
+        {"type": TaskType.SERVICE.value, "service": Service.HTTP.value},
+    ]
+
+    @staticmethod
+    def _baseline_operator(operator: str) -> str:
+        # Removing the leading "$" yields a key the query engine reads as a literal subdocument
+        # field rather than an operator, so the baseline request differs from the payload request
+        # only in whether an operator is present.
+        return operator.removeprefix("$")
+
+    @staticmethod
+    def _build_bracket_url(url: str, params: list[str], operator: str, value: str) -> str:
+        # The brackets and "$" are sent literally so the target's query parser reads them as a nested
+        # operator object; percent-encoding them would decode to the same thing before parsing.
+        base = strip_query_string(url)
+        query = "&".join(f"{name}[{operator}]={value}" for name in params)
+        return f"{base}?{query}"
+
+    @staticmethod
+    def _build_json_body(params: list[str], operator: str, value: Any) -> str:
+        return json.dumps({name: {operator: value} for name in params})
+
+    def _contains_error(self, url: str, response: HTTPResponse | None) -> str | None:
+        if response is None:
+            return None
+
+        for message in NOSQL_ERROR_MESSAGES:
+            if re.search(message, response.content):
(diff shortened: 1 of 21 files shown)


artemis/modules/nosql_injection_detector.py
```diff
@@ -0,0 +1,313 @@
...
+        self.log.info("Scanning URLs: %s", urls)
+        message: list[dict[str, Any]] = []
+        untestable_urls: list[str] = []
+        stop_on_first_match = Config.Modules.NoSqlInjectionDetector.NOSQL_INJECTION_STOP_ON_FIRST_MATCH
+
+        for current_url in urls:
+            all_params = collect_parameters(current_url)
+            blind_enabled = not self._has_dynamic_content(current_url)
+
+            url_completed = False
```
kadewu: If I'm seeing correctly, if every url will not have parameters - each would be marked as not completed - therefore task will end up in ERROR status? Expected?

artemis/modules/nosql_injection_detector.py
```diff
@@ -0,0 +1,313 @@
...
+
+    def scan(self, urls: list[str]) -> tuple[list[dict[str, Any]], list[str]]:
+        self.log.info("Scanning URLs: %s", urls)
+        message: list[dict[str, Any]] = []
+        untestable_urls: list[str] = []
+        stop_on_first_match = Config.Modules.NoSqlInjectionDetector.NOSQL_INJECTION_STOP_ON_FIRST_MATCH
+
+        for current_url in urls:
+            all_params = collect_parameters(current_url)
+            blind_enabled = not self._has_dynamic_content(current_url)
```
kadewu: We probably can skip those requests if we do not plan to test this url when no params found.

artemis/modules/nosql_injection_detector.py
```diff
@@ -0,0 +1,313 @@
...
+
+        error = self._contains_error(base, response_payload)
+        if error and not self._contains_error(base, response_baseline):
+            return error, True
+        return None, True
+
+    def _has_dynamic_content(self, url: str) -> bool:
+        # Two identical requests that already differ mean the endpoint is non-deterministic (timestamps,
+        # tokens), so the blind differential below would flag it on every run. Skip those.
+        return responses_differ(self.forgiving_http_get(url), self.forgiving_http_get(url))
```
kadewu, line 119: Aren't we also doing POST requests?

artemis/modules/nosql_injection_detector.py
```diff
@@ -0,0 +1,313 @@
...
+    BRACKET_OPERATOR_PAYLOADS,
+    JSON_OPERATOR_PAYLOADS,
+    NOSQL_ERROR_MESSAGES,
+    PARAMS_PER_BATCH,
+)
+from artemis.task_utils import get_target_url
+
+JSON_HEADERS = {"Content-Type": "application/json"}
+
+# A random value no record holds ("$ne" then matches every record, "$eq" none), kept to 8 hex chars
```
kadewu: Nitpick: I would call it 'Randomly generated value that is unlikely to exists'

kadewu (commented):
I'm having a problem fully grasp the whole change.
Can you maybe extract some logical parts, so it's easier to follow the code as whole. E.g. by splitting scan()

Other question is:
How many http requests are we doing, it seems that we may hit quite a lot. Did you make any performance tests of that module?

artemis/modules/nosql_injection_detector.py
imraflip: done, reworded it to "Randomly generated value that is unlikely to exist in any record". I also added the reason it's cut to 8 hex chars - a full batch of bracket params has to fit under the url length limit

artemis/modules/nosql_injection_detector.py
imraflip: and yes, you were right, this one was a real bug. The guard only measured GET, so a noisy POST endpoint would give us a FP, and worse, a real vuln got skipped whenever only the GET side was noisy

I split it into `_has_dynamic_get_content()` and `_has_dynamic_post_content()`, the POST one sending two identical empty-body json posts. It's 2 extra requests per url, and both guards are measured once per url instead of per batch so it doesn't add up

artemis/modules/nosql_injection_detector.py
imraflip: yeah, you're right. `url_completed` only got set inside the batch loop, so a url with no params would skip the loop entirely, stay False, and land in `untestable_urls` - and if that happened for every url the task went to ERROR even though nothing actually failed

fixed with the same guard as your other comment: a url with no params is skipped before any probe runs, so it never gets into `untestable_urls`. ERROR is only left for urls where the probes ran but couldn't reach their baseline

artemis/modules/nosql_injection_detector.py
imraflip: `scan()` now does a `continue` before any probe fires when `collect_parameters()` comes back empty, so nothing goes out for a url we're not gonna test anyway

imraflip:
Thanks for the review :)! Split `scan()` into `_scan_url()`, `_iter_batch_findings()`, and four `_check_error_get` / `_check_error_post` / `_check_blind_get` / `_check_blind_post` methods, one per probe strategy - `scan()` is just the loop over urls now.

On the request count: params are batched 50 at a time, so a url with no params of its own is covered by the 396-name wordlist in 8 batches. Each batch is 20 requests, plus 4 determinism probes per url - so 164 per url, 82 GET and 82 POST. Unbatched it'd be around 7.9k. The confirmation loop only re-runs a probe that already matched, so it costs nothing on a clean url, and `NOSQL_INJECTION_STOP_ON_FIRST_MATCH` is on by default.

Also I moved the shared helpers to `artemis/injection_helpers.py` and added `test/unit/test_injection_helpers.py`, and rebased onto main.

PR #3070, opened 2026-08-25
Add cpe utils

kadewu:


artemis/cpe_tools/cpe_main_process.py
```diff
@@ -0,0 +1,187 @@
...
+                "Run `python3 -m artemis.cpe_tools.cpe_main_process` to download it.",
+                nvd_dir,
+            )
+            index = {}
+    INDEX_CACHE[nvd_dir] = index
+    return index
+
+
+def build_index(nvd_dir: Path | None = None) -> Path:
+    directory = (Path(nvd_dir) if nvd_dir else get_nvd_dir()).resolve()
```
kazet, line 220: let's filter deprecated CPEs

artemis/cpe_tools/cpe_main_process.py
```diff
@@ -0,0 +1,187 @@
...
+            if not isinstance(cpe_name, str):
+                continue
+            title = None
+            for entry in cpe.get("titles", []):
+                if isinstance(entry, dict) and entry.get("lang") == "en":
+                    title = entry.get("title")
+                    break
+            if not title:
+                continue
+            yield cpe_name, title
```
kazet: let's keep URLs for the purpose of WordPress plugin matching (like here https://github.com/CERT-Polska/Artemis/pull/3054/changes#diff-5556ed4c8bfce815435dd2f73f617ebbc1e1a8783466d509218370a880e424feR58)

artemis/cpe_tools/cpe_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    cache_key = (nvd_dir, normalized)
+    cached: object = LOOKUP_CACHE.get(cache_key, _MISSING)
+    if cached is not _MISSING:
+        return cached if isinstance(cached, str) else None
+
+    index = ensure_index(nvd_dir)
+
+    cpe = index.get(normalized)
+    if cpe is not None:
+        result: str | None = with_version(cpe, "*")
```
kazet: let's also strip sw edition (e.g. "free", "paid"): https://github.com/CERT-Polska/Artemis/pull/3054/changes#diff-5556ed4c8bfce815435dd2f73f617ebbc1e1a8783466d509218370a880e424feR20

artemis/cpe_tools/cpe_utils.py
```diff
@@ -0,0 +1,108 @@
...
+def _cache_put(key: tuple[Path, str], value: str | None) -> None:
+    while len(LOOKUP_CACHE) >= MAX_LOOKUP_CACHE_SIZE:
+        LOOKUP_CACHE.pop(next(iter(LOOKUP_CACHE)), None)
+    LOOKUP_CACHE[key] = value
+
+
+def with_version(cpe: str, version: str) -> str:
+    # A CPE 2.3 well-formed name escapes ":" inside fields as %3a, so splitting on
+    # ":" always yields exactly 13 components: cpe, 2.3, part, vendor, product,
+    # version, update, edition, language, sw_edition, target_sw, target_hw, other.
```
X-R3Y24: Note on `with_version` and on the planned `sw_edition` stripping: some CPE names carry an escaped colon inside a field, so `split(":")` returns 14 components instead of 13, and indices 5 and 9 then point at the wrong ones. Example: `cpe:2.3:a:cgiirc:cgi\:irc:0.5.7:*:*:*:*:*:*:*`, where index 5 is `irc`, not `0.5.7`.

Hotfix: split on `(?<!\\):` - that handles the single-escape case, which is all this feed seems to contain; strictly speaking CPE 2.3 also allows an escaped backslash, so a fully correct parser would need to account for that too.

(The comment above assumes `%3a` escaping - on the current feed no `cpeName` contains `%3a`, they all use `\:`.)

artemis/cpe_tools/cpe_main_process.py
```diff
@@ -0,0 +1,290 @@
...
+        for product in data.get("products", []):
+            cpe = product.get("cpe") or {}
+            if cpe.get("deprecated"):
+                continue
+            cpe_name = cpe.get("cpeName")
+            if not isinstance(cpe_name, str):
+                continue
+            title = None
+            for entry in cpe.get("titles", []):
+                if isinstance(entry, dict) and entry.get("lang") == "en":
```
kazet:
```suggestion
                if isinstance(entry, dict) and entry.get("lang").lower().startswith("en"):
```

kazet:
```suggestion
                if isinstance(entry, dict) and entry.get("lang").lower().startswith("en"):
```

PR #3071, opened 2026-08-25
emit URL tasks from robots module for pipeline-wide path scanning

### LINKED ISSUE #2405: Feature Request: Allow robots module to emit URL tasks for discovered paths
Hi @kazet,

While exploring the Artemis modules and running the project locally, I noticed something about the `robots` module that I wanted to ask about.

Currently, the module downloads and parses `robots.txt`, extracts `Allow` and `Disallow` paths, and then checks those paths mainly for directory index exposure. However, many of these paths can point to interesting application endpoints such as `/admin`, `/backup`, `/dev`, etc.

Right now those discovered paths are only used within the `robots` module itself. I was wondering if it might make sense to instead (or additionally) emit URL tasks for these paths using `self.add_task`, so they can go through the rest of the Artemis pipeline.

That way other modules such as:

- `sql_injection_detector`
- `lfi_detector`
- `admin_panel_login_bruter`
- `webapp_identifier`
- `nuclei`
- `directory_index`

could analyze those endpoints as well.

The idea would roughly be:

robots.txt
↓
extract paths
↓
emit URL tasks
↓
other Artemis modules analyze those endpoints

This might help reuse the discovered endpoints more effectively and increase scanning coverage without duplicating logic across modules.

Before trying to implement this, I wanted to check whether this approach fits the intended design of the Artemis task pipeline.

If it does, I’d be happy to work on a PR for it.

Thanks!

ManthanNimodiya:
Closes #2405

## What

The robots module parses `robots.txt` and discovers paths that site owners intentionally (or accidentally) want to hide. Previously these paths were only checked for directory index exposure inside the module itself.

This PR emits a `TaskType.URL` task for each non-trivial discovered path, feeding them into the broader Artemis scanning pipeline.

## Changes

- `artemis/modules/robots.py` -> added `_emit_url_tasks()` which emits a `TaskType.URL` task per discovered path, applying the same filters already used for directory-index checks (skips `NOT_INTERESTING_PATHS`, wildcard paths, and deduplicates across groups).
- `artemis/modules/lfi_detector.py` -> added `TaskType.URL` to filters so it picks up and probes paths discovered by robots.
- `artemis/modules/sql_injection_detector.py` -> same.
- `test/modules/test_robots.py` -> added assertion that the expected URL tasks are emitted for the test host.

## Why

Paths listed in `robots.txt` are often sensitive endpoints (admin panels, API routes, internal tools). By routing them through the pipeline, modules like `lfi_detector` and `sql_injection_detector` will crawl and probe them, surfacing vulnerabilities that would be missed when scanning only from the root URL.

### DIFF
--- a/artemis/modules/lfi_detector.py
+++ b/artemis/modules/lfi_detector.py
@@ -38,6 +38,7 @@ class LFIDetector(ArtemisBase):
     identity = "lfi_detector"
     filters = [
         {"type": TaskType.SERVICE.value, "service": Service.HTTP.value},
+        {"type": TaskType.URL.value},
     ]
 
     def create_url_with_batch_payload(self, url: str, param_batch: List[str], payload: str) -> str:
--- a/artemis/modules/robots.py
+++ b/artemis/modules/robots.py
@@ -110,11 +110,31 @@ class RobotsScanner(ArtemisBase):
         result = RobotsResult(response.status_code, groups, self._download(current_task, url, groups))
         return result
 
+    def _emit_url_tasks(self, current_task: Task, url: str, groups: List[RobotsGroup]) -> None:
+        seen: set = set()
+        for g in groups:
+            for path in g.allow + g.disallow:
+                if any(re.match(p, path) for p in NOT_INTERESTING_PATHS):
+                    continue
+                if "*" in path:
+                    continue
+                path = path.rstrip("$")
+                full_url = f"{url}{path}"
+                if full_url in seen:
+                    continue
+                seen.add(full_url)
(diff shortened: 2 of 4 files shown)


ManthanNimodiya:
@kadewu have a look when you get time

kadewu:
It was never considered yet, if we want to extend other modules with such findings. I will not close this for now, and once decision is made, I will revisit this PR.

kazet:
In my opinion, emitting URL tasks for each found URL in robots.txt will take too long and cause too big load on scanned systems.

PR #3072, opened 2026-08-25
Naming fix

kazet:


artemis/web_technology_identification.py
```diff
@@ -35,6 +35,9 @@ def _parse_tech(raw: Dict[str, Any]) -> Technology:
     else:
         name, version = raw_name, None
 
+    if name == "Basic":
```
bulkowy: As we got at least one of these cases, I would create dict-mapping `{received: intended}` and use it. Also add info of source where `Basic -> Basic Auth` comes from

bulkowy (approved):

PR #3079, opened 2026-08-27
fix s3mock

kadewu:
additionaly changed the variable that was changed in 5.0


bulkowy (approved):

kadewu:
Two tests are not related to this change, will fix in separate PR.

PR #3082, opened 2026-08-28
Add CPE to wordpress plugins

X-R3Y24:
## What this does

`wordpress_plugins` already reports every plugin it detects as a `CMS_PLUGIN`
asset. This fills in the one field those assets were always missing, `cpe`, so a
plugin can be matched against the CVE database and correlated with other tools'
output, which the free-form `additional_type` cannot do.

```python
cpe=lookup_cpe_by_plugin_slug(slug, "wordpress")
```

## Why it looks this way

Neither source this module consumes publishes a CPE. The WordFence
vulnerability feed names software by `type`/`name`/`slug` only, its v3 feed
format documents no CPE field, and the wordpress.org plugin API does not mention
CPEs at all. The CVE records where WordFence is the CNA carry neither
`affected[].cpes` nor `cpeApplicability`.

What does tie the two together is NVD's own CPE dictionary. A plugin's CPE
carries its wordpress.org directory URL among its references:

```
cpe:2.3:a:rocklobster:contact_form_7:3.7.1:*:*:*:*:wordpress:*:*
    refs: http://wordpress.org/plugins/contact-form-7/changelog/
```

That reference is an authoritative join between the slug the scanner detects and
the name NIST assigned. `cpe_tools` builds the index from it, so this module only
has to consult it.

It matters that the name comes from there rather than from the slug's spelling,
because the two often differ: `google-analytics-for-wordpress` is
MonsterInsights in NVD, `insert-headers-and-footers` is WPCode,
`host-webfonts-local` is OMGF. Plugins get rebranded and keep their old slug, and
no spelling rule recovers that.

The version slot stays wildcarded, following the convention in `Asset.cpe`: the
detected version stays in the asset's `version` field, and a consumer that needs
a versioned CPE combines the two, the way `cve_lookup` does before querying NVD.

Most plugins get `None`, and that is the correct answer. NVD names a plugin when
it first gets a CVE, so plugins that never had a vulnerability reported are
simply absent from the dictionary. `Asset.cpe` already documents an absent CPE as
the normal case rather than a failure.

## Tests

`WordpressPluginAssetCPETest` patches the lookup and checks what this reporter puts
on the Asset: the CPE in `cpe`, the detected version in `version`, and a plugin the
dictionary doesn't name still reported as an asset without a CPE.

## Notes for the reviewer

The lookup only answers once `cpe-nvd-service` has downloaded the dictionary into
the volume `autoreporter` reads. `docker-compose.test-e2e.yaml` has an `autoreporter`
but no `cpe-nvd-service` and no volume, so e2e runs will see `None` for every CPE.
Nothing asserts on CPEs there today, so nothing breaks; flagging it in case someone
later wonders why.

### DIFF
--- a/artemis/reporting/modules/wordpress_plugins/reporter.py
+++ b/artemis/reporting/modules/wordpress_plugins/reporter.py
@@ -1,6 +1,6 @@
 import os
 import re
-from typing import Any, Callable, Dict, List, Optional, Set
+from typing import Any, Callable, Dict, List, Set
 
 from packaging import version
 
@@ -18,10 +18,6 @@ from artemis.reporting.base.templating import ReportEmailTemplateFragment
 from artemis.reporting.utils import get_target_url, get_top_level_target
 
 logger = utils.build_logger(__name__)
-
-
-def _get_cpe(slug: str) -> Optional[str]:
-    return lookup_cpe_by_plugin_slug(slug, "wordpress")
 
 
 class WordpressPluginsReporter(Reporter):
@@ -197,7 +193,7 @@ class WordpressPluginsReporter(Reporter):
                 name=get_target_url(task_result),
                 additional_type="wordpress-plugin:" + slug,
                 version=data.get("version", ""),
-                cpe=_get_cpe(slug),
+                cpe=lookup_cpe_by_plugin_slug(slug, "wordpress"),
             )
             for slug, data in task_result["result"].get("plugins", {}).items()
         ]
--- a/artemis/reporting/modules/wordpress_plugins/reporter.py
+++ b/artemis/reporting/modules/wordpress_plugins/reporter.py
@@ -1,10 +1,11 @@
 import os
 import re
-from typing import Any, Callable, Dict, List, Set
+from typing import Any, Callable, Dict, List, Optional, Set
 
 from packaging import version
 
 from artemis import utils
+from artemis.cpe_tools.cpe_utils import lookup_cpe_by_plugin_slug
 from artemis.fallback_api_cache import FallbackAPICache
 from artemis.reporting.base.asset import Asset
 from artemis.reporting.base.asset_type import AssetType
@@ -17,6 +18,10 @@ from artemis.reporting.base.templating import ReportEmailTemplateFragment
 from artemis.reporting.utils import get_target_url, get_top_level_target
 
 logger = utils.build_logger(__name__)
+
+
+def _get_cpe(slug: str) -> Optional[str]:
+    return lookup_cpe_by_plugin_slug(slug, "wordpress")
 
 
 class WordpressPluginsReporter(Reporter):
@@ -192,6 +197,7 @@ class WordpressPluginsReporter(Reporter):
                 name=get_target_url(task_result),
                 additional_type="wordpress-plugin:" + slug,
                 version=data.get("version", ""),
+                cpe=_get_cpe(slug),
             )
             for slug, data in task_result["result"].get("plugins", {}).items()
         ]
--- a/test/unit/test_wordpress_plugin_asset_cpe.py
+++ b/test/unit/test_wordpress_plugin_asset_cpe.py
@@ -1,0 +1,70 @@
+import unittest
+from typing import Any, Dict, List, Optional
+from unittest.mock import patch
+
+from artemis.reporting.base.asset import Asset
+from artemis.reporting.base.asset_type import AssetType
+from artemis.reporting.modules.wordpress_plugins.reporter import (
+    WordpressPluginsReporter,
+)
+
+# An IP is used instead of a domain so that Asset.__post_init__ doesn't perform a DNS lookup.
+TARGET = "http://127.0.0.1:80/"
+
+CONTACT_FORM_7_CPE = "cpe:2.3:a:rocklobster:contact_form_7:*:*:*:*:*:wordpress:*:*"
+
+# What artemis.cpe_tools would answer, so these tests don't depend on the NVD dictionary having been
+# downloaded - that mapping is built and tested there, this module only has to consult it correctly.
+FAKE_INDEX = {"contact-form-7": CONTACT_FORM_7_CPE}
+
+
+def _fake_lookup(slug: str, cms: str, version: Optional[str] = None) -> Optional[str]:
+    assert cms == "wordpress", "a WordPress plugin must be looked up under the WordPress namespace"
+    return FAKE_INDEX.get(slug.strip().lower())
+
+
+def _assets(plugins: Dict[str, Any]) -> List[Asset]:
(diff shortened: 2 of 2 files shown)


artemis/reporting/modules/wordpress_plugins/reporter.py
```diff
@@ -183,6 +188,7 @@ def get_assets(task_result: Dict[str, Any]) -> List[Asset]:
                 name=get_target_url(task_result),
                 additional_type="wordpress-plugin:" + slug,
                 version=data.get("version", ""),
+                cpe=_get_cpe(slug),
```
kadewu: we can just directly call it and already include the version even if we want to add this in separate pr

```suggestion
                cpe=lookup_cpe_by_plugin_slug(slug, "wordpress", version=data.get("version", None)),
```

kadewu, line 191: we can just directly call it and already include the version even if we want to add this in separate pr

```suggestion
                cpe=lookup_cpe_by_plugin_slug(slug, "wordpress", version=data.get("version", None)),
```

test/unit/test_wordpress_plugin_asset_cpe.py
```diff
@@ -0,0 +1,172 @@
...
+        self.assertIsNone(_get_cpe("kirki"))
+
+    def test_unusable_slug_yields_no_cpe(self) -> None:
+        # Slugs also come from scraping the scanned site, so they aren't guaranteed to be well-formed.
+        for slug in ["", "   ", "-", "with space", "wp:plugin", "../../etc/passwd"]:
+            with self.subTest(slug=slug):
+                self.assertIsNone(_get_cpe(slug))
+
+
+class EndToEndThroughCpeToolsTest(unittest.TestCase):
```
kadewu: we control cpe_tools, so no need to duplicate tests that we already have

test/unit/test_wordpress_plugin_asset_cpe.py
```diff
@@ -0,0 +1,172 @@
...
+        self.assertIsNone(asset.cpe)
+
+    def test_software_without_a_plugin_reference_is_not_reachable_by_slug(self) -> None:
+        # A CPE is only a plugin's if the dictionary says so through a wordpress.org reference; matching
+        # on the name alone would hand "nginx" to any site running a plugin with that slug.
+        (asset,) = _assets({"nginx": {"version": "1.25.3"}})
+        self.assertIsNone(asset.cpe)
+
+
+class RealLookupContractTest(unittest.TestCase):
```
kadewu: Same here

kadewu (commented):
I've started adding comments to tests, but just a general comment would work here. We are not working with external service here, but with our library that is already tested. No reason to duplicate tests here - especially that it can be missleading. If you want to add test, then the most reasonable place to do that would be like other reporters (or maybe even we need to setup properly for e2e here - but I think it can be done in seperate PR, as adding CPE right now hold the value)

test/unit/test_wordpress_plugin_asset_cpe.py
X-R3Y24: Agreed, removed both. `cpe_tools` has its own tests, no reason to duplicate them here.
What's left only covers this module's own behaviour, with the lookup patched.

artemis/reporting/modules/wordpress_plugins/reporter.py
X-R3Y24: On the version: I was asked last week to keep the CPE version-less - the detected
version is already in `Asset.version`, and `Asset.cpe` documents the version slot as
wildcarded, with consumers combining the two (`cve_lookup` does exactly that before
querying NVD). So I left it as is.

Happy to switch if you would rather go the other way, but then `wp_scanner` (#3084)
would need the same treatment, since it does pass the version; probably worth settling
that separately.

Kept `_get_cpe` as a thin wrapper because the tests patch it, but I can inline it if
you prefer.

kadewu: I'm not sure if I follow. We want to add version to all CPE, or not?

test/unit/test_wordpress_plugin_asset_cpe.py
kadewu: The rest of comment about tests still applies here. They are not really testing anything beside 'if mocked method is working' basically reducing to one testing behaviour 'is the lookup_cpe called'

artemis/reporting/modules/wordpress_plugins/reporter.py
X-R3Y24: Not to any of them. Versions are getting added in a different place,
i.e. `Asset.cpe` stays version-less and consumers combine it with `Asset.version`
when they need a versioned CPE (`cve_lookup` already does that).

So this PR keeps it wildcarded. `wp_scanner` (#3084) is the odd one out, it does
pass the version. I'm cleaning that up in the version-unification PR.

test/unit/test_wordpress_plugin_asset_cpe.py
X-R3Y24: Dropped `GetCPETest`. Kept the two in `WordpressPluginAssetCPETest`: they don't test
the lookup, they test what this reporter puts on the Asset - CPE in `cpe`, detected
version in `version`, and a plugin without a CPE still reported as an asset. That's
this module's contract rather than `cpe_tools`'. If you'd prefer them in e2e instead,
just let me know.

On `_get_cpe`: kept it, since `nuclei/reporter.py` has the same one-line wrapper -
following the existing convention here.

kadewu: If we want to test the assests as whole, I would suggest to do it consistently with already existing tests under test/reporting.

Regarding `_get_cpe`, I still don't see a value in having such overhead method. Just because it's done somwhere already, doesn't mean it's the way to go.

Both is nitpicking from my site, so I'm leaving approve, but would be nice to have those changes to clean a code a little.

kadewu (dismissed):

test/unit/test_wordpress_plugin_asset_cpe.py
X-R3Y24: Dropped `_get_cpe`.

On moving the tests: the ones under test/reporting are `BaseReportingTest` integration
tests - they scan a live container through `obtain_webapp_task_result`. Doing it that way
here also needs the NVD dictionary, and docker-compose.test.yaml has neither
`cpe-nvd-service` nor the volume, so the CPE assertion would pass locally and check nothing
in CI.

Though I can do it similarly to `cve_lookup` with `test-mock-nvd` - a small fixture under
test/data/ with `CPE_NVD_DIR` pointing at it. That would cover the `wp_scanner` CPE as well.
Separate PR or here, whichever you prefer.

PR #3083, opened 2026-08-28
hot fix joomla version issue before proper fix is done

kadewu:
https://github.com/endoflife-date/endoflife.date/issues/10831

### DIFF
--- a/artemis/modules/joomla_scanner.py
+++ b/artemis/modules/joomla_scanner.py
@@ -1,4 +1,5 @@
 import re
+from datetime import date
 from typing import Any, Dict, List, Union
 
 from karton.core import Task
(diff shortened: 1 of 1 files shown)


artemis/modules/joomla_scanner.py
```diff
@@ -54,6 +55,9 @@ def run(self, current_task: Task) -> None:
                 # Get latest release in repo from GitHub API
 
                 if gh_api_response.json()["tag_name"] != joomla_version and self.is_version_obsolete(joomla_version):
+                    # FIXME: reason https://github.com/endoflife-date/endoflife.date/issues/10831
+                    if joomla_version == "5.4.8" and date.today() < date(2026, 9, 11):
```
kazet:
```suggestion
                    if joomla_version == "5.4.8" and date.today() < date(2026, 10, 11):
```

PR #3102, opened 2026-09-03
Unify CPE version filling between cve_lookup and cpe_tools

X-R3Y24:
## Summary

Unifies the two independent implementations of "insert a version into a CPE" into a
single `with_version()` in `artemis/cpe_tools/cpe_utils.py`, and fixes a splitting bug
that made the `cve_lookup` copy read the wrong field on ~5,000 NVD CPE names.

## Why

`Asset.cpe` holds a wildcard in the version slot by definition; the version lives
separately in `Asset.version`. Setting the slot is the consumer's job, done right before
use - e.g. `cve_lookup`, before querying NVD. That job had two divergent implementations:

| | `cve_lookup._fill_cpe_version` | `cpe_utils.with_version` |
|---|---|---|
| splitting | `cpe.split(":")` - **wrong**, see below | `split_cpe()` - correct |
| validates the version | yes (regex) | no |

CPE 2.3 escapes a colon inside a field (`cpe:2.3:a:cgiirc:cgi\:irc:0.5.7:...`), so a
naive `split(":")` yields 14 components instead of 13 and index 5 lands on part of the
product name rather than the version. This currently affects **5,093 of 1,823,483** names
in the NVD dictionary (escaped colons occur in fields 3/4/5 - 815/4,993/100 respectively).
`split_cpe()` returns exactly 13 components for all 1,823,483 (as of now).

The user-visible consequence in `cve_lookup` was worse than a missing version: for those
products `_has_concrete_version` read the product text (`irc`) as the version and returned
`True`, so the module queried NVD with a **wildcard-version** `cpeName`, which NVD answers
with 404 - recorded as a *failed lookup* rather than a skip. Those products never got a
real CVE lookup at all.

## What changed

- **One `with_version(cpe, version)`** in `cpe_utils.py` - correct splitting plus version
  validation. `_fill_cpe_version` is gone; `cve_lookup` and the three `lookup_*` helpers
  all go through it.
- The old "don't overwrite a slot that already holds a version" guard is dropped: it never
  fired. Currently in wappalyzergo every fingerprint with a CPE carries `*` in the version
  slot, and the `lookup_*` path widens to the family first, so the slot is always a
  wildcard by the time it gets there.
- **The version regex now also accepts `*` (ANY) and `-` (NA)**, the two special field
  values CPE 2.3 defines. Without them, widening a dictionary CPE back to its versionless
  family would stop working once the setter validates its input.
- `lookup_*` behaviour: an invalid version (`"latest"`) is now rejected and the slot stays
  wildcarded, where before it was substituted verbatim. No production caller passes a
  version today - #3103 removed the last one from `wp_scanner` - so nothing changes in
  practice.
- **`cve_lookup`**: `_fill_cpe_version` and `_CPE_VERSION_RE` removed; `_has_concrete_version`
  and `_cpe_product_key` now use `split_cpe()` too. `_has_concrete_version` keeps its own
  separate responsibility (whether to query NVD at all).
- **Version regex anchored with `\Z` instead of `$`**: in Python `$` also matches before a
  trailing newline, so `"1.0\n"` passed validation and put a newline inside the CPE.

## Tests

`WithVersionTest` covers the escaped colon (both directions), invalid versions
(`latest`, `v2.4`, `1.0:extra`, leading space, trailing newline), an empty/`None` version,
the special values `*` and `-`, replacing a slot that already holds a version, and a CPE
shorter than six fields. The escaped-colon case for `_has_concrete_version` is pinned where
that function lives.

The escaped-colon cases fail against the old naive split - that's the regression they guard.

### DIFF
--- a/artemis/cpe_tools/cpe_utils.py
+++ b/artemis/cpe_tools/cpe_utils.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from pathlib import Path
 
 from artemis.cpe_tools.cpe_main_process import (
@@ -36,17 +37,50 @@ _STOPWORDS = frozenset(
 )
 
 
-def with_version(cpe: str, version: str) -> str:
-    # Components are: cpe, 2.3, part, vendor, product, version, update, edition,
-    # language, sw_edition, target_sw, target_hw, other. Replace the version field
-    # (index 5) and re-join;
-    VERSION_FIELD_INDEX = 5
+# Components of a cpe:2.3 name are: cpe, 2.3, part, vendor, product, version, update,
+# edition, language, sw_edition, target_sw, target_hw, other.
+_VERSION_FIELD_INDEX = 5
 
+# A version slot must look like a real version: it starts with a digit and carries only
+# characters a version is made of. The version reaching ``fill_version`` is whatever a
+# scanner reported next to the technology, so guarding against a stray value such as
+# "latest" - or anything carrying a ":", which would silently add a component - keeps us
+# from building a malformed CPE that NVD matches nothing for. The match is anchored with
+# ``\Z`` rather than ``$``, which in Python also matches before a trailing newline and
+# would let "1.0\n" through into the CPE.
+_VERSION_RE = re.compile(r"^[0-9][0-9A-Za-z.\-+]*\Z")
+
+
+def with_version(cpe: str, version: str) -> str:
+    """Replace the version field of a cpe:2.3 name with ``version``, whatever it holds now.
+
+    Used to normalize a dictionary CPE down to its wildcard family. To insert a version
+    that a scanner reported, use ``fill_version`` instead - it validates the value and
+    leaves an already-concrete slot alone.
+    """
     parts = split_cpe(cpe)
-    if len(parts) <= VERSION_FIELD_INDEX:
+    if len(parts) <= _VERSION_FIELD_INDEX:
         return cpe
-    parts[VERSION_FIELD_INDEX] = version
+    parts[_VERSION_FIELD_INDEX] = version
     return ":".join(parts)
+
+
+def fill_version(cpe: str, version: str | None) -> str:
+    """Substitute a scanner-reported version into the wildcard version slot of a cpe:2.3 name.
+
+    A CPE we store (e.g. ``Asset.cpe``) keeps ``*`` in the version slot by definition, with
+    the version living alongside it, so filling the slot is the consumer's job - done here,
+    right before the CPE is used to query something.
+
+    The CPE comes back unchanged when the version is missing, doesn't look like a version,
+    or the slot already carries something concrete that we would rather not contradict.
+    """
+    if not version or not _VERSION_RE.match(version):
+        return cpe
+    parts = split_cpe(cpe)
+    if len(parts) <= _VERSION_FIELD_INDEX or parts[_VERSION_FIELD_INDEX] != "*":
+        return cpe
+    return with_version(cpe, version)
 
 
 def resolve(nvd_dir: Path, normalized: str) -> str | None:
@@ -87,7 +121,7 @@ def lookup_cpe(name: str, version: str | None = None) -> str | None:
     family_cpe = resolve(nvd_dir, normalized)
     if family_cpe is None:
         return None
-    return with_version(family_cpe, version) if version else family_cpe
+    return fill_version(family_cpe, version)
 
 
 def lookup_cpe_by_plugin_slug(slug: str, cms: str, version: str | None = None) -> str | None:
@@ -104,7 +138,7 @@ def lookup_cpe_by_plugin_slug(slug: str, cms: str, version: str | None = None) -
     cpe = plugins.get(f"{cms}:{slug.strip().lower()}")
     if cpe is None:
         return None
-    return with_version(cpe, version) if version else with_version(cpe, "*")
+    return fill_version(with_version(cpe, "*"), version)
 
 
 def lookup_cpe_by_url(url: str, version: str | None = None) -> str | None:
@@ -126,4 +160,4 @@ def lookup_cpe_by_url(url: str, version: str | None = None) -> str | None:
     cpe = urls.get(key)
     if cpe is None:
         return None
-    return with_version(cpe, version) if version else with_version(cpe, "*")
+    return fill_version(with_version(cpe, "*"), version)
--- a/artemis/cpe_tools/cpe_utils.py
+++ b/artemis/cpe_tools/cpe_utils.py
@@ -41,21 +41,16 @@ _STOPWORDS = frozenset(
 # edition, language, sw_edition, target_sw, target_hw, other.
 _VERSION_FIELD_INDEX = 5
 
-# A version slot must look like a real version: it starts with a digit and carries only
-# characters a version is made of. The version reaching ``fill_version`` is whatever a
-# scanner reported next to the technology, so guarding against a stray value such as
-# "latest" - or anything carrying a ":", which would silently add a component - keeps us
-# from building a malformed CPE that NVD matches nothing for. The match is anchored with
-# ``\Z`` rather than ``$``, which in Python also matches before a trailing newline and
-# would let "1.0\n" through into the CPE.
+# A version has to look like one: a digit first, then only characters versions are made of.
+# Anchored with ``\Z``, because Python's ``$`` also matches before a trailing newline.
 _VERSION_RE = re.compile(r"^[0-9][0-9A-Za-z.\-+]*\Z")
 
 
 def with_version(cpe: str, version: str) -> str:
     """Replace the version field of a cpe:2.3 name with ``version``, whatever it holds now.
 
-    Used to normalize a dictionary CPE down to its wildcard family. To insert a version
-    that a scanner reported, use ``fill_version`` instead - it validates the value and
+    Used to normalize a dictionary CPE down to its wildcard family. To insert a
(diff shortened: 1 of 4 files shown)


artemis/cpe_tools/cpe_utils.py
```diff
@@ -36,19 +37,52 @@
 )
 
 
+# Components of a cpe:2.3 name are: cpe, 2.3, part, vendor, product, version, update,
+# edition, language, sw_edition, target_sw, target_hw, other.
+_VERSION_FIELD_INDEX = 5
+
+# A version slot must look like a real version: it starts with a digit and carries only
```
kadewu: Can we shorten that comment a little? I think just one sentence should already describe enough the purpose of that regex.

artemis/cpe_tools/cpe_utils.py
```diff
@@ -36,19 +37,52 @@
...
-    if len(parts) <= VERSION_FIELD_INDEX:
+    if len(parts) <= _VERSION_FIELD_INDEX:
         return cpe
-    parts[VERSION_FIELD_INDEX] = version
+    parts[_VERSION_FIELD_INDEX] = version
     return ":".join(parts)
 
 
+def fill_version(cpe: str, version: str | None) -> str:
+    """Substitute a scanner-reported version into the wildcard version slot of a cpe:2.3 name.
```
kadewu: I'm not sure about moving this method here and leaving the `with_version`, those are doing literally the same thing, with one side_effect difference. I would suggest to refactor that.

And one nitpick about comments: We should not in such place mention anything about outside code that is using this code. This place should and do not know anything about it, e.g: `We protect against malfunctioned version` is much better than saying that it comes from some scanner. And reason for that is simple, if in the future we change the scanner, nobody would remmeber to update the comment.

artemis/cpe_tools/cpe_utils.py
X-R3Y24: Done.

artemis/cpe_tools/cpe_utils.py
X-R3Y24: Agreed with the second part. 

As for the first part they aren't the same.
- `with_version(cpe, "*")`
erases a concrete version from a dictionary CPE to get the family - that's how `resolve`
and both `lookup_cpe_by_*` normalise their results.
- `fill_version` does the reverse, and
only ever writes into a slot that is still a wildcard, after validating the value.
Merging them would need a flag to pick which of the two behaviours you want, which reads
worse than two named functions.

kadewu: Both are replacing the version in cpe. I'm not sure if the guard in fill_version is needed at all, can be checked.
But just becasue 'with_version' with specific call is used as normalization, doesn't mean that method purpose is only for normalization.

X-R3Y24: Unified, guard removed. Checked it first: currently in wappalyzergo all fingerprints with a CPE carry `*` in the version slot, so it never fired.

The regex now also accepts `*` (ANY) and `-` (NA), otherwise widening a dictionary CPE
back to its family would stop working.

artemis/cpe_tools/cpe_utils.py
```diff
@@ -36,16 +37,32 @@
...
 
 
-def with_version(cpe: str, version: str) -> str:
-    # Components are: cpe, 2.3, part, vendor, product, version, update, edition,
-    # language, sw_edition, target_sw, target_hw, other. Replace the version field
-    # (index 5) and re-join;
-    VERSION_FIELD_INDEX = 5
+# Components of a cpe:2.3 name are: cpe, 2.3, part, vendor, product, version, update,
+# edition, language, sw_edition, target_sw, target_hw, other.
+_VERSION_FIELD_INDEX = 5
```
kadewu: Can be moved back to the method, as it's used only by one method.

artemis/cpe_tools/cpe_utils.py
```diff
@@ -36,16 +37,32 @@
...
+
+def with_version(cpe: str, version: str | None) -> str:
+    """Set the version field of a cpe:2.3 name to ``version``.
+
+    A name carrying ``*`` in the version slot denotes the product as a whole; setting that
+    slot narrows it to a single release, and setting it back to ``*`` widens it again.
+
+    The CPE comes back unchanged when the version is missing, doesn't look like a version,
+    or the name is too short to have a version field.
+    """
```
kadewu:
```suggestion
        """Set the CPE 2.3 version field, leaving invalid inputs unchanged."""
```

artemis/cpe_tools/cpe_utils.py
```diff
@@ -104,7 +121,8 @@ def lookup_cpe_by_plugin_slug(slug: str, cms: str, version: str | None = None) -
     cpe = plugins.get(f"{cms}:{slug.strip().lower()}")
     if cpe is None:
         return None
-    return with_version(cpe, version) if version else with_version(cpe, "*")
```
kadewu: I would rather lean to not change this line now.
It's always one less unnecessary method call.

Same below

artemis/cpe_tools/cpe_utils.py
```diff
@@ -36,16 +37,32 @@
...
-    VERSION_FIELD_INDEX = 5
+# Components of a cpe:2.3 name are: cpe, 2.3, part, vendor, product, version, update,
+# edition, language, sw_edition, target_sw, target_hw, other.
+_VERSION_FIELD_INDEX = 5
 
+# A version has to look like one: a digit first, then only characters versions are made of.
+# ``*`` (ANY) and ``-`` (NA) are the two special values CPE 2.3 defines for a field, and are
+# accepted so that a name can also be reset to its versionless family.
+# Anchored with ``\Z``, because Python's ``$`` also matches before a trailing newline.
+_VERSION_RE = re.compile(r"^(?:[0-9][0-9A-Za-z.\-+]*|\*|-)\Z")
```
kadewu: For me slightly easier to read this regex, but it's just opinion (and I dropped ^, as we use it only with match)
```suggestion
_VERSION_RE = re.compile(r"(?:[0-9][0-9A-Za-z.+-]*|[*-])\Z")
```

artemis/cpe_tools/cpe_utils.py
kadewu, line 42: nitpick: Can be moved back to the method, as it initially was.

artemis/cpe_tools/cpe_utils.py
kadewu, line 48: For me slightly easier to read this regex, but it's just opinion (and I dropped ^, as we use it only with match)
```suggestion
_VERSION_RE = re.compile(r"(?:[0-9][0-9A-Za-z.+-]*|[*-])\Z")
```

artemis/cpe_tools/cpe_utils.py
kadewu, line 59:
```suggestion
        """Set the CPE 2.3 version field, leaving invalid inputs unchanged."""
```

X-R3Y24: Done!

artemis/cpe_tools/cpe_utils.py
X-R3Y24: Done!

artemis/cpe_tools/cpe_utils.py
X-R3Y24: Done!

artemis/cpe_tools/cpe_utils.py
```diff
@@ -104,7 +94,8 @@ def lookup_cpe_by_plugin_slug(slug: str, cms: str, version: str | None = None) -
     cpe = plugins.get(f"{cms}:{slug.strip().lower()}")
     if cpe is None:
         return None
-    return with_version(cpe, version) if version else with_version(cpe, "*")
+    family = with_version(cpe, "*")
```
kazet, line 97: why there are two consecutive with_version  calls?

artemis/cpe_tools/cpe_main_process.py
```diff
@@ -109,15 +156,36 @@ def _iter_entries(chunks_dir: Path) -> Iterator[tuple[str, str, list[str]]]:
...
+    """Add one title claim to the title index, resolving collisions between families.
+
+    Cutting the version off merges the releases of a product, which is the point, but it
+    also merges products that differ only by version; a key several families claim is
+    marked ambiguous rather than given to whichever chunk sorted first. Claims are
+    ranked, though: an untrimmed title is a name NVD published, a trimmed key is our own
+    inference, so a full title outranks every derived one and only equals can collide.
+    """
+    previous = titles.get(key)
+    outranks = names_a_product and key not in named
```
kazet: "outranks" and "outranked" makes it nonnecessarily complex, can it be simplified?

X-R3Y24: Simplified - one flag instead of two. The `not outranked` case turned out to mean "both
claims are of the same rank", so it collapses into `names_a_product == previously_named`,
which is what the docstring says anyway ("only equals can collide").

Rebuilt the index both ways on the real feed, zero differences.

artemis/cpe_tools/cpe_utils.py
X-R3Y24: The first call widens the dictionary CPE to its family; the second sets the caller's
version. It matters when the version is absent or invalid - the second call then returns
its argument unchanged, so what comes back is the family with `*` rather than whatever
concrete version the dictionary entry happened to carry.

PR #3109, opened 2026-09-07
Classifier: require full match for ASN detection

### LINKED ISSUE #3108: Classifier: any input starting with "as<digits>" is misrouted to the ASN path (crash-retry loop on valid domains)
## Summary

The ASN detection in `artemis/modules/classifier.py` uses `re.match(ASN_REGEX, data)`, which is **prefix-anchored only** - it succeeds if the input merely *starts* with `as<digits>`, not if the whole input *is* an ASN:

```python
ASN_REGEX = "[aA][sS][0-9][0-9]*"          # line 17
...
if re.match(ASN_REGEX, data):              # lines 91, 123, 221
```

## Reproducing input

Any user-supplied domain whose first label happens to start with `as` followed by digits - e.g. **`as7822.example.com`** - triggers this. Plenty of real domains look like this (hostnames like `as2001.corp.example.org`, vanity domains starting with `as1234...`).

## What happens

1. In `run()` (line 221), the prefix match succeeds, so the input is treated as an ASN and passed to `get_ip_prefixes_for_asn()` (line 223).
2. That builds a RIPEstat request with the domain as the `resource` parameter - the API rejects it, and the task ends up in an error/retry loop, hammering the RIPEstat API each time.

So a perfectly valid domain never gets scanned: it's misrouted to a code path that can only fail. The same false positive also affects `is_supported()` (line 91) and `is_service_target()` (line 123), though there it's masked - both fall through to the domain checks afterwards and happen to reach the same final answer.

## Proposed approach

Switch `re.match` -> `re.fullmatch` at the three call sites:

```diff
-if re.match(ASN_REGEX, data):
+if re.fullmatch(ASN_REGEX, data):
```

Why I think this is safe - I traced all three call paths:

- **Real ASNs are unaffected:** `AS123` / `as48282` still fullmatch (the regex keeps its case-insensitive prefix, so I left `ASN_REGEX` itself untouched rather than lowercasing it - `is_supported()` is also called externally with mixed-case input like `"AS123"`, which I confirmed in `test/modules/test_classifier.py`).
- **Domains like `as7822.example.com` fall through to `_is_ip_or_domain()`**, which already handles them correctly as domains - they just previously never got that far.
- **No other behavior changes:** for `is_supported`/`is_service_target`, the *result* is identical for every legitimate input; only the routing reason changes. The only inputs whose behavior differs are exactly the broken ones.

I've implemented this on a branch with a regression test that runs the full module on `as7822.example.com` with `get_ip_prefixes_for_asn` patched and asserts it is never called (previously it would be, and error out), plus a positive control that `AS123` still takes the ASN path. All 10 classifier tests pass and black/isort/mypy/flake8 are clean.

Would you be open to a PR along these lines? And of course, if you'd prefer a different approach - e.g. validating with `ipaddress`-style parsing, or moving the ASN check after the domain classification instead - I'd be happy to follow your lead instead, as you know this code far better than I do. Just let me know how you'd like it handled and I'll happily adjust.

AdeshDeshmukh:
Fixes #3108

## Problem

The ASN detection in the classifier used `re.match(ASN_REGEX, data)`, which is prefix-anchored only. Any input merely *starting* with `as<digits>` - e.g. the valid domain `as7822.example.com` - was misrouted to the ASN path: `get_ip_prefixes_for_asn()` built a RIPEstat request with the domain as the `resource` parameter, the API rejected it, and the task ended up in an error/retry loop, hammering the RIPEstat API each time. Valid domains like this were therefore never scanned.

## Fix

`re.match` -> `re.fullmatch` at the three call sites (`is_supported`, `is_service_target`, `run`):

```diff
-if re.match(ASN_REGEX, data):
+if re.fullmatch(ASN_REGEX, data):
```

- Real ASNs (`AS123`, `as48282`) are unaffected - the regex keeps its case-insensitive prefix and `ASN_REGEX` itself is untouched, since `is_supported()` is also called externally with mixed-case input.
- Domains like `as7822.example.com` now fall through to `_is_ip_or_domain()` and take the normal domain flow.
- For `is_supported`/`is_service_target` the result is identical for every legitimate input; only the broken inputs change behavior.

## Tests

Added to `test/modules/test_classifier.py`:

- `test_asn_prefix_input_is_classified_as_domain` - runs the full module on `as7822.example.com` with `get_ip_prefixes_for_asn` patched and asserts it is **never called** (previously it would be, and error out); asserts the input routes to the domain path (`domain_that_may_not_exist`).
- `test_asn_still_detected` - positive control: `AS123` / `as48282` still fullmatch, `as7822.example.com` does not.
- Assertions for `is_supported("as7822.example.com")` (True) and `is_service_target("as7822.example.com")` (False).

## Verification

- [x] `pytest test/modules/test_classifier.py` - 10/10 passed
- [x] `pre-commit run --files artemis/modules/classifier.py test/modules/test_classifier.py` - black, isort, mypy, flake8 all pass

### DIFF
--- a/artemis/modules/classifier.py
+++ b/artemis/modules/classifier.py
@@ -88,7 +88,7 @@ class Classifier(ArtemisBase):
         if "://" in data:
             return direct_url_scanning.is_scannable_url(data)
 
-        if re.match(ASN_REGEX, data):
+        if re.fullmatch(ASN_REGEX, data):
             return True
 
         if to_ip_range(data):
@@ -120,7 +120,7 @@ class Classifier(ArtemisBase):
             return False
         if "://" in data:
             return True
-        if re.match(ASN_REGEX, data) or to_ip_range(data):
+        if re.fullmatch(ASN_REGEX, data) or to_ip_range(data):
             return False
         return not Classifier._is_ip_or_domain(Classifier._clean_ipv6_brackets(data))
 
@@ -218,7 +218,7 @@ class Classifier(ArtemisBase):
             )
             return
 
-        if re.match(ASN_REGEX, data):
+        if re.fullmatch(ASN_REGEX, data):
             ips = []
             for prefix in get_ip_prefixes_for_asn(data):
                 self.log.info(f"Converted {data} to IP prefixes, processing {prefix}")
--- a/test/modules/test_classifier.py
+++ b/test/modules/test_classifier.py
@@ -1,3 +1,4 @@
+import re
 from test.base import ArtemisModuleTestCase
 from typing import Any, Dict, List, NamedTuple
 from unittest.mock import patch
@@ -6,7 +7,7 @@ from karton.core import Task
 
 from artemis import direct_url_scanning
 from artemis.binds import TaskType
-from artemis.modules.classifier import Classifier
+from artemis.modules.classifier import ASN_REGEX, Classifier
 
 
 class ExpectedTaskData(NamedTuple):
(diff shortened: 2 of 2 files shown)


AdeshDeshmukh:
@kazet PTAL

AdeshDeshmukh:
@kazet , as per ur suggestions i opened a PR. PTAL

PR #3128, opened 2026-09-08
Key the title index by product rather than release

X-R3Y24:
## What

Rebuilds the NVD title index in `artemis/cpe_tools/cpe_main_process.py` so that it holds
one entry per product instead of one per release, cutting it from 1 715 473 entries to
236 229 (160 MB -> 20.6 MB on disk, 368 MB -> 65 MB in memory).

## Why

`_build_indices` keyed the title index by `normalize(title)`, but an NVD title names a
single release - "JoomlaWorks K2 2.8.0 for Joomla!", "Vim 9.0.2176". Every release of
every product therefore got its own key, while the feed only contains 150 146 distinct
`vendor:product` families. Three consequences:

- `title-index.json` weighed 160 MB and the parsed dict 368 MB of RSS, in every process
that does CPE lookups (autoreporter and each reporting module, not just the service).
- A miss falls through to the token-subset scan in `resolve()`, which walks the whole
index: ~1 s per unresolved product name.
- 94% of the stored values carried a concrete version that the caller overwrote on the
very next line, so the index was paying for data nobody read.

## Fix

`title_key()` derives the key by removing the version from the title. The version is
**not** guessed from the title - product names are full of digits ("Windows Server 2022",
"Cisco IP Phone 7960") - it is read from the `version` component of the CPE name and that
exact token run is deleted, scanning right to left. Versionless CPEs carry `-` or `*`
there, which normalize to nothing, so their titles are left untouched. Values are stored
as `with_version(cpe, "*")`, so a hit needs no rewriting (`with_version` moved from
`cpe_utils` to `cpe_main_process`; the build needs it and the reverse import would be
circular).

**Key design decision - collisions.** Cutting the version off merges the releases of one
product, which is the point, but it also merges products that differ only by version
("Google Chrome 1.0" exists as both `a:google:chrome` and `o:google:chrome`). Keeping the
first family read would answer such a query with whichever chunk sorted first, i.e.
arbitrarily. Such a key is instead marked ambiguous and reports no match - the same answer
`resolve()` already gives when a query matches several families.

Claims are **ranked** before that rule applies: a title that survived trimming intact is a
name NVD published for a product, whereas a key left over after cutting a release off a
longer title is our own inference. A full-title claim outranks every derived one, and only
claims of equal standing can make a key ambiguous. Without the ranking more keys would
become ambiguous, and the casualties would include "Google Chrome" and "Apache Software
Foundation Tomcat" - names the index answers today and that NVD does publish at product
level.

Measured on the real feed under `artemis/cpe_tools/nvdcpe-2.0/`:

| | before | after |
| ---------------------- | ----------- | --------------------------- |
| entries | 1 715 473 | 236 229 (1 848 ambiguous) |
| `title-index.json` | 160 MB | 20.6 MB |
| parsed dict (RSS) | 367.6 MB | 65.4 MB |
| unresolved lookup | 1.1-1.2 s | 0.13-0.16 s |

`lookup_cpe("WordPress")` and `lookup_cpe("HikaShop")` return what they returned before;
`lookup_cpe("AcyMailing")` stays `None`, and that is not an index defect - NVD ships two
vendors, `a:acyba:acymailing` and `a:acymailing:acymailing`, so the query is genuinely
ambiguous. The honest cost of the change: 1848 keys in the new index are marked ambiguous
and report no match rather than an arbitrary pick. Looking through them, they are dominated
by hardware and industrial software this scanner never detects - ARM Cortex parts, 3Com
switches, ABB panels, SCADA software - plus names that are genuinely ambiguous in NVD
(`adobe flash` vs `adobe flash player`, or a bare `acl`). Nothing on the list is something
a web scan produces.

## Tests

- `test_title_index_is_keyed_per_product_not_per_release` - asserts on the index itself via
`ensure_title_index()`: two releases produce one key, the key is the title without the
version, the versioned keys are absent, and the value has `*` in the version slot.
Verified by mutation - reverting the trimming in `title_key` makes this test, and only
this test, fail; the lookup-level tests pass either way because `resolve()`'s
token-subset fallback hides the difference.
- `test_version_stripped_from_title_key` - verifies two releases of one product resolve
through the single shared key, with and without a caller-supplied version.
- `test_digits_in_product_name_are_kept` - verifies "Microsoft Windows Server 2022" keeps
the 2022 that belongs to the name while losing the version that does not.
- `test_titles_differing_only_by_version_do_not_become_ambiguous` - verifies merging
releases of one family stays a single unambiguous answer.
- `test_version_stripping_can_make_a_title_ambiguous` - verifies a key claimed by two
families reports no match, and that a query that is only a subset of it does too.
- `test_title_naming_a_product_outranks_a_trimmed_one` - verifies the ranking rule: the
versionless entry wins over the key left over after trimming a different family.
- `test_duplicated_title_picks_first` renamed to `test_duplicated_title_returns_none` -
the intended behaviour change, previously it asserted the arbitrary first-seen answer.

## Notes / Out of scope

- `_build_indices` fills three indices in one loop; only the title one changed. The plugin
index (keyed by CMS slug) and the URL index are untouched - the version problem does not
apply to either.
- The `update` component (`beta`, `rc1`, `sp1`) is deliberately **not** stripped. Doing so
would take the index to ~208 000 keys, but every extra field trimmed is another chance to
eat a word that belongs to a product name, and it buys 12% on a number already reduced
7.3x. Possible follow-up if index size becomes a constraint again.
- `_strip_sw_edition` is unchanged, so family CPEs can still carry a `target_sw`
(`...:k2:*:*:*:*:*:joomla\!:*:*`). That is pre-existing behaviour, not something this PR
set out to normalize.
- The token-subset fallback in `resolve()` is unchanged; it only learned to treat an
ambiguous key as a no-match. `cpe_utils.py` is otherwise touched only to stop rewriting
a version the index now stores wildcarded.
- Touches `with_version`, which #3102 also rewrites; will rebase once that lands.

## Before deploying

The index sitting in the `cpe-nvd-data` volume needs to be cleared once. `build_version_token`
is a hash of the index *contents*, so it cannot notice that the keying scheme changed, and it
is only recomputed by `build_index()` - an index left there by an older Artemis would keep
being served until the next refresh cycle (up to `CPE_NVD_REFRESH_INTERVAL_SECONDS`, default
24 h). Since `resolve()` no longer rewrites the version, it would hand back CPEs like
`cpe:2.3:a:wordpress:wordpress:-:...` instead of the family. Deleting the index files from the
volume is enough - otherwise that refresh cycle fixes it on its own.


artemis/cpe_tools/cpe_main_process.py
```diff
@@ -29,6 +29,24 @@
 URL = "url"
 _INDEX_KINDS = {TITLE: INDEX_TITLE_FILENAME, PLUGIN: PLUGIN_INDEX_FILENAME, URL: URL_INDEX_FILENAME}
 
+# Bump whenever the keys or values of an index change. The version token is a hash of the
+# index contents, so it cannot notice that: readers compare this number instead, and
+# rebuild from the chunks when an index left in the volume does not match.
+INDEX_FORMAT_VERSION = 2
```
kadewu: Why are we doing this? I don't think this is needed.

artemis/cpe_tools/cpe_main_process.py
```diff
@@ -29,6 +29,24 @@
...
+
+# A version has to look like one: a digit first, then only characters versions are made of.
+# ``*`` (ANY) and ``-`` (NA) are the two special values CPE 2.3 defines for a field, and are
+# accepted so that a name can also be reset to its versionless family.
+# Anchored with ``\Z``, because Python's ``$`` also matches before a trailing newline.
+VERSION_RE = re.compile(r"^(?:[0-9][0-9A-Za-z.\-+]*|\*|-)\Z")
+
+# Stored in the title index instead of a CPE when a key is claimed by more than one
+# vendor:product family. Not a valid CPE, so it cannot collide with a real value.
+AMBIGUOUS_TITLE = ""
```
kadewu, line 43: why not remove from the index then?

artemis/cpe_tools/cpe_main_process.py
```diff
@@ -109,15 +169,37 @@ def _iter_entries(chunks_dir: Path) -> Iterator[tuple[str, str, list[str]]]:
...
     plugins: dict[str, str] = {}
     urls: dict[str, str] = {}
     for cpe_name, title, refs in _iter_entries(chunks_dir):
         cpe = _strip_sw_edition(cpe_name)
-        key = normalize(title)
-        if key and key not in titles:
-            titles[key] = cpe
+        key, names_a_product = title_key(title, cpe_name)
+        if key:
+            # The index holds the family CPE; the caller sets whatever version it needs.
```
kadewu: I'm not sure why this comment is here.

artemis/cpe_tools/cpe_utils.py
```diff
@@ -37,44 +37,16 @@
...
-    return ":".join(parts)
-
-
 def resolve(nvd_dir: Path, normalized: str) -> str | None:
     def _tokens(normalized: str) -> list[str]:
         return [t for t in normalized.split() if len(t) > 1 and t not in _STOPWORDS]
 
     index = ensure_title_index(nvd_dir)
 
+    # Index values are already version-wildcarded, so a hit needs no rewriting.
```
kadewu: Not sure what this comment is about? If about why we removed a line, then it's gonna be confusing in the future.

X-R3Y24: Removed

artemis/cpe_tools/cpe_main_process.py
X-R3Y24: Removed

artemis/cpe_tools/cpe_main_process.py
X-R3Y24: Tried it and measured the result: 651 of the 1,848 names stop returning `None` and start
returning a wrong CPE instead. Without the key the query falls through to the token-subset
scan in `resolve()`, which finds a longer title the query is a subset of - `ARM Cortex A15`
resolves to `o:arm:cortex-a15_firmware` instead of the hardware entry, `ASUS RT-AC51U`
likewise, `Adobe Integrated Runtime AIR SDK` lands on `air_sdk_&_compiler`.

The marker is what keeps "several families claim this name" from degrading into "here's
whichever one the fallback happens to reach". Since these CPEs feed `cve_lookup`, a wrong
product is worse than no answer.

artemis/cpe_tools/cpe_main_process.py
X-R3Y24: I checked what happens without it. `build_version_token` hashes the index contents, so an
index built by an older Artemis validates against itself: I built one with the pre-change
code and `index-version.json` carries only `{"hash": ...}`, with 1,715,473 entries keyed
per release and concrete versions in the values (`cpe:2.3:a:3com:3cdaemon:-:*`). Since
`resolve()` no longer rewrites the version, serving that index would hand those back
verbatim, and it would keep being served until the next refresh cycle - up to 24 h.

With the format number the reader notices the mismatch and rebuilds from the chunks; I
verified `lookup_cpe("WordPress")` still returns the wildcard family against a stale index.

Worth knowing: the rebuild happens in memory only, so the stale file stays on disk until
`cpe-nvd-service` rewrites it. Writing it back would save the per-process rebuild - left
out of this PR, happy to add it if you'd rather have it here.

kadewu: Yes I know how the cache mechanism here works., The thing here is that there is no reason for this overengineering. The index will be rebuilt, but if we really want to rebuild it can as well be done manually after deploy.

> Worth knowing: the rebuild happens in memory only, so the stale file stays on disk until
cpe-nvd-service rewrites it.

And that is another argument why we don't want to add such a thing to a code.

X-R3Y24: Dropped it. Added a note to the description that the volume needs clearing
once after deploy, so it doesn't get lost.

test/unit/test_cpe_utils.py
```diff
@@ -149,11 +153,112 @@ def test_ambiguous_query_returns_none(self) -> None:
         # "Widget" matches two different vendor:product families.
         self.assertIsNone(lookup_cpe("Widget"))
 
-    def test_duplicated_title_picks_first(self) -> None:
-        # The same title shared by two families resolves to the first one seen.
+    def test_duplicated_title_returns_none(self) -> None:
+        # The same title shared by two families is ambiguous, not first-one-seen.
```
kadewu, line 157: I think LLMs have tendencies to leave such comments - it's always worth double checking that. `not first-one-seen` will only generate confusion in the future if someone reads it.

kadewu (approved):
Beside things I addressed in the parent branch, LGTM

test/unit/test_cpe_utils.py
X-R3Y24: Done!

test/unit/test_cpe_utils.py
```diff
@@ -149,11 +153,112 @@ def test_ambiguous_query_returns_none(self) -> None:
...
         self.assertIsNone(lookup_cpe("Widget"))
 
-    def test_duplicated_title_picks_first(self) -> None:
-        # The same title shared by two families resolves to the first one seen.
+    def test_duplicated_title_returns_none(self) -> None:
+        # A title claimed by two families is ambiguous, so the lookup reports no match.
+        self.assertIsNone(lookup_cpe("Duplicated Product"))
+
+    def test_title_index_is_keyed_per_product_not_per_release(self) -> None:
+        # Asserted on the index itself: lookup_cpe() would also answer these through the
```
kazet: I don't quite understand this comment

artemis/cpe_tools/cpe_main_process.py
```diff
@@ -29,6 +29,19 @@
 URL = "url"
 _INDEX_KINDS = {TITLE: INDEX_TITLE_FILENAME, PLUGIN: PLUGIN_INDEX_FILENAME, URL: URL_INDEX_FILENAME}
 
+# Position of the version field in a cpe:2.3 name.
+VERSION_FIELD_INDEX = 5
+
+# A version has to look like one: a digit first, then only characters versions are made of.
```
kazet: I don't think this is correct grammar

artemis/cpe_tools/cpe_main_process.py
```diff
@@ -29,6 +29,19 @@
...
+# ``*`` (ANY) and ``-`` (NA) are the two special values CPE 2.3 defines for a field, and are
+# accepted so that a name can also be reset to its versionless family.
+# Anchored with ``\Z``, because Python's ``$`` also matches before a trailing newline.
+VERSION_RE = re.compile(r"(?:[0-9][0-9A-Za-z.+-]*|[*-])\Z")
```
kazet: can it be made simpler?

test/unit/test_cpe_utils.py
```diff
@@ -149,11 +153,112 @@ def test_ambiguous_query_returns_none(self) -> None:
...
+            ],
+        )
+        # Both releases collapse onto one key holding the version-wildcarded CPE
+        # (target_sw is part of the family and is kept).
+        self.assertEqual(
+            lookup_cpe("JoomlaWorks K2 for Joomla"),
+            "cpe:2.3:a:joomlaworks:k2:*:*:*:*:*:joomla\\!:*:*",
+        )
+        self.assertEqual(
+            lookup_cpe("JoomlaWorks K2 for Joomla", version="2.8.0"),
```
kazet: do we use lookup_cpe with version parameter?

artemis/cpe_tools/cpe_main_process.py
X-R3Y24: Not without loosening what it accepts - `\w` would let underscores and non-ASCII through,
and `[0-9A-Za-z.+-]` is the exact set CPE 2.3 versions use. The alternation is there only
for `*` and `-`, which aren't versions but are the two values that widen a name back to its
family.

artemis/cpe_tools/cpe_main_process.py
X-R3Y24: Fixed

test/unit/test_cpe_utils.py
X-R3Y24: Rewritten, hopefully clearer now.

test/unit/test_cpe_utils.py
X-R3Y24: No - `wp_scanner` is the only caller and #3103 removed the version it used to pass. Dropped
that assertion; it was covering unused surface.

artemis/cpe_tools/cpe_main_process.py
X-R3Y24: Done

PR #3131, opened 2026-09-08
nuclei more accurate poc when redirections happen

kadewu:


artemis/reporting/modules/nuclei/reporter.py
```diff
@@ -93,6 +93,36 @@ def _get_cpe(vulnerability: Dict[str, Any]) -> Optional[str]:
...
+        return None
+
+    try:
+        method, target, protocol = request.splitlines()[0].split(" ")
+        assert method in ["GET", "POST"], f"{method} is not 'GET'/'POST'"
+        assert (
+            target.startswith("http://") or target.startswith("https://") or target.startswith("/")
+        ), f"{target} should start with proto or /"
+        assert protocol.startswith("HTTP/"), f"{protocol} should start with HTTP/"
+    except IndexError:
```
X-R3Y24, line 104: `except IndexError` doesn't cover what this block raises, and the consequence is larger than
a missing PoC. Ran the function from this branch against request lines nuclei produces:

```
GET, valid                       -> ('/search', 'q=1')
PUT (CVE-2017-12615)             -> AssertionError: PUT is not 'GET'/'POST'
DELETE (apache-druid-log4j-rce)  -> AssertionError: DELETE is not 'GET'/'POST'
extra space (raw template)       -> ValueError: too many values to unpack (expected 3)
```

This runs for every finding carrying a `request`, not only the redirect case. Both templates
run under the default `NUCLEI_SEVERITY_THRESHOLD=high_and_above` and neither is in
`NUCLEI_TEMPLATES_TO_SKIP` - same for CVE-2018-1335 (PUT, high) and CVE-2017-8046 (PATCH,
critical).

Neither `reports_from_task_result` nor `DataLoader._initialize_data_if_needed` wraps that
call, so one such finding stops the loop over every task result, not just its own.

Not sure what you'd prefer here - widening the `except` would work, though catching
`AssertionError` is a bit odd; plain `if ...: return None` reads cleaner to me.

artemis/reporting/modules/nuclei/reporter.py
```diff
@@ -244,6 +284,7 @@ def _is_url_without_query_fragment(url: str) -> bool:
                             "template_name": template,
                             "original_template_name": original_template_name,
                             "curl_command": vulnerability.get("curl-command", None),
+                            "request": vulnerability.get("request", None),
```
X-R3Y24: `request` in `additional_data` - I grepped and nothing reads it: no Jinja template
references it, and the only two hits in the reporting code are this line and the
`extract_request_target` call above. `curl_command` already carries the same headers.
Is it meant for something later, or can it go?

kazet: done

artemis/reporting/modules/nuclei/reporter.py
kazet: added more HTTP methods. If a weird input is encountered I prefer the reporter to crash than report something weirdg

X-R3Y24 (dismissed):

PR #3132, opened 2026-09-09
Detecting technologies

kazet:


artemis/modules/nuclei.py
```diff
@@ -423,9 +426,11 @@ def __init__(self, *args: Any, **kwargs: Any):
                     and not item.startswith("http/exposures/logs/git-")
                 ]
 
-            if "exposed_panels" in Config.Modules.Nuclei.NUCLEI_TEMPLATE_LISTS:
-                template_lists_raw["exposed_panels"] = [
-                    item for item in listed_templates if item.startswith(EXPOSED_PANEL_TEMPLATE_PATH_PREFIX)
+            if "technologies" in Config.Modules.Nuclei.NUCLEI_TEMPLATE_LISTS:
```
X-R3Y24: Any backward compatibility concern with renaming `exposed_panels` to `technologies`?
If someone has `NUCLEI_TEMPLATE_LISTS` set explicitly in their `.env` with
`exposed_panels`, the condition will never match after this change and panels silently
stop being scanned - no error, just fewer findings.

X-R3Y24 (dismissed):

PR #3135, opened 2026-09-10
ResourceLock: release only locks owned by the caller (owner-checked, Lua)

### LINKED ISSUE #3110: bug : ResourceLock.release() can delete a lock now owned by another process (unconditional delete, no ownership check)
### Summary

Every lock stores an owner ID (`self.lid`, `resource_lock.py:73`) and `acquire()` writes it to Redis via `SET ... nx` (line 90) — but `release()` never checks it:

```python
def release(self) -> None:              # resource_lock.py:103-107
    with LOCKS_TO_SUSTAIN_LOCK:
        if self.res_name in LOCKS_TO_SUSTAIN:
            del LOCKS_TO_SUSTAIN[self.res_name]
    REDIS.delete(self.res_name)         # unconditional - deletes whoever now holds the lock
```

The `lid` was recorded but never used on release, so any `release()` call destroys whatever holder currently owns the key, not just the caller's lock.

---

### A Concrete Race in the Current Code

`check_connection_to_base_url_and_save_error()` (`module_base.py:824-870`) — the WAF-skip path — creates a fresh `ResourceLock` object for the same `lock-{scan_destination}` (giving it a different `lid`) and calls `release()` on it when WAF is detected (lines 856, 869). The real owner — the lock object returned by `_take_and_lock_tasks()` — also releases later, after the batch (`module_base.py:406-408`). 

**The sequence:**
1. Scanner A holds `lock-target`.
2. A's WAF-skip releases it early (via the fresh object with the same name).
3. Scanner B acquires the now-free lock and starts scanning.
4. A's owner lock releases at batch end — deleting B's lock.
5. Scanner C can now enter the "locked" target while B is still inside it.

Two scanners racing the same target is precisely what this subsystem exists to prevent. The same unconditional-delete pattern also exists in `release_all_locks()` (line 81), which is invoked at the start of every worker iteration (`module_base.py:337`).

---

### A Second-Order Race of the Same Family

`sustain_locks()` (line 40) refreshes with `REDIS.set(key, value, ex=...)` — an unconditional overwrite with our `lid`. If a lock expired and was re-acquired by another process during a Redis hiccup, our heartbeat would silently overwrite their ownership value, corrupting who the holder is from Redis's point of view.

---

### Proposed Approach

Redis already ships the exact primitive needed: a one-line Lua compare-and-delete. The owner ID is already stored and threaded through `LOCKS_TO_SUSTAIN`, so nothing new needs to be tracked:

```python
_COMPARE_AND_DELETE_SCRIPT = (
    "if redis.call('get', KEYS[1]) == ARGV[1] then return redis.call('del', KEYS[1]) else return 0 end"
)
```

- **`release()`** → delete only if `GET(res_name) == self.lid`
- **`release_all_locks()`** → delete using the `lid` already stored in `LOCKS_TO_SUSTAIN`
- **Optionally, `sustain_locks()`** → a matching compare-and-PEXPIRE, so the heartbeat can never overwrite a foreign owner

> **Deliberate behavior note:** After this change, the WAF-skip path's `lock.release()` becomes a safe no-op (that fresh object is not the owner) — the destination is then released exactly once, by its real owner, at batch end. The net effect matches the intended semantics; the stray release just can't harm a foreign holder anymore.

---

### Status

I have this implemented on a branch, with the tests I believe make the case reviewable:
- A small dependency-free fake Redis (dict-backed, implements `SET nx` / `GET` / `DEL` / the two scripts) so ownership transfer is tested deterministically, no timing or sleeps involved.
- The headline scenario above as an actual test: Scanner A releases → Scanner B acquires → Scanner A's stray release must leave Scanner B's lock intact, and a third holder is refused.
- A test for the exact WAF-path shape (fresh lock object, different `lid`, `release()` is a no-op).
- Tests verifying owner release (including release after a heartbeat refresh) still works, and that `release_all_locks` cannot delete a foreign-owned key.
- The existing lock tests ported to the same fake; full `test/unit` sweep shows no regressions, and `black`/`isort`/`mypy`/`flake8` are clean.

I'm aware this is code you wrote and know deeply, so before I open the PR — does this approach look right to you? If you'd design it differently (e.g., a token passed at release time instead of the object's `lid`, or `RELEASE_IF_MINE` semantics, or leaving `sustain_locks()` for a follow-up), I'd be genuinely glad to follow your direction — you'll see the tradeoffs here faster than I will. Just let me know how you'd like it handled and I'll happily adjust..!

AdeshDeshmukh:
Fixes #3110

## Problem

Every lock records an owner id (`self.lid`, `resource_lock.py:73`) and `acquire()` stores it in Redis via `SET ... nx` - but `release()` never consulted it and called `REDIS.delete()` unconditionally. Any release call therefore deleted *whatever holder currently owned the key*, possibly a different process entirely.

A concrete race exists in the current code: `check_connection_to_base_url_and_save_error()` (`module_base.py:824-870`, the WAF-skip path) creates a **fresh** `ResourceLock` object for the same `lock-{scan_destination}` (a different `lid`) and calls `release()` on it, while the real owner (the lock from `_take_and_lock_tasks()`) also releases after the batch (`module_base.py:406-408`). The interleaving - WAF-skip releases early, another scanner acquires, the original owner releases at batch end deleting the other scanner's lock - lets two scanners race the same target, exactly what this subsystem exists to prevent. The same unconditional pattern existed in `release_all_locks()` and, second-order, in `sustain_locks()` (unconditional `SET` overwriting whoever now owns the key).

## Fix (contained in `resource_lock.py`, no `module_base.py` changes)

Redis already provides the exact primitive needed - an atomic Lua compare-and-delete; the owner `lid` was already threaded through `LOCKS_TO_SUSTAIN`, so nothing new needed tracking:

```python
_COMPARE_AND_DELETE_SCRIPT = (
    "if redis.call('get', KEYS[1]) == ARGV[1] then return redis.call('del', KEYS[1]) else return 0 end"
)
```

- `release()` -> delete only if `GET(res_name) == self.lid`
- `release_all_locks()` -> delete using the `lid` already stored per key in `LOCKS_TO_SUSTAIN`
- `sustain_locks()` -> a matching compare-and-`PEXPIRE` heartbeat, so it can refresh but never *claim* a foreign-owned key

Deliberate behavior note: the WAF-skip path's `lock.release()` becomes a safe no-op (that fresh object was never the owner) - the destination is released exactly once, by its real owner, at batch end.

## Tests

Extended `test/unit/test_resource_lock.py` (existing tests ported to the same harness) with a small dependency-free **FakeRedis** (dict-backed, honoring `SET nx` plus the two Lua scripts) so ownership transfer is tested deterministically, with no sleeps or timing dependence:

- the headline scenario as an actual test: A acquires -> A releases -> B acquires -> A's stray release **must leave B's lock intact**, and a third holder must still be refused
- the exact WAF-path shape: fresh lock object, same key, different `lid`, `release()` is a no-op
- owner release still works, including after a heartbeat refresh (guards against over-tightening)
- `release_all_locks` cannot delete a foreign-owned key; sustain compare-and-set respects ownership

## Verification

- [x] `pytest test/unit/test_resource_lock.py` - 15/15 passed
- [x] Full `test/unit` sweep - no regressions (remaining failures are byte-identical on clean main: they need the Docker test environment)
- [x] Both Lua scripts also validated directly against a real Redis (owner delete -> 1, foreign delete -> 0 with key intact, owner heartbeat -> 1)
- [x] `pre-commit run` - black, isort, mypy, flake8 all pass

### DIFF
--- a/artemis/resource_lock.py
+++ b/artemis/resource_lock.py
@@ -31,13 +31,32 @@ LOCKS_TO_SUSTAIN_LOCK = threading.Lock()
 
 REDIS = Redis.from_url(Config.Data.REDIS_CONN_STR)
 
+# Deletes the key only if it is still owned by the expected owner, to prevent one process from
+# releasing a lock that has already expired and been re-acquired by another process.
+_COMPARE_AND_DELETE_SCRIPT = (
+    "if redis.call('get', KEYS[1]) == ARGV[1] then return redis.call('del', KEYS[1]) else return 0 end"
+)
+
+# Refreshes the key's expiry only if it is still owned by the expected owner, for the same reason.
+_COMPARE_AND_SET_SCRIPT = (
+    "if redis.call('get', KEYS[1]) == ARGV[1] then " "return redis.call('pexpire', KEYS[1], ARGV[2]) else return 0 end"
+)
+
+
+def _compare_and_delete(res_name: str, owner_lid: str) -> bool:
+    return bool(REDIS.eval(_COMPARE_AND_DELETE_SCRIPT, 1, res_name, owner_lid))  # type: ignore
+
+
+def _compare_and_set(res_name: str, owner_lid: str, ex_seconds: int) -> bool:
+    return bool(REDIS.eval(_COMPARE_AND_SET_SCRIPT, 1, res_name, owner_lid, str(ex_seconds * 1000)))  # type: ignore
+
 
 def sustain_locks() -> None:
     while True:
         try:
             with LOCKS_TO_SUSTAIN_LOCK:
                 for key, value in LOCKS_TO_SUSTAIN.items():
-                    REDIS.set(key, value, ex=LOCK_HEARTBEAT_TIMEOUT)
+                    _compare_and_set(key, value, LOCK_HEARTBEAT_TIMEOUT)
         except Exception:
             logger.exception("Failed to sustain locks, will retry")
         time.sleep(1)
@@ -76,9 +95,9 @@ class ResourceLock:
     @staticmethod
     def release_all_locks(logger: Logger) -> None:
         with LOCKS_TO_SUSTAIN_LOCK:
-            for lock in list(LOCKS_TO_SUSTAIN.keys()):
-                logger.info(f"Releasing lock: {lock} -> {LOCKS_TO_SUSTAIN[lock]}")
-                REDIS.delete(lock)
+            for lock, owner_lid in list(LOCKS_TO_SUSTAIN.items()):
+                logger.info(f"Releasing lock: {lock} -> {owner_lid}")
+                _compare_and_delete(lock, owner_lid)
             LOCKS_TO_SUSTAIN.clear()
 
     def acquire(self) -> None:
@@ -104,7 +123,7 @@ class ResourceLock:
         with LOCKS_TO_SUSTAIN_LOCK:
             if self.res_name in LOCKS_TO_SUSTAIN:
                 del LOCKS_TO_SUSTAIN[self.res_name]
-        REDIS.delete(self.res_name)
+        _compare_and_delete(self.res_name, self.lid)
 
     def __enter__(self) -> None:
         self.acquire()
--- a/test/unit/test_resource_lock.py
+++ b/test/unit/test_resource_lock.py
@@ -1,7 +1,8 @@
 import logging
 import threading
 import unittest
-from unittest.mock import MagicMock, patch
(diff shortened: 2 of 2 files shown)


kadewu:
I've closed already multiple same looking PR's. I understand that no lock ownership is a problematic, but our lock mechanism is not that crucial for the operating system. The tradeoff of WAF lock releasing is bigger than possible lock race condition.

PR #3137, opened 2026-09-10
RedisCache.flush(): delete only this cache's keys, not flushall()

### LINKED ISSUE #3136: bug : RedisCache.flush() calls flushall() — clearing one module's cache would wipe Karton queues, locks, and all other caches
## 1. Summary

`RedisCache` (`artemis/redis_cache.py`) is the namespaced cache wrapper utilized across scanning modules (`module_base.py:96`), the crawl pipeline (`crawling.py:26`), and the Wordfence index (`wordfence.py:49`). Keys are namespaced under `cache.{name}:{key}`. However, `flush()` bypasses namespacing entirely:

```python
def flush(self) -> None:
    self.redis.flushall()  # artemis/redis_cache.py:23-24
```

`flushall()` purges all databases across the Redis instance. Because this shared instance also manages Karton task queues, task state, module locks, analysis counters, and metrics, invoking `flush()` results in an unrecoverable wipe of scanner state: pending workloads vanish, distributed locks drop, and in-flight counters desynchronize.

---

## 2. Scope & Status: Latent Vulnerability

A full audit of `.flush()` call sites confirms they are currently confined to Karton's internal task-cache, standard I/O stream flushes, and test setup (`test/base.py:28`). No live production paths invoke `RedisCache.flush()`.

The risk lies in API ergonomics: calling `module.cache.flush()` is standard practice for maintenance tasks, administrative endpoints, or debug workflows. An innocent invocation intended to clear an individual cache partition would inadvertently bring down the active workload.

---

## 3. Proposed Approach

Update `RedisCache.flush()` to restrict deletions strictly to the caller's key prefix, preserving the expected API contract without destructive side effects:

```python
def flush(self) -> None:
    # Delete only this cache's keys. Never flushall()/flushdb() here: Karton
    # queues, locks, and other caches share the same Redis instance/database,
    # so flushdb would be equally destructive.
    safe_name = self.cache_name.translate(_GLOB_ESCAPE_TABLE)
    batch: List[Any] = []
    for key in self.redis.scan_iter(match=f"{safe_name}:*", count=1000):
        batch.append(key)
        if len(batch) >= 1000:
            self.redis.delete(*batch)
            batch = []
    if batch:
        self.redis.delete(*batch)
```

### Design Considerations
- **Non-blocking iteration:** Uses `scan_iter` rather than `KEYS` to avoid latency spikes and thread contention on a shared Redis instance.
- **Chunked pipelines:** Deletions execute in batches of 1,000 keys to bound memory overhead while minimizing network round-trips.
- **Redis-compliant glob escaping:** Unlike `fnmatch`, Redis glob patterns treat `[]` as an empty, never-matching set; closing brackets must be escaped as `\]`. The escape table maps `* → [*]`, `? → [?]`, `[ → [[]`, `] → \]`, and `\ → \\`.
- **Exclusion of `flushdb()`:** Excluded because Karton queues, synchronization primitives, and cache entries reside within the same Redis database.

---

## 4. Test Plan

A dedicated test suite (`test/unit/test_redis_cache.py`) backed by an in-memory mock engine adhering to Redis `stringmatchlen` semantics:

- **Scoped purge:** Validates that flushing `cache.modA:*` does not touch `cache.modB:*`, `karton.queue.*`, or lock/counter keys.
- **Multi-instance isolation:** Asserts two distinct `RedisCache` namespaces operate independently without cross-invalidation.
- **Empty state handling:** Confirms no-op behavior on an empty namespace without errors or unnecessary `DELETE` calls.
- **Pattern literal validation:** Verifies that cache names with special glob characters (`[`, `]`, `?`, `\`) do not evaluate as wildcards.
- **Contract integrity:** Standard `get`/`set` round-trip regression verification.

---

## 5. Alternatives & Discussion

- **Remove `flush()` entirely:** If maintaining a manual flush method is deemed unnecessary given the absence of production callers, removing the method outright is a valid option that eliminates the risk surface.
- **Scoped implementation (preferred):** Retaining a safe, scoped implementation preserves expected caching primitives for future administrative interfaces or targeted cache-busting workflows without risk to shared state.

AdeshDeshmukh:
Fixes #3136

## Problem

`RedisCache.flush()` called `redis.flushall()`, wiping **all databases on the entire shared Redis instance** - Karton task queues, task state, module locks, analysis counters, and every other cache - instead of clearing just this cache's `cache.{name}:*` keys. A deployment hit by this doesn't degrade, it loses its entire pending workload.

As noted in the issue: no production code calls `flush()` today (verified - all other `.flush()` hits are Karton's own cache, file streams, or test setup), so this is a loaded footgun rather than a live outage. The method's name and location invite correct-looking usage (`module.cache.flush()` in future tooling or a UI button), and the first such innocent call would take down the whole deployment.

## Fix (contained in `artemis/redis_cache.py`)

Delete only this cache's keys via non-blocking SCAN with batched deletes and Redis-correct glob escaping of the cache name:

```python
safe_name = self.cache_name.translate(_GLOB_ESCAPE_TABLE)
batch: List[Any] = []
for key in self.redis.scan_iter(match=f"{safe_name}:*", count=1000):
    batch.append(key)
    if len(batch) >= 1000:
        self.redis.delete(*batch)
        batch = []
if batch:
    self.redis.delete(*batch)
```

- `scan_iter`, not `KEYS` - non-blocking cursor scan on the busy shared Redis
- Batched `delete` (1000) - one round-trip per batch, bounded memory even for large crawl caches
- Glob-escaping verified Redis-correct against real Redis (note: unlike fnmatch, Redis treats `[]` as an empty, never-matching class, so `]` is escaped as `\]`): `*`->`[*]`, `?`->`[?]`, `[`->`[[]`, `]`->`\]`, `\`->`\\`
- `flushdb` deliberately not used - queues/locks live in the same DB behind the same shared client, so it would be equally destructive

## Tests

New `test/unit/test_redis_cache.py` with a small dependency-free fake Redis (dict-backed, bytes-faithful, matcher mirroring Redis `stringmatchlen` semantics - differential-tested against a real Redis on adversarial patterns with zero mismatches):

- scoped flush removes only `cache.modA:*` while `karton.queue.*`, `lock-*`, counters and `cache.modB:*` survive
- two `RedisCache` instances - one's flush leaves the other's keys intact
- empty flush issues zero deletes; glob metacharacters (`[`, `]`, `?`, `\`) in a cache name match literally; `get`/`set` contract untouched

## Verification

- [x] `pytest test/unit/test_redis_cache.py` - 4/4 passed, zero fixtures needed (imports only `redis`)
- [x] Real-Redis end-to-end - seeded mixed keys, `flush()`, only the cache's own keys gone
- [x] `pre-commit run` - black, isort, mypy, flake8 all pass

### DIFF
--- a/artemis/redis_cache.py
+++ b/artemis/redis_cache.py
@@ -1,6 +1,17 @@
-from typing import Optional
+from typing import Any, List, Optional
 
 from redis import Redis
+
+# Escapes Redis glob metacharacters so cache names are always matched literally.
+# Note: unlike fnmatch/Tcl, Redis treats [] as an empty (never-matching) class,
+# so ] must be escaped as \] rather than []]. Verified against real Redis.
+_GLOB_ESCAPE_TABLE = {
+    ord("*"): "[*]",
+    ord("?"): "[?]",
+    ord("["): "[[]",
+    ord("]"): "\\]",
+    ord("\\"): "\\\\",
+}
 
 
 class RedisCache:
@@ -21,4 +32,15 @@ class RedisCache:
         self.redis.set(f"{self.cache_name}:{key}", value, ex=timeout)
 
     def flush(self) -> None:
-        self.redis.flushall()
+        # Delete only this cache's keys. Never flushall()/flushdb() here: Karton
+        # queues, locks, and other caches share the same Redis instance/database,
+        # so flushdb would be equally destructive.
+        safe_name = self.cache_name.translate(_GLOB_ESCAPE_TABLE)
+        batch: List[Any] = []
+        for key in self.redis.scan_iter(match=f"{safe_name}:*", count=1000):
+            batch.append(key)
+            if len(batch) >= 1000:
+                self.redis.delete(*batch)
+                batch = []
+        if batch:
+            self.redis.delete(*batch)
--- a/test/unit/test_redis_cache.py
+++ b/test/unit/test_redis_cache.py
@@ -1,0 +1,175 @@
+import unittest
+from typing import Any, Dict, List, Optional, Tuple
+
+from artemis.redis_cache import RedisCache
+
+
+def redis_glob_match(name: bytes, pattern: bytes) -> bool:
+    """Match bytes against a Redis glob pattern, mirroring Redis stringmatchlen
+    semantics (which differ from fnmatch in edge cases - e.g. Redis treats []
+    as an empty, never-matching class, so fnmatch must NOT be used here)."""
+
+    def match(si: int, pi: int) -> bool:
+        while pi < len(pattern):
+            c = pattern[pi : pi + 1]
+            if c == b"*":
+                while pi < len(pattern) and pattern[pi : pi + 1] == b"*":
+                    pi += 1
+                if pi == len(pattern):
+                    return True
+                for skip in range(si, len(name) + 1):
+                    if match(skip, pi):
+                        return True
+                return False
+            if si >= len(name):
+                return False
+            if c == b"?":
+                si += 1
+                pi += 1
+            elif c == b"\\":
+                pi += 1
+                if pi >= len(pattern) or name[si : si + 1] != pattern[pi : pi + 1]:
+                    return False
+                si += 1
+                pi += 1
+            elif c == b"[":
+                pi += 1
+                negate = pi < len(pattern) and pattern[pi : pi + 1] == b"^"
+                if negate:
+                    pi += 1
+                found = False
+                while True:
+                    if pi >= len(pattern):
+                        return False
+                    if pattern[pi : pi + 1] == b"]":
+                        # No first-position exception (unlike fnmatch): in Redis,
+                        # [] is an empty, never-matching class.
+                        pi += 1
+                        break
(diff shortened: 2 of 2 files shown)


kadewu:
If it's dead code then I see no reason for this change, we can just drop it.

AdeshDeshmukh:
indeed...
However @kadewu sir before opening an issue , i was considering to discuss the approach and relevancy..!
I dont saw any slack or other link in our readme... What will be a good place from ur pov..!

kadewu:
Regarding this PR I don't think there is much to discuss, if you are willing, just simply delete the dead code.
I believe it's perfectly fine to discuss things here, but feel free to join our Discord as well: https://discord.gg/Q7eTsHnpn4

AdeshDeshmukh:
yes sir i am. considering to discuss from onwards  before filing an issue , so this kind of mistakes wont happen in the future..!

kazet:
Thanks for the PR, although it may be made simpler: https://github.com/CERT-Polska/Artemis/pull/3141

PR #3140, opened 2026-09-11
Improving subdomain enumeration speed

kazet:


test/modules/test_subdomain_enumeration.py
```diff
@@ -10,6 +10,9 @@
 class SubdomainEnumerationScannerTest(ArtemisModuleTestCase):
     karton_class = SubdomainEnumeration  # type: ignore
 
+    def test_dns_brute(self) -> None:
+        self.assertEqual(self.karton.get_subdomains_by_dns_brute_force("cert.pl"), ["www.cert.pl"])
```
kadewu:
```suggestion
        self.assertEqual(self.karton.get_subdomains_by_dns_brute_force(set(), "cert.pl"), ["www.cert.pl"])
```
and best would be to move it to external_network

kadewu (dismissed):

PR #3144, opened 2026-09-14
Add e2e coverage for CPE in WordPress reporters

X-R3Y24:
## What
 
Gives the reporting integration tests a real (if tiny) NVD CPE dictionary, and adds e2e coverage that
the WordPress reporters actually write a CPE into `Asset.cpe` - for `wordpress_plugins` and
`wp_scanner`.
 
## Why
 
The `Asset.cpe` assertions in `test/reporting/` were passing in CI without checking anything.
 
`Config.CpeDictionary.CPE_NVD_DIR` defaults to the bundled `artemis/cpe_tools/nvdcpe-2.0`, which is
gitignored - it exists only on a machine that has run the downloader. `docker-compose.test.yaml`
never set that variable for the `test` container, so in CI the directory is simply absent.
`_ensure_index()` then finds no chunks, logs *"NVD CPE dictionary not found; CPE lookups will return
None"*, and returns `{}`. Every `lookup_cpe*()` returns `None`, every reporter leaves `Asset.cpe` as
`None`, and an assertion that an asset carries a CPE passes vacuously.
 
Locally the same assertion passed for the opposite reason - the dictionary happens to be downloaded
there. That local/CI split is what kept it hidden: the tests were green everywhere, and green meant
two different things.
 
This is the coverage #3082 (plugin CPEs) should have had end to end.
 
## Fix
 
- `test/data/mock_nvd_cpe/` - a stand-in dictionary, ten entries copied verbatim from the real feed.
- `CPE_NVD_DIR: /opt/test/data/mock_nvd_cpe` on the `test` service. No new mount needed; `./test/` is
already mounted there.
- `test-wordpress-with-plugins` - an nginx container serving a WordPress-shaped static site with two
plugin readmes. A stock WordPress container has none, and the module identifies plugins by fetching
`/wp-content/plugins/<slug>/readme.txt`.
- The new container is added to the readiness loop in `scripts/test-unit`, alongside the other
static-site services.
Two design decisions worth stating:
 
**The fixture stores raw feed chunks, not a ready-made index.** `_ensure_index()` finds no index file
and falls through to `_build_indices()`, so the index builder runs on every test run rather than
being bypassed by a checked-in artifact.
 
**The tests do not patch `lookup_cpe*`** - that is the thing under test. What is stubbed is the
module's own dependency on api.wordpress.org (`FallbackAPICache.get`), so the tests need no network.
That matters: wordpress.org is intermittently blocked from GitHub runners, which is why
`test/reporting/test_wordpress_plugins.py` skips there.
 
## Tests
 
- `WordpressPluginsAutoreporterIntegrationTest.test_asset_discovery` - verifies a detected plugin
carries the CPE the dictionary names for its slug, with `*` left in the version slot while the
detected version stays in `Asset.version`; and that a plugin the dictionary does not know is still
reported as an asset, with `cpe=None`.
- `WPScannerAutoreporterIntegrationTest.test_asset_discovery` - verifies the `wordpress` CMS asset
carries the WordPress CPE. Nothing checked this end to end before.
Both assertions fail when the dictionary is empty. Pointing the variable at an empty directory fails
exactly those two and nothing else - `version='5.9.3'` stays intact and `test_reporting` still passes:
Two more under `test/external_network/`, which needs the network and runs as a separate
workflow:
- `test_nvd_cpe_feed.py` pulls one chunk of the live feed - streamed, stopped after the first
file,  and asserts that the fields `cpe_tools` parses are still shaped the way the code assumes:
13 components out of `split_cpe`, a title with `lang == "en"`, `refs` as mappings carrying a `ref`,
`deprecated` as a bool, and `wordpress.org/plugins/<slug>/` references still being published.
Assertions are on proportions rather than every record - NVD is someone else's dataset with
the occasional oddity, and a canary that goes red on one bad row in 140,000 gets switched off.
- `test_nvd_cpe_dictionary.py` builds the whole dictionary from the live feed into a scratch
directory and looks three products up: by title, by plugin slug and by url. No fixture and
no mock beyond pointing `CPE_NVD_DIR` at a temporary path - the download, the unpacking,
`_build_indices` and the lookups all run for real. Takes about 4 minutes.
- Also updated: `test_a_single_description_is_left_alone` in `test_asset_deduplication.py`
asserted `asset.cpe is None`. That held only because the lookup had nothing to answer with -
with the fixture in place, `wp_scanner` now sets the WordPress CPE and the assertion fails.
It is the same vacuous pass this PR is about, caught in a test that was already there.
 
```
CPE_NVD_DIR=/opt/test/data/mock_nvd_cpe    3 passed
CPE_NVD_DIR=/nonexistent                   2 failed, 1 passed
 
-    cpe=None,
+    cpe='cpe:2.3:a:wordpress:wordpress:*:*:*:*:*:*:*:*',
     version='5.9.3',   <- unchanged
```
 
## Notes / Out of scope
 
- **`docker-compose.test-e2e.yaml` deliberately untouched.** That stack scans
`test-smtp-server.artemis` with only `mail_dns_scanner` and `classifier` enabled, so no
CPE-producing reporter runs in it and the only asserted assets are domain ones with `cpe: None`.
Wiring the fixture in would change nothing.
- **The fixture is larger than the assertions need, on purpose, and its README says so.** Only two
entries are load-bearing; the deprecated names, akismet's ref with a `#developers` fragment and the
extra WordPress releases are there so the lookups run against something feed-shaped instead of a
two-element map. Dropping any single entry still leaves both tests green. Version trimming,
per-product title keying and deprecated filtering are asserted in `test/unit/test_cpe_utils.py`, on
fixtures written for exactly that - this directory is not where that behaviour is covered, and the
README is explicit about it so nobody reads the entry list as a coverage claim.
- **The `cpe=None` assertion on kirki carries no weight on its own** - an empty dictionary satisfies
it just as well. It means something only next to contact-form-7 in the same method, so the two must
not be split into separate tests. There is a comment saying so.
- The url index is built by the same `_build_indices()` call as the other two, so the fixture covers
it on the way through. That no reporter calls `lookup_cpe_by_url` yet is the current state of the
repo, not something this PR leaves undone.


test/reporting/test_wordpress_plugins_autoreporter_integration.py
```diff
@@ -64,8 +68,8 @@ def test_asset_discovery(self) -> None:
         self.assertEqual(contact_form_7.asset_type, AssetType.CMS_PLUGIN)
         self.assertEqual(contact_form_7.name, URL)
         cpe = contact_form_7.cpe
-        assert cpe is not None
         self.assertEqual(cpe, CONTACT_FORM_7_CPE)
+        assert cpe is not None  # narrowing for split_cpe below
```
kazet, line 72: can we assert that the CPE is equal to a given one?

test/data/mock_nvd_cpe/README.md
```diff
@@ -0,0 +1,31 @@
+# Mock NVD CPE dictionary
+
+A stand-in for the NVD CPE 2.0 dictionary, pointed at by `CPE_NVD_DIR` in the test container so that
+`artemis.cpe_tools.cpe_utils` answers lookups in the integration tests. Without it the tests run
+against an empty directory in CI - every lookup returns `None`, and an assertion that an asset
```
kazet: hm, I don't understand this sentence

test/reporting/test_wordpress_plugins_autoreporter_integration.py
X-R3Y24: It is, one line above: `self.assertEqual(cpe, CONTACT_FORM_7_CPE)`. The `assert` below it
only narrows the type for mypy - `Asset.cpe` is `str | None` and `split_cpe()` on line 76
takes a `str`.

test/data/mock_nvd_cpe/README.md
X-R3Y24: Changed it. Hopefully clearer now

kazet (dismissed):

